PR #3, opened 2020-09-03
fix runnable documentation examples

HippoBaro:
### What does this PR do?

Cargo compiles and runs code snippets in code documentation when testing
code. This commit updates them so that they are valid and use the latest
APIs.

Because scipio is a TPC async framework, it is sometimes difficult to
demonstrate a feature in the documentation (for example, the `Parker`
object). In these cases, they are marked as `no_run`.

Some other code snippets were left as-is because they rely on APIs that
don't exist anymore. For example, `multitask` examples that rely on a
now removed `LocalExecutor::ticker`.

### Motivation

`cargo test` was failing and some of the documentation was out-of-date

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -11,13 +11,13 @@
 //!
 //! Run a single-threaded and a multi-threaded executor at the same time:
 //!
-//! ```
+//! ```no_run
 //! use async_channel::unbounded;
-//! use async_executor::{Executor, LocalExecutor};
+//! use scipio::{Executor, LocalExecutor};
 //! use easy_parallel::Parallel;
 //!
 //! let ex = Executor::new();
-//! let local_ex = LocalExecutor::new();
+//! let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
 //! let (trigger, shutdown) = unbounded::<()>();
 //!
 //! Parallel::new()
@@ -296,9 +296,9 @@ impl ExecutorQueues {
 /// # Examples
 ///
 /// ```
-/// use async_executor::LocalExecutor;
+/// use scipio::LocalExecutor;
 ///
-/// let local_ex = LocalExecutor::new(None);
+/// let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
 ///
 /// local_ex.run(async {
 ///     println!("Hello world!");
@@ -317,13 +317,13 @@ impl LocalExecutor {
     /// # Examples
     ///
     /// ```
-    /// use async_executor::LocalExecutor;
+    /// use scipio::LocalExecutor;
     ///
     /// // executor is a single thread, but not bound to any particular CPU.
-    /// let local_ex = LocalExecutor::new(None);
+    /// let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
     ///
     /// // executor is a a single thread, bound to logical CPU 1.
-    /// let bound_ex = LocalExecutor::new(Some(1));
+    /// let bound_ex = LocalExecutor::new(Some(1)).expect("failed to create local executor");
     /// ```
     pub fn new(binding: Option<usize>) -> io::Result<LocalExecutor> {
         let (p, _) = parking::pair();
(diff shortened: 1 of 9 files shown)


glommer:
Hi

Very good progress! I tried it locally and I still see some tests failing, and also at some point the docs test just hang.
Does it always pass for you?

I'll go through the examples now

src/executor.rs
```diff
@@ -11,13 +11,13 @@
 //!
 //! Run a single-threaded and a multi-threaded executor at the same time:
```
glommer, line 12: for next time => if you are making changes close to descriptions like this, it's a good time to update the descriptions (I should have done it)

there is no multi-threaded executor, only async-executor had one.

This won't block the merge.

src/executor.rs
```diff
@@ -11,13 +11,13 @@
...
 //!
-//! ```
+//! ```no_run
```
glommer, line 14: This is the main example at the top level of the framework.

It should not be marked as no_run. Maybe we can demonstrate it with std::thread creating a simple thread.
`shutdown()` will also obviously not work. This is a good candidate for us to just use our internal `Timer` future.

HippoBaro:
Some tests should still be failing as they were left untouched. `cargo test` shouldn't hang however.
On my local machine I get `test result: FAILED. 68 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out` consistently across runs.

src/executor.rs
```diff
@@ -726,10 +727,10 @@ impl<T> Task<T> {
     /// # Examples
     ///
     /// ```
-    /// use async_executor::{Executor, Task};
+    /// use scipio::{LocalExecutor, Timer};
     /// use futures_lite::future;
     ///
-    /// let ex = Executor::new();
+    /// let ex = LocalExecutor::new(None).expect("failed to create local executor");
```
glommer, line 733: my understanding is that `yield_now` blocks the thread, which is why we have our own version, called `later()`.
At the very least we should use `later()` in the examples.

I would like to see this one fixed, please add a commit on top addressing this.

glommer (changes requested):

glommer:
All right. This is clear progress so I'd like to merge. We'll deal with the remaining failures later.

I am requesting one change, which is to get rid of yield_now() on examples in favor of `later()`. The others are just suggestions and can be done `later()`. (tu-dum-tsss!)

src/executor.rs
HippoBaro: This was changed in e4955970a4a2b85427f537df42d595d7a79082bc in favor of a simpler example that uses `std::thread`s.

src/executor.rs
HippoBaro: I tool e4955970a4a2b85427f537df42d595d7a79082bc as an opportunity to update the documentation to:
```
//! # Examples
//!
//! Run four single-threaded executors concurrently:
//!
//! ```no_run
//! use scipio::{LocalExecutor, Timer};
//!
//! for i in 0..4 {
//!     std::thread::spawn(move || {
//!         let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
//!         local_ex.run(async {
//!             Timer::new(std::time::Duration::from_millis(100)).await;
//!             println!("Hello world!");
//!         });
//!     });
//! }
//! ```
```

This is shorter and simpler to understand without knowledge of external crates.

src/executor.rs
```diff
@@ -727,20 +727,20 @@ impl<T> Task<T> {
...
-    /// use scipio::{LocalExecutor, Timer};
+    /// use scipio::{LocalExecutor, Timer, Task};
     /// use futures_lite::future;
     ///
     /// let ex = LocalExecutor::new(None).expect("failed to create local executor");
     ///
     /// ex.spawn(async {
     ///     loop {
     ///         println!("I'm a background task looping forever.");
-    ///         future::yield_now().await;
```
glommer, line 737: A general comment about your commit message. Doing it here just because github doesn't allow comments on commit messages =)

This API is not unpublished as you call it. It's part of standard rust. The problem is that it is blocking.

I wish there was a way to detect and block automatically usage of this to make sure users are not falling into this trap. If that doesn't prove to be possible then we should update the README.

code is good.

glommer (approved):

glommer:
@HippoBaro I still have a failing test on my end. However this is a very clear improvement and since it works for you it will be harder for you to figure out anyway.

Great job, I'm merging it and we can figure out the failure later.

PR #8, opened 2020-09-04
add proper .gitignore to cleanup git status

sitano:


glommer:
Thank you! I think your subject got broken, though. "to cleanup git st". What is st?

.gitignore
```diff
@@ -0,0 +1,4 @@
+.idea
+*.iml
+/target
+Cargo.lock
```
glommer, line 4: Cargo.lock is indeed pretty important and target is good too.

The others I don't mind, it seems like some IDE stuff, right?

glommer (changes requested):

.gitignore
sitano: > Cargo.lock is indeed pretty important and target is good too.

Do you mean that `/target` and `cargo.lock` must not be in .gitignore? Or you agree?

Others are IntelliJ project files.

sitano:
> Thank you! I think your subject got broken, though. "to cleanup git st". What is st?

`git st` is an alias for `git status`

.gitignore
sitano: I mean you are never going to commit /target and lock into the repo, right? Then they must be hidden from the git status.

glommer: I was just voicing agreement, sorry if I confused you.

As far as I know we can add files from various IDEs here too, so I am cool with that.

glommer (approved):

PR #9, opened 2020-09-04
fix DmaFile.drop() fd != -1 file drop

sitano:
`std::mem::drop()` does nothing when drop references.
so that `drop(&self.file)` must be invalid.

The only way to close a file is to call a syscall with fd.

Other option could be using `ManualDrop<>`: https://stackoverflow.com/questions/41053542/forcing-the-order-in-which-struct-fields-are-dropped

Disclaimer
---

I didn't actually run it to test it. Sorry.

### DIFF
--- a/src/dma_file.rs
+++ b/src/dma_file.rs
@@ -240,7 +240,7 @@ I will close it and turn a leak bug into a performance bug. Please investigate",
                 self.path,
                 self.as_raw_fd()
             );
-            drop(&self.file);
+            let _ = unsafe { libc::close(self.file.as_raw_fd()) };
         }
     }
 }


glommer (commented):
Hello Ivan. 

It was my understanding that drop() would call drop recursively on all member elements of the struct.

Because DmaFile owns std::fs::File, it should call drop on it and end up closing.

Did you notice that we are failing to close, or just figured that out from reading the description of drop?

It would be helpful to run a quick test where you open and close a file and run that through strace to see what's going on

glommer:
Hi Ivan

I just wrote a unit test to demonstrate that this behavior is indeed working. In general, this kind of change is stronger if it comes accompanied by a unit test that fails before your change, but passes after.

I have added a PR with the unit test here:
https://github.com/DataDog/scipio/pull/11

You are right that this drop is not needed - many thanks for that! But we don't actually have to replace it with the explicit close so the code gets even simpler.

sitano:
> It was my understanding that drop() would call drop recursively on all member elements of the struct.

Yes. When an element is dropped all it's content will also be dropped.

> Because DmaFile owns std::fs::File, it should call drop on it and end up closing.

Yes. However, DmaFile implements `Drop`. But `Drop.drop()` can't drop what its not borrowing. drop() has &mut self which ref. So it can't drop fields from File. Also File does not impl Drop so you can't file.drop() explicitly. Drop has its internals.

> Did you notice that we are failing to close, or just figured that out from reading the description of drop?

I figured out from reading code and static analysis warning on `drop`. I am thinking adding it in next PR.

> It would be helpful to run a quick test where you open and close a file and run that through strace to see what's going on

Totally. Sorry, I was just looking through the code in the middle of the working day. Test totally makes sense.

PR #10, opened 2020-09-04
Adds syntax highlighting

sobolevn:


glommer:
very nice trick! I had no idea that existed, it feels like magic

PR #12, opened 2020-09-04
Move library to "scipio" directory

penberg:
### What does this PR do?

Make the top-level directory a Cargo workspace so that we can easily add
things like "examples".

### Motivation

See above.

### Related issues

N/A

### Additional Notes

N/A

### Checklist

[-] I have added unit tests to the code I am submitting
[-] My unit tests cover both failure and success scenarios
[-] If applicable, I have discussed my architecture
[-] The new code I am adding is formatted using `rustfmt`


glommer:
Hi @penberg . I am a bit doubtful about the value of this.

In all other crates I've looked so far, `examples`, `tests` and other are in the same level as `src`.
Take for instance `async-io`

```
async-io]$ ls
Cargo.lock  Cargo.toml  CHANGELOG.md  examples  LICENSE-APACHE  LICENSE-MIT  LICENSE-THIRD-PARTY  README.md  src  target  tests
```

What is it exactly that you are trying to do and failing ?

penberg:
Most of the libraries I have looked at follow this pattern. See, for example, https://github.com/tokio-rs/tokio. The benefit of using Cargo workspaces is that you have the ability to break your monolith crate if needed.

glommer:
So what you are saying is that the flat layout we are adopting would prevent using workspaces? I know for a fact that @HippoBaro was interested in using Scipio inside a workspace.

If this is a blocker for using workspaces I am ok with the change.

penberg:
@glommer It's not a blocker for using the library in a workspace. However, look at the "async-io" repository -- the examples are not _built_ as part of "cargo build"...

HippoBaro:
I am currently using scipio inside a local workspace I created. It makes using my IDE easier. I am not against using a workspace upstream either. I like the idea of having a nested crate containing example programs that only have access to the public APIs. Something that people can easily use as a starting point.

matklad:
To give some more context here:

There are two ways to add things like examples or benchmarks. 

* the first is to use build-in convention, and add top-level `/examples` and `/benches` directories to an existing cargo packages
* fully custom way of creating a cargo workspace, and adding examples as separate packages. 

The workspace approach might be beneficial for huge, sprawling workspaces like tokio.

However, if you have only a single crate, sticking to non-workspace setup seems better to me, as it works with build-in cargo commands like `cargo run --example my-example` or `cargo bench`.

So, I suggest to maybe go back to non-workspace setup, at least until we want to split scipio into a bucnh of subcrates. Even then, a setup like this would work I think:

```
Cargo.toml
src/
  lib.rs # scipio main code
examples/ # scipio examples
scipio-utils/ # hypothetical second workspace crate
  Cargo.tom
```

PR #13, opened 2020-09-04
add makefile with audit and clippy support

sitano:
Add linter support

Try

```
$ make clippy
$ make audit
```

https://github.com/rust-lang/rust-clippy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,72 @@
+SHELL := /bin/bash
+PROJECT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+
+BIN_PATH = $(CURDIR)/bin
+CARGO_TARGET_DIR ?= $(CURDIR)/target
+
+# Build-time environment, captured for reporting by the application binary
+BUILD_INFO_GIT_FALLBACK := "Unknown (no git or not git repo)"
+BUILD_INFO_RUSTC_FALLBACK := "Unknown"
+export SCIPIO_ENABLE_FEATURES := ${ENABLE_FEATURES}
+export SCIPIO_BUILD_TIME := $(shell date -u '+%Y-%m-%d %I:%M:%S')
+export SCIPIO_BUILD_RUSTC_VERSION := $(shell rustc --version 2> /dev/null || echo ${BUILD_INFO_RUSTC_FALLBACK})
+export SCIPIO_BUILD_GIT_HASH ?= $(shell git rev-parse HEAD 2> /dev/null || echo ${BUILD_INFO_GIT_FALLBACK})
+export SCIPIO_BUILD_GIT_TAG ?= $(shell git describe --tag || echo ${BUILD_INFO_GIT_FALLBACK})
+export SCIPIO_BUILD_GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD 2> /dev/null || echo ${BUILD_INFO_GIT_FALLBACK})
+
+# https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199/68
+export CARGO_BUILD_PIPELINING=true
+
+default: release
+
+.PHONY: all
+
+clean:
+	cargo clean
+	rm -rf bin dist
+
+
+## Development builds
+## ------------------
+
+all: format build test
+
+dev: format clippy
+	@env FAIL_POINT=1 make test
+
+build:
+	cargo build
+
+## Static analysis
+## ---------------
+
+unset-override:
+	@# unset first in case of any previous overrides
+	@if rustup override list | grep `pwd` > /dev/null; then rustup override unset; fi
+
+pre-format: unset-override
+	@rustup component add rustfmt
+
+format: pre-format
+	@cargo fmt --all -- --check >/dev/null || \
+	cargo fmt --all
+
+doc:
+	@cargo doc --workspace --document-private-items
+
+pre-clippy: unset-override
+	@rustup component add clippy
+
+clippy: pre-clippy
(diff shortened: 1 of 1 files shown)


glommer:
Hi Ivan,

Thanks a lot, I love the idea of adding even more linters. @HippoBaro , could we add those to our CI flow ?

Our CI is very poor at the moment because what we tried to set up didn't support io_uring, but I'd like to add at least formatting and compilation to the CI ASAP.

I am going to merge this, but right now the clippy stuff is returning a lot of output. @sitano normally I would delay merging this until you fixed all the issues raised by the tool. However because today is Friday, Monday is a holiday, and both Hippo and I are quite active today I am afraid this will just generate you a ton of conflicts.

But please do follow up with a patch that fix the complaints that the tool raised.

glommer:
For posterity, I like `return` statements and I have been trying to use them. That is one of the top complaints the tool raised.

However no matter how much I like a particular style, my hatred for style related discussions trumps it all.
If we can mechanically enforce a style that is well accepted in the community, I don't care about anyone thinks - including myself - about style preferences.

HippoBaro:
Hi @sitano 

I'm all in for more linters but I wonder why a Makefile is necessary. We can run these tools directly with cargo, which seems like a more consistent way to interact with the tooling ecosystem of rust. What do we gain from it?

glommer:
It's a fair point.

Ivan, because this falls more into @HippoBaro 's turf I am yielding to him entirely. I think the linters are a good step and I am fine with whatever you guys come up with.

My only ask is that you fix the issues that the linter finds. If it is known broken nobody will use it, and if nobody will use it it doesn't make sense to have.

sitano:
> I'm all in for more linters but I wonder why a Makefile is necessary. We can run these tools directly with cargo, which seems like a more consistent way to interact with the tooling ecosystem of rust. What do we gain from it?

I of course was not sure you would like the idea of having a Makefile that calls Cargo. Usually calling those tools require some flags building so that its not very convenient to run them directly. Also with make or script its convenient that you have predefined set of actions and it lowers the bar for new comers as they see standard flows targets.

Maybe at this point it may be called overengineering. I basically wanted to draw some attention to the linter. I am going to address its warnings of course and provide a configuration that will allow having returns.

sitano:
> But please do follow up with a patch that fix the complaints that the tool raised.

Yes, I thought that merging it first and then fixing warnings separatelly totally makes sense.

sitano:
> For posterity, I like return statements and I have been trying to use them. That is one of the top complaints the tool raised.

I will provide configuration that allows that.

> However no matter how much I like a particular style, my hatred for style related discussions trumps it all.
If we can mechanically enforce a style that is well accepted in the community, I don't care about anyone thinks - including myself - about style preferences.

Using rustfmt must do the job. That is another good point of having make or some scripts.

sitano:
> My only ask is that you fix the issues that the linter finds. If it is known broken nobody will use it, and if nobody will use it it doesn't make sense to have.

Will do.

sitano:
If the makefile is not wanted at this point feel free to close the thing. I will check if clippy supports config file.

Daniel-B-Smith:
I would like to second the 'go with the known style' point. Expressions instead of explicit returns are perhaps a little surprising coming from other languages, but seeing the explicit returns would be somewhat jarring to someone who reads a lot of Rust code.

penberg:
FWIW, I see other projects also use a top-level `Makefile`. For example, Wasmer comes to mind: https://github.com/wasmerio/wasmer. It does seem to make Rust builds slightly simpler, so perhaps it's worth considering.

glommer:
Interested in hearing @Daniel-B-Smith 's opinion on the Makefile thing.

Here's the context: Daniel came up with a way to enforce the rule that borrows on unsafe cells don't cross suspension points. He is doing that in the clippy linter.

It would be nice to enforce all of that locally in an easy manner. CI is good, but it shortens the cycle if you don't have to wait for the CI run to see that your code is broken.

If there is a way to do all of that with Cargo I am all for it.
But we may want to have a script, Makefile, or whatever to make sure that the project is built using our linters.

HippoBaro:
I suggest we separate the issue in two. Let's use this one to decide what linter to use and with what configuration (+ the associated code fixes). In another we can setup CI to use the linter and decide if we want to bundle multiple tools into a script. 

If we want to encourage contributors to use these tools before committing, we should document how to create a pre-commit hook.

Daniel-B-Smith:
I did a quick audit of crates and projects I've used/investigated*, and the only repos with a makefile or install script were those with explicit C++ dependencies (grpcio and tikv). For the Rust project I set up CI for internally, we haven't had any need for makefiles. We can (and probably should) enable the Clippy lint @glommer is mentioning in the code. It is much harder to compile Rust for WASM than it is for Linux, so it doesn't shock me that WASM focused crates would end up with a more complicated build system than others. Almost any configuration we need can be done at the crate root `lib.rs` or in Cargo.toml. I don't think there is a ton of cost to using a makefile, and my sample is pretty biased towards a couple of types of projects. But my opinion would be to just configure everything in Cargo until we explicitly hit something that requires extra scripting.

As a related aside, I only saw two projects that had `clippy.toml` annotations. Hashbrown adds some identifiers to `doc-valid-idents` to use some camel case names in documentation, and Prost configures the maximum number of function args to 8 from the default of 7.

\* tokio, async-std, smol, parking lot, hashbrown, rayon, tikv, grpcio, prost, hyper, tonic, ripgrep, and fd.

glommer:
Closing based on the discussion.
I definitely think it is nicer to have this coming from cargo and or commit hooks

PR #14, opened 2020-09-04
sqe: always fallback to allocation of a posix buffer

glommer:
We don't allocate a buffer right now, passing an empty closure
if we are not in the poll ring. That's because I/O only goes through
the poll ring.

However we'd like to relax those restrictions soon. In preparation
for that, everybody passes a buffer. Later, when we implement uring
buffers, the poll ring will be able to pass registered buffers.

### Checklist


HippoBaro:
This is going to make falling back to the sleepable queue easier to implement.

PR #17, opened 2020-09-04
add unit tests to basic file io APIs

HippoBaro:
### What does this PR do?

Add unit test for all basic file APIs (create, open, close, rename,
allocate, read, write).
Scipio currently only supports NVMe-backed volumes formatted with XFS or
EXT4. We therefore can't use the system-provided tmpfs for our test so
instead we let the user decide what directory scipio should use to host
tests.

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -446,28 +446,268 @@ impl DmaFile {
     }
 }
 
+#[cfg(test)]
+fn make_test_directory(test_name: &str) -> PathBuf {
+    // Scipio currently only supports NVMe-backed volumes formatted with XFS or EXT4.
+    // We therefore can't use the system-provided tmpfs for our test so instead we let the user
+    // decide what directory scipio should use to perform integration tests.
+    // For more information regarding this limitation, see the README
+    match std::env::var("SCIPIO_TESTROOTDIR") {
+        Err(_) => {
+            eprintln!(
+                "Scipio currently only supports NVMe-backed volumes formatted with XFS \
+                or EXT4. To run io-related tests, please set SCIPIO_TESTROOTDIR to a NVMe-backed \
+                directory path in your environment."
+            );
+            panic!("missing SCIPIO_TESTROOTDIR environment variable")
+        }
+        Ok(path) => {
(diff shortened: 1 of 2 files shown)


glommer:
General comment as I tried to run this,:

```
warning: unused variable: `file`
   --> scipio/src/dma_file.rs:664:9
    |
664 |     let file = std::fs::File::create(root_path.join("testfile")).expect("failed to create file");
    |         ^^^^ help: if this is intentional, prefix it with an underscore: `_file`
    |
    = note: `#[warn(unused_variables)]` on by default
```

Let's keep this warn free.

I am still seeing two failures with this. I will investigate on my side too but let's hold merge until then

scipio/src/dma_file.rs
```diff
@@ -466,3 +491,182 @@ fn fallback_drop_closes_the_file() {
         });
     }
 }
+
+#[test]
+fn file_create_close() {
+    let ex = LocalExecutor::new(None).unwrap();
```
glommer, line 497: take a look at the `test_executor` macro in lib.rs. It is supposed to make this a little bit more ergonomic.
While that's not a big deal, in this case you have your own helper function to create the directory that youy always call. So you can wrap that into a macro that lives in this file and don't have to repeat yourself.

Not a blocker

glommer (changes requested):

scipio/src/sys/posix_buffers.rs
```diff
@@ -58,6 +58,12 @@ impl PosixDmaBuffer {
     pub fn as_ptr(&self) -> *const u8 {
         unsafe { self.data.add(self.trim) }
     }
+
+    pub fn memset(&self, value: u8) {
+        for i in self.as_mut_bytes() {
```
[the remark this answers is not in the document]
HippoBaro: Fixed in 77edd96

scipio/src/dma_file.rs
HippoBaro: Fixed in 8981a5a

glommer (approved):

PR #21, opened 2020-09-05
transparent support for tmpfs file io

HippoBaro:
### What does this PR do?

This PR adds support for file io on files that don't support `O_DIRECT`.

It uses a `SourceFd` union of `RawFd`s in the file APIs to indicate to
the Reactor whether a file supports poll operations on the `poll_ring`.
`SourceFd` is trivially convertible to a `RawFd`.

This detection is performed when a file is first created or opened.

For now we support two configurations:
* either the fd can be open with `O_DIRECT` and we assume it is pollable
* or it is not pollable at all and will use the `main_ring`

This excludes some categories of device like HDDs (compatible with
`O_DIRECT` but not with polling). For now (and most probably, ever) we
ignore this case.

Also, this PR changes the way we unit test file io in the following ways:

We now run the unit tests on both tmpfs and, optionally, on a user-provided
directory that supports poll mode io_uring interactions. If none is
provided via the `SCIPIO_TEST_POLLIO_ROOTDIR` environment variable, a
message is printed on stderr, but the tests don't fail.

I modified all tests to run twice and included a mechanism for them to
change their assertions depending on the current target directory. This
allows `file_allocate` to expect success for non `IO_DIRECT` files and
expect a failure otherwise.

I am not certain this is the right approach in practice though. It may be
simpler to just duplicate all tests.

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -306,7 +306,7 @@ impl DmaFile {
         }
 
         let mut f = enhance!(res, "Opening", Some(&path), None)?;
-        f.o_direct_alignment = 4096;
+        f.o_direct_alignment = 512;
         Ok(f)
     }
 
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -186,6 +186,7 @@ pub struct DmaFile {
     // facilitate error displaying.
     path: Option<PathBuf>,
     o_direct_alignment: u64,
+    pollable: bool,
 }
 
 impl DmaFile {
@@ -227,6 +228,7 @@ impl Default for DmaFile {
             file: unsafe { std::fs::File::from_raw_fd(-1) },
             path: None,
             o_direct_alignment: 4096,
+            pollable: true,
         }
     }
 }
@@ -250,6 +252,7 @@ impl DmaFile {
         path: &Path,
         flags: libc::c_int,
         mode: libc::c_int,
+        pollable: bool,
     ) -> io::Result<DmaFile> {
         let source = Reactor::get().open_at(dir, path, flags, mode);
         let fd = source.collect_rw().await?;
@@ -257,6 +260,7 @@ impl DmaFile {
             file: unsafe { std::fs::File::from_raw_fd(fd as _) },
             path: Some(path.to_path_buf()),
             o_direct_alignment: 4096,
+            pollable,
         })
     }
 
@@ -274,12 +278,12 @@ impl DmaFile {
         // try to open the file with O_DIRECT if the underlying media supports it
         let flags =
             libc::O_DIRECT | libc::O_CLOEXEC | libc::O_CREAT | libc::O_TRUNC | libc::O_WRONLY;
-        let mut res = DmaFile::open_at(-1 as _, &path, flags, 0o644).await;
+        let mut res = DmaFile::open_at(-1 as _, &path, flags, 0o644, true).await;
 
         if res.is_err() {
             // we failed to open the file, try again without O_DIRECT
             let flags = flags & !libc::O_DIRECT;
-            res = DmaFile::open_at(-1 as _, &path, flags, 0o644).await;
+            res = DmaFile::open_at(-1 as _, &path, flags, 0o644, false).await;
         }
 
         let mut f = enhance!(res, "Creating", Some(&path), None)?;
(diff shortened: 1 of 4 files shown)


scipio/src/dma_file.rs
```diff
@@ -271,31 +271,39 @@ impl DmaFile {
...
-            Some(&path),
-            None
-        )?;
-        f.o_direct_alignment = 512;
-        Ok(f)
+        let mut res = DmaFile::open_at(-1 as _, &path, flags, 0o644).await;
+
+        if res.is_err() {
+            // we failed to open the file, try again without O_DIRECT
+            let flags = libc::O_CLOEXEC | libc::O_RDONLY;
```
glommer, line 329: same comment as per the write flags : best to keep the initial variable and just remove the flag we want to do away with.

scipio/src/dma_file.rs
```diff
@@ -70,6 +71,10 @@ fn align_down(v: u64, align: u64) -> u64 {
     v & !(align - 1)
 }
 
+pub trait AsSourceFd {
```
glommer, line 74: I don't quite like this direction.

RawFd is very central to rust and used everywhere.

It is probably better to encode this information in the `SourceType`. We do that already for other things in the system and would have the added benefit of consistency.

See for instance how we can encode a buffer inside the SourceType for DmaRead.

Another option that would be acceptable would be to use a special `SourceType`. So we would have for instance `DmaRead` and `PollableDmaRead` Both directions would be acceptable so long as we keep RawFd untouched

scipio/src/sys/uring.rs
```diff
@@ -490,7 +490,10 @@ macro_rules! queue_request_into_ring {
 
 macro_rules! queue_storage_io_request {
     ($self:expr, $source:ident, $op:expr) => {{
-        queue_request_into_ring!($self.poll_ring, $source, $op)
+        match $source.raw {
```
glommer, line 493: This is good. If you take my suggestion to encode pollable in the Source (either a new `SourceType` or a boolean inside the existing `SourceType`) you can just match here by type instead of raw.

The overall direction is solid, just the specific placement (rawfd vs sourcetype) is not.

scipio/src/dma_file.rs
```diff
@@ -484,19 +484,25 @@ impl DmaFile {
...
+    // We therefore let the user decide what directory scipio should use to host the unit tests in.
     // For more information regarding this limitation, see the README
-    match std::env::var("SCIPIO_TESTROOTDIR") {
+    match std::env::var("SCIPIO_TEST_POLLIO_ROOTDIR") {
         Err(_) => {
             eprintln!(
                 "Scipio currently only supports NVMe-backed volumes formatted with XFS \
-                or EXT4. To run io-related tests, please set SCIPIO_TESTROOTDIR to a NVMe-backed \
-                directory path in your environment."
+                or EXT4. To run io-related tests, please set SCIPIO_TEST_POLLIO_ROOTDIR to a \
```
glommer, line 503: I think I am okay with that.

We can reconsider later. Is the message visible enough? `cargo test` usually doesn't show output unless you specify `--nocapture` My fear is that in practice nobody but us will run this, and gosh, even we can forget.

glommer (changes requested):
Hippo, the Raw vs Source thing has to be changed. This is the kind of change for which a history rewrite at the end is beneficial. I'd advise not doing it just yet - let's wait to get everything sorted first.

The rest is okay, minor suggestions for improvement but otherwise good

HippoBaro:
New version differs too much from the original so I rewrote history. 

New version doesn't use the `SourceFd` anymore but rather parameterize
the existing `SourceType`.
It adds a boolean state to the `DmaRead` and `DmaWrite` `SourceType`s
to indicate to the reactor if the operation can be submitted on the poll
ring or should stay on the main one.

This solution is admittedly much simpler and cleaner.

scipio/src/dma_file.rs
HippoBaro: This approach has the main advantage that every test run on both implementations "for free" and it'll be easy for new tests to do so as well.
The big caveat is as you said the risk that end up never running the poll mode tests which are really important. `cargo test` does indeed swallow the message which is a big issue in this case. 

Not sure what the right choice is.

glommer: Ok. I spent some time thinking about this.

Let's try this way.

It would be nice if we could add least add this to a file people will read. This feels like a job for a getting started file, but for now I think the readme is enough. We also need to make sure that our CI always have this set.

glommer (approved):

scipio/src/dma_file.rs
```diff
@@ -271,14 +271,18 @@ impl DmaFile {
...
             libc::O_DIRECT | libc::O_CLOEXEC | libc::O_CREAT | libc::O_TRUNC | libc::O_WRONLY;
-        let mut f = enhance!(
-            DmaFile::open_at(-1 as _, &path, flags, 0o644).await,
-            "Creating",
-            Some(&path),
-            None
-        )?;
+        let mut res = DmaFile::open_at(-1 as _, &path, flags, 0o644).await;
+
+        if res.is_err() {
```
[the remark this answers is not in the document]
glommer: @HippoBaro I approved the PR but this one needs to change. You can put that patch on top.

The boolean thing is not important so not a blocker.

PR #26, opened 2020-09-07
Improve executor ergonomics

glommer:
### What does this PR do?

This PR greatly improves the ergonomics of creating and dealing with executors

### Motivation

It really sucked before.

### Checklist


HippoBaro (approved):

scipio/src/executor.rs
```diff
@@ -322,28 +352,99 @@ impl LocalExecutor {
...
+    /// use scipio::LocalExecutor;
+    ///
+    /// // executor is a single thread, but not bound to any particular CPU.
+    /// LocalExecutor::spawn_new("myname", None, async move {
+    ///     println!("hello");
+    /// });
+    ///
+    /// LocalExecutor::wait_on_executors();
+    /// ```
+    pub fn spawn_new<F, T>(name: &'static str, binding: Option<usize>, fut: F)
```
HippoBaro, line 388: It is very surprising and counter intuitive that a `spawn_new` function would not return.

scipio/src/executor.rs
```diff
@@ -322,28 +352,99 @@ impl LocalExecutor {
...
+    /// # Examples
+    /// use scipio::LocalExecutor;
+    ///
+    /// // executor is a single thread, but not bound to any particular CPU.
+    /// LocalExecutor::spawn_new("myname", None, async move {
+    ///     println!("hello");
+    /// });
+    ///
+    /// LocalExecutor::wait_on_executors();
+    pub fn wait_on_executors() {
```
HippoBaro, line 430: Seems like a nice utility but I see a couple of problems with it:

- people know how to use threads and join them. This function + the void `spawn_new` make something simple complicated (at least, specific to this library);
- it prevents you from from stopping only part of your executors.

scipio/src/executor.rs
```diff
@@ -322,28 +352,99 @@ impl LocalExecutor {
...
+    ///
+    /// # Examples
+    /// ```
+    /// use scipio::LocalExecutor;
+    ///
+    /// let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
+    /// println!("My ID: {}", local_ex.id());
+    /// ```
+    pub fn id(&self) -> usize {
+        self.id
```
HippoBaro, line 447: Since this id is public, it would make sense to use it to decorate panic messages. Probably not the scope of this PR but nice to keep in mind.

scipio/src/executor.rs
```diff
@@ -717,6 +818,34 @@ impl<T> Task<T> {
...
+    /// ```
+    /// use scipio::{LocalExecutor, Task};
+    ///
+    /// let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
+    ///
+    /// local_ex.run(async {
+    ///     println!("my ID: {}", Task::<()>::id());
+    /// });
+    /// ```
+    pub fn id() -> usize
```
HippoBaro, line 838: Having `later()` or `local()` be proxies to the `LocalExecutor` is fine since those two methods either deal with or return a new `Task`. This one is just confusing. 

I think we should have a way to access the local executor (ala `engine()`). Those utility functions, in turn, should call into that.

HippoBaro (approved):
Nice QOL improvement. however I don't like the fact that `LocalExecutor::spawn_new` stores the thread handle internally. I think it would be simpler to just return a handle to the newly created thread to the user so they can `join()` ant their leisure. That would allow you to get rid of `SPAWNED_EXECUTORS` and `wait_on_executors` as they both seem superfluous to me.

Couple of other comments, other than that, this is good to go.

scipio/examples/hello_world.rs
```diff
@@ -0,0 +1,43 @@
...
+        }));
+    }
+    join_all(tasks).await;
+}
+
+fn main() {
+    // There are two ways to create an executor, demonstrated in this example.
+    //
+    // We can create it in the current thread, and run it separately later...
+    let ex = LocalExecutor::new(Some(0)).unwrap();
```
penberg, line 25: `unwrap()` in examples is unfortunate, because you end up under-designing API error handling. This should be trivial to fix to return the error, no?

scipio/examples/hello_world.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    // Or we can spawn a new thread with an executor inside.
+    LocalExecutor::spawn_new("hello", Some(1), async move {
```
penberg, line 28: The _new_ in `spawn_new` looks odd because it's implied by the verb. How about just `spawn`? Or `spawn_task`, `spawn_thread`?

scipio/examples/hello_world.rs
sitano: Internals has the following code example:

```rust
let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
local_ex.run(async {
     Timer::new(std::time::Duration::from_millis(100)).await;
     println!("Hello world!");
});
```

scipio/src/executor.rs
sitano: yet it's interesting what a client would expect to be returned as far as it produces both: a thread, an executor and a fiber.

scipio/examples/hello_world.rs
sitano: Naming is unfortunate here. There is already a `spawn` that pushes a fiber on existing executor:

```rust
    pub fn spawn<T: 'static>(&self, future: impl Future<Output = T> + 'static) -> Task<T> {
```

scipio/src/executor.rs
sitano: The thing somewhat recalls:

```rust
        LocalExecutor::new(None)
            .and_then(|x: LocalExecutor| -> _ { Ok(x.run(async { println!("Hello") })) })
            .expect("good")
    )
```

scipio/src/executor.rs
glommer: The problem is that in Rust, you can only safely use thread local things inside a scope. You cannot have an `engine()` function like seastar. At most you could have `with_engine(Function)`. (see the uses of `with`). Between those two things I prefer to provide individual functions.

scipio/src/executor.rs
glommer: The reason I didn't do it this way is that it forces you to have a vector to store the handles. In the search for ergonomics that's one more thing to care about.

It's not the end of the world. I will consider it.

scipio/examples/hello_world.rs
glommer: What do you propose? In examples usually you don't have context do do any error handling. I see a lot of other libraries using unwrap.

scipio/examples/hello_world.rs
glommer: I agree `spawn_new` is not ideal but there are many other executor related functions called `spawn` and its variants.
I don't like `spawn_task` because this spawns an executor, not a a task.

How about we just go full verbose then and call it `spawn_executor` ?

scipio/src/executor.rs
glommer: The only reasonable thing to return is a join handle.

Actually my preference would be to return an object that contains the join handle, and joins on it when it goes out of scope.

However I tested that approach. And it seems like the object can (and usually will) go out of scope before the end of the block, as soon as Rust deems it will not only be used. So that makes it essentially impossible to fire executors in parallel

glommer: You are right that there is another way to write this Ivan, not like yours, a bit more complex (we are forcing a task queue, remember).

This is just a more ergonomic way

scipio/examples/hello_world.rs
glommer: expect is just unwrap() with an error message. Both panics, and I don't consider either to be proper error handling.
Unwrap is just less verbose.

If there is a preference towards `expect` I can use it, but I don't see a big difference

scipio/examples/hello_world.rs
penberg: `spawn_executor` seems fine to me.

scipio/examples/hello_world.rs
penberg: I was not arguing for `expect()` either -- I was arguing for _returning_ the error from `main()` -- e.g. `fn main() -> Result<(), std::io::Error>`.

glommer: would just using `?` suffice ?

I think we lack context to do anything more complex than that in an example. 
(I am also not sure we can do `?` in examples, but assuming we can)

penberg: Yup, that was my suggestion: change the return type of `main` and propagate the error with `?`.

glommer: <facepalm>

you are talking about the hello world example.

I thought you were talking about the examples on top of the functions. Sorry, that's my bad

glommer:
@penberg @HippoBaro I have pushed two commits that fixes those issues.
Once you guys are cool with them I will fold them into the previous patches.

scipio/src/executor.rs
glommer: :+1:

scipio/src/executor.rs
glommer: changed.

scipio/examples/hello_world.rs
```diff
@@ -0,0 +1,44 @@
...
+    // If you create the executor manually, you have to run it like so.
+    //
+    // spawn_new() is the preferred way to create an executor!
+    ex.run(async move {
+        hello().await;
+    });
+
+    // The newly spawned executor runs on a thread, so we need to join on
+    // its handle so we can wait for it to finish
+    handle.join().unwrap();
```
sitano, line 42: This is exactly why making main() -> std::io::Result instead of expect() was a bad idea:

```
  --> scipio/examples/hello_world.rs:42:18
   |
22 | fn main() -> Result<()> {
   |              ---------- expected `std::io::Error` because of this
...
42 |     handle.join()?;
   |                  ^ the trait `std::convert::From<std::boxed::Box<dyn std::any::Any + std::marker::Send>>` is not implemented for `std::io::Error`
```

now the code has inconsistent error handling so that even main return is useless because all other errors can't fit its type.

sitano: btw, the fact Rust has many incompatible variants of Result<> type is painful. Maybe its worth having your own Result<> and return everywhere 1 type and make it compatible with at least 1 std (std::io ie).

glommer: This is only a problem (and apparently one known to the rust community) with `.join()`
If you google this error message it should take you to places where the Rust folks tell you not to `?` on join.
(although now that I am searching again I have been so far unable).

You will get similar results if you spawn a thread manually and then `join` it, so I am not too concerned.

My understanding is that it would be possible to make this better and `?` on join if we implemented an error wrapper on `io::Error` that is send, but I am not too concerned about that either: if a thread fails in an unrecoverable way (at least for us), you are likely to want to panic the whole thing anyway: otherwise you are left with executors that are running while others have failed.

I feel like the example is good now: The user can see how to handle errors that happen at spawn time, and for asynchronous errors coming from threads we are recommending unwrap.

glommer: I agree about the errors, but in practice it seems to me that everyone standardized on `io::Result` so as long as your errors are convertible to and from it, all should be good

glommer:
FYI: I wrote a slightly more complex example and I am having issues with the requirement of `Send + Sync` that I imposed in the future passed to spawn local. I will investigate the best way to solve it, but may delay this.

There is no rush

glommer:
Requiring `Send + Sync` at the interface levels is causing Rust to require the market for stuff that we use deep down the engine, which is not acceptable:

```
   ::: /home/glauber/scipio/scipio/src/executor.rs:387:12
    |
387 |         F: Send + 'static,
    |            ---- required by this bound in `scipio::LocalExecutor::spawn_executor`
    |
    = help: within `scipio::sys::Source`, the trait `std::marker::Sync` is not implemented for `*mut u8`
```

scipio/examples/hello_world.rs
glommer: Hi Daniel - as we discussed, this won't work in practice because `JoinHandle` transforms the error into a boxed `std::any::Any` whose size is not known.

I don't know of a very good way to `?` from a join, but this is a problem that `thread::spawn` has already. We can fix the example when we find a good way (I'd like to), but tabling for now so we can make progress

glommer:
Rewrote history into https://github.com/DataDog/scipio/pull/27
(empty diff from the current HEAD here)

I feel like the version we have now address most of the comments, and the few points left can be done separately.

Thank you all!

PR #29, opened 2020-09-07
detect a file preferred alignment upon opening

HippoBaro:
Use `sysfs` to fetch the underlying device preferred alignment, with a
fallback of 512b for reads and 4KiB for writes.
This implementation has the major drawback that it blocks as it uses
standard IO routines (missing `readlink` in our implementation).

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -290,12 +290,13 @@ impl DmaFile {
         use std::fs;
         use std::path;
 
-        // dev major + minor gives enough information about te file to determine physical location
+        // dev major + minor is enough to determine physical location
         let stats = self.statx().await?;
         let major = stats.stx_dev_major;
         let minor = stats.stx_dev_minor;
 
         // /sys/dev/block/major:minor is a symlink to the device in /sys/devices/
+        // if the file doesn't exist, we are opening a tempfs-located file without O_DIRECT
         let dir_path = format!("/sys/dev/block/{}:{}", major, minor);
         let mut dir = path::Path::new(dir_path.as_str()).to_path_buf();
         dir = dir.parent().unwrap().join(std::fs::read_link(&dir)?);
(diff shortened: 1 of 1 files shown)


HippoBaro:
@glommer I don't expect to merge as-is considering this will block the executor's thread every time a file is opened. Still looking for reviews though.

scipio/src/dma_file.rs
```diff
@@ -285,6 +285,48 @@ impl DmaFile {
         Reactor::get().alloc_dma_buffer(size)
     }
 
+    /// Returns the system's preferred alignment for the file
+    async fn fetch_device_alignment(&self) -> io::Result<u64> {
+        use std::fs;
+        use std::path;
+
+        // dev major + minor gives enough information about te file to determine physical location
```
glommer: the file

scipio/src/dma_file.rs
```diff
@@ -285,6 +285,48 @@ impl DmaFile {
...
+        let minor = stats.stx_dev_minor;
+
+        // /sys/dev/block/major:minor is a symlink to the device in /sys/devices/
+        let dir_path = format!("/sys/dev/block/{}:{}", major, minor);
+        let mut dir = path::Path::new(dir_path.as_str()).to_path_buf();
+        dir = dir.parent().unwrap().join(std::fs::read_link(&dir)?);
+
+        // if the device directory is a partition (contains a file named "partition"),
+        // we navigate to its parent directory, the host device
+        let device = if dir.join("/partition").exists() {
```
glommer, line 306: This is ok but a bit wasteful or every open.

It would be better if we could parse sysfs entirely and return a struct that contains all its fields. It is totally fine if the only field we really parse is the one you need, since from that point on adding new fields becomes a mechanical task. For files that return an integer we can implement all of them by just passing their names though.

Once you have that you can store that structure in a hashmap by device id an avoid the parsing.

Long term I wanted to have an API that opens a file in a directory because that will allow us to use uring's registered fds. In that case we can do even better and store the result in the directory - not even needing the stat

scipio/src/dma_file.rs
```diff
@@ -285,6 +285,48 @@ impl DmaFile {
...
+        use std::fs;
+        use std::path;
+
+        // dev major + minor gives enough information about te file to determine physical location
+        let stats = self.statx().await?;
+        let major = stats.stx_dev_major;
+        let minor = stats.stx_dev_minor;
+
+        // /sys/dev/block/major:minor is a symlink to the device in /sys/devices/
+        let dir_path = format!("/sys/dev/block/{}:{}", major, minor);
```
glommer, line 300: Please note that the paths are different if you are dealing with an lvm or raid device. Take a look at how the seastar implementation of `iotune` handles that. If we discover that the device is an array we need to jump from there to its constituent devices

glommer (changes requested):

glommer:
@HippoBaro This will not block the executor. Those files are all in memory.
It's a good direction. See my comments for something you missed (arrays) and how we can improve the situation further (hashmap)

scipio/src/dma_file.rs
zserik: https://github.com/scylladb/seastar/blob/master/apps/iotune/iotune.cc#L109 ?

glommer: Yes but note that this PR is a bit outdated.

After it was started, I have merged support for parsing sysfs for other reasons. It shouldn't be too hard at this point to do the rest.

Cu2y (approved):

PR #35, opened 2020-09-08
simplify capture of clone vars

sitano:
Instead of doing 

```
let x = Rc<...>(...);
let f = || x.clone();
let f2 = f.clone()
```

It can be done with anonymous function like this:

```
let second = local_ex
            .spawn_into(
                (|f: Rc<RefCell<bool>>, s: Rc<RefCell<bool>>| {
                    async move {
                       ...
                    }
                })(first_started.clone(), second_status.clone()),
                tq2,
            )
            .unwrap();

```

I am not happy Rust can't derive type information for (f, s) himself so it's required to be stated it explicitly as in `|f: Rc<RefCell<bool>>, s: Rc<RefCell<bool>>|` instead of `|f, s|`.

However, there is a popular macro `enclose!` that does the job.

```
  enclose! { (first_started, second_status)
                    async move {
                  ...
                    }
}
```

### DIFF
--- a/scipio/src/executor.rs
+++ b/scipio/src/executor.rs
@@ -1058,25 +1058,23 @@ fn task_optimized_for_throughput() {
         let first_started = Rc::new(RefCell::new(false));
         let second_status = Rc::new(RefCell::new(false));
 
-        let rust_capture_sucks = move || (first_started.clone(), second_status.clone());
-        let clone_sucks = rust_capture_sucks.clone();
-
         let first = local_ex
             .spawn_into(
-                async move {
-                    let (first_started, second_status) = clone_sucks();
-                    *(first_started.borrow_mut()) = true;
+                enclose! { (first_started, second_status)
+                    async move {
+                        *(first_started.borrow_mut()) = true;
 
-                    let start = Instant::now();
-                    // Now busy loop and make sure that we yield when we have too.
-                    loop {
-                        if *(second_status.borrow()) {
-                            panic!("I was preempted but should not have been");
+                        let start = Instant::now();
+                        // Now busy loop and make sure that we yield when we have too.
+                        loop {
+                            if *(second_status.borrow()) {
+                                panic!("I was preempted but should not have been");
+                            }
+                            if start.elapsed().as_millis() > 200 {
+                                break;
+                            }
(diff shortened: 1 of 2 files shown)


sitano:
@glommer do you like approach with `enclose!`? if not pls close)

glommer:
Hi Ivan

In general I like the approach of using macros to achieve that.
In fact, for tests, I had coded a similar macro. Take a look at `make_shared_var`, `update_cond` and friends in `src/lib.rs`.

I have a couple of questions about your approach specifically:
1. There is a crate called `enclose` apparently. It is MIT licensed, so we could use it. What made you code your own macro?
2. Why are you making this available for tests only? If we do something like that I'd like to export the macro. This pattern shows up everywhere, not just in testing. Granted when I coded my macros I did them test only, but that's because I felt they were not good enough to be generic (And for tests we have them already).

sitano:
@glommer so you generally like the change?

> Hi Ivan
> 
> In general I like the approach of using macros to achieve that.
> In fact, for tests, I had coded a similar macro. Take a look at `make_shared_var`, `update_cond` and friends in `src/lib.rs`.
> 
> I have a couple of questions about your approach specifically:
> 
> 1. There is a crate called `enclose` apparently. It is MIT licensed, so we could use it. What made you code your own macro?

yes. I saw that one. I just thought why we would want to add another dependency if the only thing we require is a macro of 5 lines of code. wdyt about it?

> 2. Why are you making this available for tests only? If we do something like that I'd like to export the macro. This pattern shows up everywhere, not just in testing. Granted when I coded my macros I did them test only, but that's because I felt they were not good enough to be generic (And for tests we have them already).

I didn't yet changed all the rest of the code. But my approach is always to start with the smallest scope and then just extended if its needed somewhere.

If you like overall the idea of that macros I can rework all closures and tests to simplify them.

glommer:
I do like the change, I was surprised that you closed the PR!

About using our own macro instead of adding the crate - it makes sense and I agree. Just please mention for posterity in the changelog that this is the reason.

I want to see that exported and usable for projects consuming scipio by the time of the merge.

But good direction overall!

sitano:
Ok. Cool!

PR #36, opened 2020-09-08
add align funcs doc and inline

sitano:


sitano:
I am not sure about inline tho.

scipio/src/dma_file.rs
```diff
@@ -62,10 +62,23 @@ macro_rules! bad_buffer {
     }};
 }
 
+#[inline(always)]
+/// Align value `v` up where align = 2^n.
```
glommer, line 66: I don't like the fact that you say `align = 2^n`. It may give the reader the impression that you are to pass `n`.
I prefer that we spell this out in plain English, `align must be a power of two`

scipio/src/dma_file.rs
```diff
@@ -62,10 +62,23 @@ macro_rules! bad_buffer {
...
+///     v % align == 0 => v
+///     v % align != 0 => (1+floor(v/align))*align
 fn align_up(v: u64, align: u64) -> u64 {
     (v + align - 1) & !(align - 1)
 }
 
+#[inline(always)]
+/// Align value `v` down where align = 2^n.
+///
+/// v = 0b111, align     = 0b010
```
glommer, line 78: This seems to be describing the process through which we align.

I'd prefer the documentation to be a bit more black-box like. Give examples, preferably with hexadecimals or decimals which are more common.

How we implement this is of no concern o the user, just the expected results.

glommer (changes requested):
Thanks Ivan, this is good! However requesting a couple of minor changes.

Because this is a single patch, please fold the patch and force-push

glommer:
@sitano I'd expect the compiler to be smart enough to inline this even if we don't ask for it.
I don't mind including but unless I am mistaken about the above I don't see the need

PR #38, opened 2020-09-08
Timeraction

glommer:
### What does this PR do?

The TimerAction provides an ergonomic way to execute actions in the future. It is a simplified way to wait
on a future after a Timer expires

### Checklist

### DIFF
--- a/scipio/src/timer.rs
+++ b/scipio/src/timer.rs
@@ -4,6 +4,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
 use crate::parking::Reactor;
+use crate::task::JoinHandle;
+use crate::{Local, QueueNotFoundError, Task, TaskQueueHandle};
 use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll, Waker};
@@ -12,6 +14,10 @@ use std::time::{Duration, Instant};
 /// A timer that expires after a duration of time.
 ///
 /// Timers are futures that output the [`Instant`] at which they fired.
+/// Note that because of that, Timers always block the current task queue
+/// in which they currently execute.
+///
+/// In most situations you will want to use [`TimerAction`]
 ///
 /// # Examples
 ///
@@ -31,6 +37,7 @@ use std::time::{Duration, Instant};
 ///     sleep(Duration::from_millis(100)).await;
 /// });
 /// ```
+/// [`TimerAction`]: struct.TimerAction
 #[derive(Debug)]
 pub struct Timer {
     /// This timer's ID and last waker that polled it.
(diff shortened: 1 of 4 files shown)


scipio/src/timer.rs
```diff
@@ -133,6 +140,179 @@ impl Future for Timer {
...
+            action.await
+        });
+
+        TimerAction {
+            handle: task.detach(),
+        }
+    }
+
+    /// Creates a [`TimerAction`] that will execute the associated future once after some
+    /// time is passed
```
HippoBaro, line 235: The documentation doesn't make it clear that the timer repeat itself

scipio/src/timer.rs
```diff
@@ -133,6 +140,179 @@ impl Future for Timer {
...
+    ///     let action = TimerAction::once(Duration::from_millis(100), async move {
+    ///         println!("Execute this in 100ms");
+    ///     });
+    ///     action.join().await;
+    /// });
+    /// ```
+    /// [`TimerAction`]: struct.TimerAction
+    pub async fn join(self) {
+        self.handle.await;
+    }
```
HippoBaro, line 313: How does that work with repeated timers? Does it returns it the timer is cancelled in another fiber?

HippoBaro (changes requested):
Documentation on `periodic` needs to be updated, otherwise all good

scipio/src/timer.rs
glommer: this should not be used with repeated timers. The problem is that we can't (due to way rust works) both `cancel` and `join`: `cancel` takes a `self` parameter, not a reference, so after you call it you cannot do anything with the `JoinHandle` anymore.

We could change it like we changed `detach` but that is a bit more complex because `cancel` goes all the way down to the raw task.

One option I am thinking about is making periodic return a `Result`: the timer would not run forever in that case, but keep running until you return an error. This way we would be able to make it stop at som epoint

HippoBaro: Another option would be to require that the future passed in to the timer takes a reference to the timer itself so you have a chance to stop the timer once per interval.

HippoBaro: Yet another option that would make timers more general and useful would be to require that futures passed in return `Option<Duration/Instant>` so that the timer knows if the caller wants it to fire again. 

It would also make implementing jitter simpler.

glommer: I don't like your first suggestion, it feels like a layering violation.

Taking an action based on the return value is a lot cleaner in my opinion.
If I go this route I may also change the name from `periodic` to `repeat`.

Another thing that is missing btw on this submission is a way to fire into a specific task queue which is different than the caller task queue, saving one indirection.

glommer:
@HippoBaro Please let me know if you like this version better

HippoBaro:
> @HippoBaro Please let me know if you like this version better

I do! However could you rebase and rewrite history? Some commits have already been merged upstream (such as d8f591b)

HippoBaro (approved):

glommer:
Current state saved to https://github.com/glommer/scipio/tree/timeraction-approved

Force-pushed a new branch to this PR's `timeraction` branch

```
$ git diff glauber/timeraction-approved..glauber/timeraction
$ echo $?
0
```

merging

PR #41, opened 2020-09-09
test if io_uring support is new enough

glommer:
Do not merge yet: this depends on unmerged code to both liburing
and uring-sys


glommer:
@Daniel-B-Smith this should help you when evaluating kernel versions.

I need to merge code into liburing and uring-sys before we can merge this so hold tight

glommer:
patches merged to uring-sys and liburing

Unfortunately we have to keep carrying a diff from uring-sys because we can't update `liburing` right now inside uring-sys. But at least it is a difference that will be resolved in time.

Merging for the benefit of people getting less puzzling errors

PR #44, opened 2020-09-09
destroy Timer Action

glommer:
### What does this PR do?

This PR allows a TimerAction to be destroyed immediately, without having to spawn a new task for it.
To do that, we perform some cleanups in the timer infrastructure and remove yet another legacy Atomic.
Yay! Timers are starting to look good

### Checklist

### DIFF
--- a/scipio/src/timer.rs
+++ b/scipio/src/timer.rs
@@ -181,12 +181,13 @@ impl<T: 'static> TimerAction<T> {
     /// use scipio::{LocalExecutor, TimerAction};
     /// use std::time::Duration;
     ///
-    /// LocalExecutor::spawn_executor("test", None, || async move {
+    /// let handle = LocalExecutor::spawn_executor("test", None, || async move {
     ///     let action = TimerAction::once_in(Duration::from_millis(100), async move {
     ///         println!("Executed once");
     ///     });
     ///     action.join().await;
-    /// });
+    /// }).unwrap();
+    /// handle.join().unwrap();
     /// ```
     /// [`Duration`]: https://doc.rust-lang.org/std/time/struct.Duration.html
     /// [`TimerAction`]: struct.TimerAction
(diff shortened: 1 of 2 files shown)


glommer:
because review hasn't started, I forced pushed this branch adding more tests.
I am now confident that the implementation is sane.

HippoBaro (approved):

scipio/src/timer.rs
```diff
@@ -422,8 +429,37 @@ impl<T: 'static> TimerAction<T> {
...
+    ///         println!("Will not execute this");
+    ///     });
+    ///     action.destroy();
+    ///     action.join().await;
+    /// });
+    /// ```
+    /// [`TimerAction`]: struct.TimerAction
+    /// [`cancel`]: struct.TimerAction.html#method.cancel
+    /// [`join`]: struct.TimerAction.html#method.join
+    pub fn destroy(&self) {
```
HippoBaro, line 460: The distinction between `destroy` and `cancel` seems subtle. What's the difference between calling `destroy()` vs calling `cancel().detach()`? Both are non-blocking.
Maybe you want to make the distinction clearer and why I'd want one over the other.

glommer: The difference is indeed subtle. Behaviourally there is no difference so from a correctness point of view there is no difference.

From a performance point of view there is a difference:
`destroy()` does not generate a new task.
`cancel().detach()` does.

I thought I had mentioned that in the docs. What changes would you suggest to make that clearer ?

glommer: if you don't want to wait `destroy` is always preferable. So maybe we just need to say that explicitly ?

glommer:
I added a commit on top to fix the documentation because some of the examples were misleading as I have noticed later.
Because some of the examples were preexisting, that's a patch on top.
Merging this version

PR #45, opened 2020-09-10
Move examples to top-level "examples" directory

penberg:
### What does this PR do?

We introduced a Cargo workspace in pull request #12 so that we could
separate library and examples code. So let's move the echo and
hello_world examples to a top-level "examples" directory.

### Related issues

* Pull request #12 

### Additional Notes

Please note that cargo builds examples as part of "cargo test", not as
part of "cargo build".


glommer:
@HippoBaro please take a look at this one.

examples/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+socket2 = { version = "0.3.12", features = ["pair", "unix"] }
+iou = { git = "https://github.com/glommer/iou", tag = "poll-ring-works" }
+uring-sys = { git = "https://github.com/glommer/uring-sys", tag= "poll-ring-works" }
+nix = "0.16.0"
+aligned_alloc = "0.1"
+bitmaps = "2.1.0"
+typenum = "1.12"
+scoped-tls = "1.0.0"
+futures = "0.3.5"
+rlimit = "0.3.0"
```
HippoBaro, line 21: Aren't the dependencies transitive? importing `scipio` should import the rest, no?

HippoBaro (approved):

examples/Cargo.toml
penberg: @HippoBaro Yeah, cleaned up them. Had to import `futures`, though...

penberg:
> @penberg does that change the way examples are run?
> Currently `cargo run --example hello_world` will run the example

No changes, that same command still works.

HippoBaro:
Thanks for the PR! Having good, clean examples is super important :bow:. Merging this.

PR #46, opened 2020-09-10
fixes for semaphore

glommer:
### What does this PR do?

makes semaphore more stable, by not relying on unsafe features

### Motivation

When I started writing rust sometimes I just couldn't get stuff to work and ended up resorting to my old C++ ways.
in a "I was young and needed the money" kind of way.

Not only that, I knew the code would break eventually and it had scary fixmes. Guess what? It did.

Time to fix it.

### Checklist

### DIFF
--- a/scipio/src/local_semaphore.rs
+++ b/scipio/src/local_semaphore.rs
@@ -294,115 +294,117 @@ impl Semaphore {
     }
 }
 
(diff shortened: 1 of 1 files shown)


HippoBaro (approved):
LGTM

PR #47, opened 2020-09-10
Fix description of read_dma_aligned

espindola:
Unlike write_dma, read_dma_aligned doesn't take a buffer, it returns
one.

BTW, what are the restrictions on 'pos'? It has to be aligned no?


glommer (approved):

glommer:
Thanks.

Yes, pos has to be aligned for direct_io to work. Notice that we align it in `read_dma` (the non aligned version)

PR #48, opened 2020-09-11
Rearm TimerActions

glommer:
### What does this PR do?

Provides a way for us to rearm TimerAction
This is a common primitive found in many timer libraries.
As we prepare for that, it makes sense to split TimerAction in two, since it doesn't make sense to rearm a repeatable TimerAction

### Checklist

### DIFF
--- a/scipio/src/timer.rs
+++ b/scipio/src/timer.rs
@@ -476,6 +476,9 @@ impl<T: 'static> TimerAction<T> {
 
     /// Waits for a [`TimerAction`] to return
     ///
+    /// Returns an [`Option`] with value None if the task was canceled and Some if
+    /// the action finished successfuly
+    ///
     /// # Examples
     ///
     /// ```
@@ -491,8 +494,9 @@ impl<T: 'static> TimerAction<T> {
     /// handle.join().unwrap();
     /// ```
     /// [`TimerAction`]: struct.TimerAction
-    pub async fn join(self) {
-        self.handle.await;
+    /// [`Option`]: https://doc.rust-lang.org/std/option/enum.Option.html
+    pub async fn join(self) -> Option<T> {
+        self.handle.await
     }
 }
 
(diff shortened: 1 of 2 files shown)


scipio/src/timer.rs
```diff
@@ -6,18 +6,47 @@
...
+            Reactor::get().remove_timer(self.id);
+        }
+
+        // Update the timeout.
+        self.when = Instant::now() + dur;
+
+        if let Some(waker) = self.waker.as_mut() {
+            // Re-register the timer with the new timeout.
+            Reactor::get().insert_timer(self.id, self.when, waker);
+        }
```
HippoBaro, line 39: Can you explain why we update `self.when` even if we don't insert a new timer? Also, in what circunstances `self.waker` can be `None`?

glommer: I can, but notice please that this code is pre-existing, I am just moving here.

`self.waker` starts as None. It only gets assigned when the future is first polled() (call to `await`).

Imagine for instance that you schedule a function that will call `timer.await` for execution in a task queue that is very busy: it may take a long time for `waker` not to be `None`. There are no guarantees as to when that might happen.

We only need to go to the reactor on reset if the future was already polled and is pending (has a waker).
If the future was not polled, then all we have to do is update `when`: whenever it does get polled, it will be polled with the right `when`

In summary:
`when` has to be updated in all cases: 

* if the future was polled already, so we always have the most up2date information (waker will be `Some`, because we are waiting on it)
 * if the future was not polled, so we register the timer correctly for the right expiration (waker will be `None`)

scipio/src/timer.rs
```diff
@@ -283,48 +300,175 @@ impl<T: 'static> TimerAction<T> {
...
+    /// }).unwrap();
+    /// handle.join().unwrap();
+    /// ```
+    /// [`TimerActionOnce`]: struct.TimerActionOnce
+    /// [`Option`]: https://doc.rust-lang.org/std/option/enum.Option.html
+    pub async fn join(self) -> Option<T> {
+        self.handle.await
+    }
+
+    /// Rearm a [`TimerActionOnce`], so it fires at the specified [`Duration`] from now
```
HippoBaro, line 416: "at the specified [`Duration`] from now" -> "in the specified [`Duration`] from now"

scipio/src/timer.rs
```diff
@@ -283,48 +300,175 @@ impl<T: 'static> TimerAction<T> {
...
+    /// handle.join().unwrap();
+    /// ```
+    /// [`TimerActionOnce`]: struct.TimerActionOnce
+    /// [`Duration`]: https://doc.rust-lang.org/std/time/struct.Duration.html
+    pub fn rearm_in(&self, dur: Duration) {
+        let mut inner = self.inner.borrow_mut();
+        inner.reset(dur);
+    }
+
+    /// Rearm a [`TimerActionOnce`], for a specified [`Instant`]
```
HippoBaro, line 439: "/// Rearm a [`TimerActionOnce`], so it fires at the specified [`Instant`]"

For consistency with the other variant.

HippoBaro (approved):
Some nits, otherwise all good :+1:

PR #50, opened 2020-09-11
extract concurrency collections into `scipio::sync`

sitano:
PR extracts all concurrent collections related code into a separate module.

Changes:

- lib -> pub mod sync;
- scipio::collections::Deque
- scipio::sync::Semaphore
- scipio::time::{Timer, ...}
- ping_pong example that uses enclose!


sitano:
It seems to me that it is also worth of extracting `scipio::Timer` to `scipio::sync::Timer` as far as it is also a concurrency primitive. What do you think?

glommer:
Everything in scipio is a concurrency primitive, so if we do that, we'll end up moving everything =)

Let's deal with your proposed changes piece by piece:
* I like the fact that the AsyncDeque lives in a submodule and I was thinking of doing that anyway. So that is *good*. However I don't think the module should be called `sync` but `collections`. It's not a sync primitive. Pretty much just exporting the file. Right now there is only one collection but that's ok. 
* I am not sure I like the rename of AsyncDeque to Deque. It behaves differently enough from a normal deque to grant a special name. I need to think a bit more.
* Semaphore can live in `sync`. I thought of offering condvars as well but they are nasty.

sitano:
@glommer almost all of it that I bring here is the standard Go convention. And that's why I did it. In Go, `sync` package contains all primitives that are intended for concurrency. In the case of `scipio` its everything about `async`. `sync` mod usually includes 2 things:

- primitives (timer, mutex, sem, convar, channel, ...)
- collections

Then the second convention of Go is to not to duplicate the semantics on the package in the types names. Fore example:

- don't have `sync.SyncQueue`

So as far as we have sync.AsyncDeque, I thought it is nice convention to live it as `scipio::sync::Deque`.

Another useful convention why its not worth keeping AsyncDeque comes from Java and sounds like:

- Interfaces or Types must not reflect concurrent nature in names.

It's not always the true as Java has ConcurrentHashMap. But I also find it's useful.

Anyway, if you see sense of having `collections` as a separate module I will do.

What about Timer? Maybe also move it to the `sync`?

I also understand that foreign conventions are not applicable here, but they were something I based on.

Waiting for your input.

glommer:
You have mentioned many conventions, but none of them are Rust conventions =)
To be fair, I don't consider myself skilled enough in rust to judge. If there are *rust* conventions that we can point to here I'd like to follow them. Otherwise what I suggested should be good.

If they are go, java, c++ conventions... Then I see no reason to change.

If we separate the timer , I think it should go under `time`.

sitano:
> You have mentioned many conventions, but none of them are Rust conventions =)
> To be fair, I don't consider myself skilled enough in rust to judge. If there are _rust_ conventions that we can point to here I'd like to follow them. Otherwise what I suggested should be good.
> 
> If they are go, java, c++ conventions... Then I see no reason to change.
> 
> If we separate the timer , I think it should go under `time`.

@glommer Do you a see sense in this separation into modules overall or it's not worth it?

glommer:
As I had mentioned, I was already thinking about separating the Deque into its own module, so yes.
My suggestion would be:
`sync`: the semaphore
`collections`: the Deque
`time`: the timers

However I wouldn't rush into implementing this. I would like to hear the opinion of other people who have been contributing recently as well so that doesn't become a dictatorship =) Tagging some

@penberg @HippoBaro @Daniel-B-Smith comments welcome

HippoBaro:
I dislike the idea of putting data-structures in a `sync` module. Go does this because those data-structures are _thread-safe_. None of ours are, and so doing that would be confusing for newcomers. 
A `collections` module for datastructure is the rust-way so that's what we should do. 

As for `Timer` I'm not sure it should live in a `time` module because it is a scheduling primitive. Why not a `sched` module?

glommer:
Timer isn't really a scheduling primitive. It's a timer =)

So far I think the only agreement I see here is in separating collections. I'd suggest we move forward with that and leave the rest be.

sitano:
Shall I rename back Deque into AsyncDeque?

sitano:
Forget my interest but what do you think of renaming Semaphore into Sem? :D

glommer:
I gave this some thought.

Let's go with your suggestion of naming it just Deque. I was afraid of confusion with rust's std primitive but that one is called `VecDeque`.

I would prefer that we merge just the collections part.
Hippolyte dislikes moving the semaphore into the `sync` module and I don't think there is sufficient agreement with the timer.

Even myself, now that I think of it, I think `timer` is better than `time` as a module.
`time` is usually used for things like duration, instants, etc.

timer is not a scheduling primitive, but it is not a timekeeping primitive either.

Summary:

* The semaphore changes should disappear. Not going to merge.
* The collection changes are good as is. I will merge them.
* I am willing to merge the timer changes if you rename the module to `timer`.

sitano:
@glommer are you sure Hippolyte does not like sync::Semaphore? I saw he wrote only about data structures in sync. Asking just to make sure I didn’t miss anything.

sitano:
Do you want separate commits?

glommer:
You can always ask him, but I'd leave the Semaphore alone for now

glommer:
a single commit (that doesn't include semaphore) is fine

sitano:
@HippoBaro do you think having sync mod is cool or I missed your point?

HippoBaro:
`sync::Semaphore` is fine by me

sitano:
@glommer I have done what you asked:

- collections.Deque
- timer.Timer
- I left Semaphore, however, renamed mod into just semaphore.rs instead of local_semaphore.rs

AFAIU @HippoBaro does not object for moving concurency primitives into `sync`. It makes sense if you are gonna to add more primitives. For example Mutex, Channel, CondVar, ... If scipio is gonna have those it makes sense to introduce `sync`.

glommer:
Hi Ivan

Github tells me that there are conflicts, though.
Can you please resolve and provide an updated version ?

glommer (approved):

Issue #33, opened 2020-09-07
Poll executor for a while before going to sleep


glommer:
Going to sleep is expensive, so it is better to poll the executor for a while before going to sleep, by calling `poll_io` instead of `park` even if we have no tasks. If we always do that, though, then our CPU usage is always 100% which is bad too.

Seastar, for instance, will default to polling for some time before going to sleep.

In our implementation, an executor may or may not be pinned to a CPU:
* Executors that are pinned should never poll.
* Executors that are pinned may or may not poll, depending on the configuration.

The default should definitely be to not poll.

When polling is employed, we need to keep statistics about how much time was spent polling to inform users about "real" CPU usage vs "seen-by-Linux" CPU usage.

glommer:
This was closed by #53. However there are left overs. I'll open a new issue for them.

PR #53, opened 2020-09-12
executor::run(): spin a while before parking coroutine (#33)

sitano:
### What does this PR do?

This is a very draft of the #33 issue that offers to spin a thread for a while before parking it if it was pinned. 

### Motivation

Going to sleep is expensive, so it is better to poll the executor for a while before going to sleep, by calling poll_io instead of park even if we have no tasks. If we always do that, though, then our CPU usage is always 100% which is bad too.

### Related issues

#33

### Additional Notes

Seastar, for instance, will default to polling for some time before going to sleep.

In our implementation, an executor may or may not be pinned to a CPU:

- Executors that are pinned should never poll.
- Executors that are pinned may or may not poll, depending on the configuration.
- The default should definitely be to not poll.

When polling is employed, we need to keep statistics about how much time was spent polling to inform users about "real" CPU usage vs "seen-by-Linux" CPU usage.

### Checklist

### DIFF
--- a/scipio/src/executor.rs
+++ b/scipio/src/executor.rs
@@ -1270,3 +1270,44 @@ fn test_detach() {
         Timer::new(Duration::from_micros(100)).await;
     });
 }
+
+/// As far as impl From<libc::timeval> for Duration is not allowed.
+#[cfg(test)]
+fn from_timeval(v: libc::timeval) -> Duration {
+    Duration::from_secs(v.tv_sec as u64) + Duration::from_micros(v.tv_usec as u64)
+}
+
+#[cfg(test)]
+fn getrusage() -> libc::rusage {
+    use core::mem::MaybeUninit;
+    let mut s0 = MaybeUninit::<libc::rusage>::uninit();
+    let err = unsafe { libc::getrusage(libc::RUSAGE_THREAD, s0.as_mut_ptr()) };
+    if err != 0 {
+        panic!("getrusage error = {}", err);
+    }
+    unsafe { s0.assume_init() }
+}
+
+#[cfg(test)]
+fn getrusage_utime() -> Duration {
+    from_timeval(getrusage().ru_utime)
+}
+
+#[test]
+fn test_no_spin() {
+    use crate::timer;
+
+    let ex = LocalExecutor::make_default();
+    let task_queue = ex.create_task_queue(1000,
+                                                Latency::Matters(Duration::from_millis(10)), "my_tq");
+    let start = getrusage_utime();
+    let task = ex.spawn_into(async {
+        timer::sleep(Duration::from_secs(1)).await
+    }, task_queue).expect("failed to spawn task");
+    ex.run(async { task.await });
+
+    assert!(
+        getrusage_utime() - start < Duration::from_millis(1),
+        "expected user time on LE is less than 1 millisecond"
+    );
+}
--- a/scipio/src/executor.rs
+++ b/scipio/src/executor.rs
@@ -251,6 +251,7 @@ struct ExecutorQueues {
     executor_index: usize,
     last_vruntime: u64,
     preempt_timer_duration: Duration,
+    spin_before_park: Option<Duration>,
 }
 
 impl ExecutorQueues {
@@ -263,6 +264,7 @@ impl ExecutorQueues {
             executor_index: 1, // 0 is the default
             last_vruntime: 0,
             preempt_timer_duration: Duration::from_secs(1),
+            spin_before_park: None,
         }))
     }
 
@@ -322,7 +324,9 @@ impl ExecutorQueues {
 pub struct LocalExecutorBuilder {
     // The id of a CPU to bind the current (or yet to be created) thread
     binding: Option<usize>,
-    // A name for the thread-to-be (if any), for identification in panic messages
+    /// Spin for duration before parking a reactor
+    spin_before_park: Option<Duration>,
+    /// A name for the thread-to-be (if any), for identification in panic messages
     name: String,
 }
 
@@ -332,6 +336,7 @@ impl LocalExecutorBuilder {
     pub fn new() -> LocalExecutorBuilder {
         LocalExecutorBuilder {
             binding: None,
+            spin_before_park: None,
             name: String::from("unnamed"),
         }
     }
@@ -339,6 +344,12 @@ impl LocalExecutorBuilder {
     /// Sets the new executor's affinity to the provided CPU
     pub fn pin_to_cpu(mut self, cpu: usize) -> LocalExecutorBuilder {
         self.binding = Some(cpu);
+        self
+    }
+
+    /// Spin for duration before parking a reactor
+    pub fn spin_before_park(mut self, spin: Duration) -> LocalExecutorBuilder {
+        self.spin_before_park = Some(spin);
         self
     }
 
@@ -362,6 +373,7 @@ impl LocalExecutorBuilder {
         let mut le = LocalExecutor::new(EXECUTOR_ID.fetch_add(1, Ordering::Relaxed));
         if let Some(cpu) = self.binding {
             le.bind_to_cpu(cpu)?;
+            le.queues.borrow_mut().spin_before_park = self.spin_before_park;
         }
         match le.init() {
             Ok(_) => Ok(le),
@@ -673,6 +685,10 @@ impl LocalExecutor {
         self.queues.borrow().preempt_timer_duration
     }
 
+    fn spin_before_park(&self) -> Option<Duration> {
+        self.queues.borrow().spin_before_park
+    }
+
     fn run_one_task_queue(&self) -> bool {
         let mut tq = self.queues.borrow_mut();
         let candidate = tq.active_executors.pop();
@@ -738,6 +754,10 @@ impl LocalExecutor {
         let waker = waker_fn(|| {});
         let cx = &mut Context::from_waker(&waker);
 
+        let spin_before_park = self.spin_before_park().unwrap_or_default();
+        let spin = spin_before_park.as_nanos() > 0;
+        let mut spin_since: Option<Instant> = None;
+
         LOCAL_EX.set(self, || loop {
             if let Poll::Ready(t) = future.as_mut().poll(cx) {
                 break t;
@@ -751,7 +771,21 @@ impl LocalExecutor {
             let duration = self.preempt_timer_duration();
             self.parker.poll_io(duration);
             if !self.run_one_task_queue() {
+                if spin {
+                    if let Some(t) = spin_since {
+                        if t.elapsed() < spin_before_park {
+                            continue;
+                        }
+                        spin_since = None
+                    } else {
+                        spin_since = Some(Instant::now());
+                        continue;
+                    }
+                }
+
                 self.parker.park();
+            } else {
+                spin_since = None
             }
         })
     }
@@ -1298,16 +1332,49 @@ fn test_no_spin() {
     use crate::timer;
 
     let ex = LocalExecutor::make_default();
-    let task_queue = ex.create_task_queue(1000,
-                                                Latency::Matters(Duration::from_millis(10)), "my_tq");
+    let task_queue =
+        ex.create_task_queue(1000, Latency::Matters(Duration::from_millis(10)), "my_tq");
     let start = getrusage_utime();
-    let task = ex.spawn_into(async {
-        timer::sleep(Duration::from_secs(1)).await
-    }, task_queue).expect("failed to spawn task");
+    let task = ex
+        .spawn_into(
+            async { timer::sleep(Duration::from_secs(1)).await },
+            task_queue,
+        )
+        .expect("failed to spawn task");
     ex.run(async { task.await });
 
     assert!(
         getrusage_utime() - start < Duration::from_millis(1),
         "expected user time on LE is less than 1 millisecond"
+    );
+}
+
+#[test]
+fn test_spin() {
+    use crate::timer;
+
+    let dur = Duration::from_secs(1);
+    let ex0 = LocalExecutorBuilder::new().make().unwrap();
+    let ex0_ru_start = getrusage_utime();
+    ex0.run(async { timer::sleep(dur).await });
+    let ex0_ru_finish = getrusage_utime();
+
+    let ex = LocalExecutorBuilder::new()
+        .pin_to_cpu(0)
+        .spin_before_park(Duration::from_millis(10))
+        .make()
+        .unwrap();
+    let ex_ru_start = getrusage_utime();
+    let task = ex.spawn(async move { timer::sleep(dur).await });
+    ex.run(async { task.await });
+    let ex_ru_finish = getrusage_utime();
+
+    assert!(
+        ex0_ru_finish - ex0_ru_start < Duration::from_millis(1),
+        "expected user time on LE0 is less than 1 millisecond"
+    );
+    assert!(
+        ex_ru_finish - ex_ru_start >= Duration::from_millis(1),
+        "expected user time on LE is much greater than 1 millisecond"
     );
 }
--- a/scipio/src/timer/mod.rs
+++ b/scipio/src/timer/mod.rs
@@ -7,3 +7,20 @@
 mod timer_impl;
 
 pub use timer_impl::{Timer, TimerActionOnce, TimerActionRepeat};
+
+/// Sleep for some time.
+///
+/// ```
+/// use scipio::LocalExecutor;
+/// use scipio::timer::sleep;
+/// use std::time::Duration;
+///
+/// let ex = LocalExecutor::make_default();
+///
+/// ex.run(async {
+///     sleep(Duration::from_millis(100)).await;
+/// });
+/// ```
+pub async fn sleep(wait: std::time::Duration) {
+    Timer::new(wait).await;
+}


sitano:
@glommer a question: is general direction correct?
other questions:
- shall use CPU ticks instead of time?
- if no spin required then we must have no overhead
- shall we use cpu::yield instruction as a cheap version of waiting before next iteration of spinning?

scipio/src/executor.rs
```diff
@@ -247,6 +247,7 @@ struct ExecutorQueues {
     executor_index: usize,
     last_vruntime: u64,
     preempt_timer_duration: Duration,
+    spin_before_park: Option<Duration>,
```
glommer, line 250: good first step. But how is the user supposed to select this?

Somehow the user of the library must choose that. I suspect that the best way is instead of passing
just a cpu binding option to the executor, we will pass a struct with options.

For now it would contain the pinning settings plus the spin setting.

With time, if we add more, the API would stay the same and the creation of LocalExecutor wouldn't break.

Two important things are also missing from this PR:
- A test. We can use linux's `getrusage(2)` to make sure this has a real effect.
- Storing a metric in the executor that captures real usage vs total usage

glommer (changes requested):

examples/benches/clock.rs
```diff
@@ -0,0 +1,18 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
```
glommer, line 1: this feel out of place.

scipio/src/executor.rs
```diff
@@ -1193,3 +1235,39 @@ fn test_detach() {
...
+
+    let now = Instant::now();
+
+    getrusage(s0.as_mut_ptr());
+    local_ex.run(async {});
+    getrusage(e0.as_mut_ptr());
+
+    let start = unsafe { s0.assume_init() };
+    let end = unsafe { e0.assume_init() };
+    println!("elapsed time = {}", now.elapsed().as_micros());
```
glommer, line 1264: those prints would all be ignored and not even seen unless we pass `--nocapture`.

That is the wrong way to test this. The right way is to make sure that by looking at `rusage` we are seeing a lot more time alloted to our thread than what we thought we used.

To make sure the test is not flaky you should run it once with the system free, and another time pinned to a cpu that has many other tasks running (like just `yes`). Both versions should pass.

sitano: The code is in the middle of the development process. PR is a draft. I was half-way thro. Don't look at it until it's not a draft.

examples/benches/clock.rs
sitano: The code is in the middle of the development process. PR is a draft. I was half-way thro. Don't look at it until it's not a draft.

HippoBaro (changes requested):
CI has been updated since this PR was opened, please rebase on master to enable the new linter passes :bowing_man:

scipio/src/executor.rs
sitano: done

sitano:
@glommer I am a bit worried `ex.run(async { Timer::new(10 secs).await })` on a default non-spinning executor from the `master` consumes whole CPU core and huge u/s/stat. Don't think it's ok. However, what do you think?

Example:
```
/// env SCIPIO_TEST_POLLIO_ROOTDIR=/tmp cargo test executor::test_spin -- --nocapture
#[test]
fn test_spin() {
    use crate::timer;

    let dur = Duration::from_secs(10);
    let ex0 = LocalExecutorBuilder::new().make().unwrap();
    ex0.run(async { timer::Timer::new(dur).await });
}
```

glommer:
This looks like a bug to me.

glommer:
Ok, Ivan. I will push a fix to this. It is indeed broken. I won't bother writing unit tests, though, trusting that you will merge unit tests that will cover that soon.

Please consider merging the unit tests for the sleeping reactor sooner so we can get coverage here.

glommer:
You will need this for the test to pass, Ivan:
https://github.com/DataDog/scipio/pull/64

scipio/src/executor.rs
```diff
@@ -1270,3 +1358,57 @@ fn test_detach() {
...
+    let mut s0 = MaybeUninit::<libc::rusage>::uninit();
+    let err = unsafe { libc::getrusage(libc::RUSAGE_THREAD, s0.as_mut_ptr()) };
+    if err != 0 {
+        panic!("getrusage error = {}", err);
+    }
+    unsafe { s0.assume_init() }
+}
+
+fn ru_utime_to_duration(r: &libc::rusage) -> Duration {
+    Duration::from_micros((r.ru_utime.tv_sec * 1_000_000 + r.ru_utime.tv_usec) as u64)
```
sitano, line 1373: unify timeval

scipio/src/executor.rs
```diff
@@ -1270,3 +1358,57 @@ fn test_detach() {
         Timer::new(Duration::from_micros(100)).await;
     });
 }
+
+fn getrusage() -> libc::rusage {
```
sitano, line 1362: return something that has Debug

scipio/src/executor.rs
```diff
@@ -1270,3 +1358,57 @@ fn test_detach() {
...
+        0,
+        "expected number of spins in LE0 is 0"
+    );
+    assert!(
+        ex0.queues.borrow().ru_utime < 10000000,
+        "expected user time on LE0 is less than 1 millisecond"
+    );
+    assert!(
+        ex.queues.borrow().ru_utime >= 10000000,
+        "expected user time on LE is less than 1 millisecond"
```
sitano, line 1408: fix msg

scipio/src/timer/mod.rs
```diff
@@ -6,4 +6,22 @@
...
+/// use scipio::timer::sleep;
+/// use std::time::Duration;
+///
+/// let ex = LocalExecutor::make_default();
+///
+/// ex.run(async {
+///     sleep(Duration::from_millis(100)).await;
+/// });
+/// ```
+pub async fn sleep(wait: Duration) {
```
sitano, line 25: write Unit Test that ensures default Executor actually sleeps while waiting

scipio/src/executor.rs
```diff
@@ -1270,3 +1358,57 @@ fn test_detach() {
...
+    use core::mem::MaybeUninit;
+    let mut s0 = MaybeUninit::<libc::rusage>::uninit();
+    let err = unsafe { libc::getrusage(libc::RUSAGE_THREAD, s0.as_mut_ptr()) };
+    if err != 0 {
+        panic!("getrusage error = {}", err);
+    }
+    unsafe { s0.assume_init() }
+}
+
+fn ru_utime_to_duration(r: &libc::rusage) -> Duration {
```
sitano, line 1372: move to the local rusage struct

scipio/src/executor.rs
```diff
@@ -748,10 +819,27 @@ impl LocalExecutor {
...
-            let duration = self.preempt_timer_duration();
-            self.parker.poll_io(duration);
-            if !self.run_one_task_queue() {
-                self.parker.park();
+            self.parker.poll_io(preempt);
+
+            let r0 = SystemTime::now();
+            let ok = self.run_one_task_queue();
+            let r1 = SystemTime::now();
+            ru_runtime += r1.duration_since(r0).unwrap_or_default().as_nanos() as u64;
```
sitano, line 827: rewrite SystemTime impl to use cheaper diff calc

scipio/src/executor.rs
```diff
@@ -737,9 +787,30 @@ impl LocalExecutor {
...
+        let spin = spin_before_park.as_nanos() > 0;
+        let mut now = SystemTime::now();
+        let start = now;
+        let ru0 = getrusage();
+        let mut ru_runtime: u64 = 0;
+        let mut ru_spin: u64 = 0;
 
         LOCAL_EX.set(self, || loop {
             if let Poll::Ready(t) = future.as_mut().poll(cx) {
+                let mut q = self.queues.borrow_mut();
```
sitano, line 801: could be self.queues.borrow().stats.borrow_mut()

scipio/src/executor.rs
```diff
@@ -263,6 +276,13 @@ impl ExecutorQueues {
             executor_index: 1, // 0 is the default
             last_vruntime: 0,
             preempt_timer_duration: Duration::from_secs(1),
+            spin_before_park: None,
+            ru_total: 0,
+            ru_runtime: 0,
+            ru_utime: 0,
+            ru_stime: 0,
+            ru_nvcsw: 0,
+            ru_spin: 0,
```
sitano, line 285: to Stats struct

scipio/src/executor.rs
```diff
@@ -737,9 +787,30 @@ impl LocalExecutor {
...
+                q.ru_runtime += ru_runtime;
+                q.ru_spin += ru_spin;
+
+                let ru = getrusage();
+                q.ru_utime +=
+                    (ru_utime_to_duration(&ru) - ru_utime_to_duration(&ru0)).as_nanos() as u64;
+                q.ru_stime +=
+                    (ru_stime_to_duration(&ru) - ru_stime_to_duration(&ru0)).as_nanos() as u64;
+                q.ru_nvcsw += (ru.ru_nvcsw - ru0.ru_nvcsw) as u64;
+
```
sitano, line 813: hide it somewhere

scipio/src/executor.rs
sitano, line 827: encapsulate?

scipio/src/executor.rs
```diff
@@ -748,10 +819,27 @@ impl LocalExecutor {
...
+            let r0 = SystemTime::now();
+            let ok = self.run_one_task_queue();
+            let r1 = SystemTime::now();
+            ru_runtime += r1.duration_since(r0).unwrap_or_default().as_nanos() as u64;
+            if ok {
+                // If we have processed a task, look for the next one
+                continue;
+            }
+
+            if spin && r1.duration_since(now).unwrap_or_default() < spin_before_park {
```
sitano, line 833: join

scipio/src/executor.rs
sitano: done with sys::ResourceUsage

scipio/src/executor.rs
sitano: done with sys::ResourceUsage

scipio/src/executor.rs
sitano: done with sys::ResourceUsage

scipio/src/timer/mod.rs
sitano: new test is `test_no_spin` : `env SCIPIO_TEST_POLLIO_ROOTDIR=/tmp cargo test executor::test_no_spin -- --nocapture`

scipio/src/executor.rs
sitano: done

scipio/src/executor.rs
sitano: extracted into impl Sub for RU. will move away for Stats.

scipio/src/executor.rs
sitano: no, can't

scipio/src/executor.rs
sitano: done

sitano:
`ExecutorStats` and `ResourceUsage` are ready for review

HippoBaro:
I haven't reviewed the code thoroughly although I have a few questions,

- You use `u64` ints to represent durations, is there a reason not to use rust's `std::time::Duration`? 
- It took me a while to realize that you are using operator overloading. I am generally against that. I am new to rust however so I am wondering if you had specific reasons for doing so. As a side-node, it looks to me like this could be quite wasteful (imagine only one counter in non-zero [here](https://github.com/DataDog/scipio/pull/53/files#diff-a695e210f07ac031389c3fb56ddb9879R787), you'll end up doing a lot of spurious additions).

glommer:
I am fine with implementing `Add` for the getrusage type so you can sum two entities that has the same type.
I am a bit more on the fence with ExecutorStats += getrusage. Operator overloading is a polemic feature, I think the best way to use it is in scenarios where their semantics is super obvious and fundamentally well defined.

Summing two getrusage instances works: all fields are monotonic, they are all well known in advance, and what the fields are is not changing anytime soon.

So I think it would be cleaner - and totally acceptable to have a field getrusage inside `ExecutorStats`. Then you could do `stats.rusage += x`

sitano:
> I haven't reviewed the code thoroughly although I have a few questions,
> 
> * You use `u64` ints to represent durations, is there a reason not to use rust's `std::time::Duration`?

I can use them yes. I am worried that its too big (u64 + u32 2 fields). Its fine for us to go just with a single u64 nanos. I also hoped to rewrite SystemTime to have

1) more simplistic elapsed(), which could ignore nonmonotonicity and would not return Result
2) will return u64 from the beginning

Anyway, I can move to Duration completely. Shall I?

> * It took me a while to realize that you are using operator overloading. I am generally against that. I am new to rust however so I am wondering if you had specific reasons for doing so. As a side-node, it looks to me like this could be quite wasteful (imagine only one counter is non-zero [here](https://github.com/DataDog/scipio/pull/53/files#diff-a695e210f07ac031389c3fb56ddb9879R787), you'll end up doing a lot of spurious additions).

I just wanted to simplify code in Executor.Run. If you think it makes sense I can replace it with just functions.

sitano:
I removed Add<RU> for Stats impl and replaced it with the method `account_usage`.
I also extracted measure call into `Stats.call_runtime`.
I introduced `LocalExecutor::want_spin` to simplify internal loop logic and returned back original nesting.
Now I am heavily blocked by the tests that don't work for me with: `failed to create file: No such file or directory (os error 2)`.

sitano:
Ok, all tests pass besides `file_fallocate_alocatee` under and without load, so I am converting it.

HippoBaro:
> thread 'dma_file::test::file_fallocate_alocatee' panicked at 'fallocate failed: Operation not supported (os error 95), op: Pre-allocate space path /tmp/storage_media/file_fallocate_alocatee/testfile with fd 6', scipio/src/dma_file.rs:633:13

Reading your change from #65 it is to be expected. tmpfs doesn't support O_DIRECT nor `fallocate`.

HippoBaro:
> Anyway, I can move to Duration completely. Shall I?

No, if you believe a `u64` makes the code simpler in the reactor, go for it! However when the user requests the stats, it would be nice to convert them to standard rust duration.

Once you feel like the PR is ready for a proper review, it would be nice to break your changes into multiple commits.

sitano:
> > thread 'dma_file::test::file_fallocate_alocatee' panicked at 'fallocate failed: Operation not supported (os error 95), op: Pre-allocate space path /tmp/storage_media/file_fallocate_alocatee/testfile with fd 6', scipio/src/dma_file.rs:633:13
> 
> Reading your change from #65 it is to be expected. tmpfs doesn't support O_DIRECT nor `fallocate`.

Oh. Didn't know about O_DIRECT. However, the internet says it must support `fallocate`:

- https://lwn.net/Articles/467424/ tmpfs: add fallocate support from 2011
- https://man7.org/linux/man-pages/man2/fallocate.2.html support for tmpfs since Linux 3.5

sitano:
> > Anyway, I can move to Duration completely. Shall I?
> 
> No, if you believe a `u64` makes the code simpler in the reactor, go for it! However when the user requests the stats, it would be nice to convert them to standard rust duration.

It's a good point about simplicity. I must think about it. I thought the duration calculations could be done in a more performant manner because basically, we don't need u128 and back in time travels can be just omitted. 

> Once you feel like the PR is ready for a proper review, it would be nice to break your changes into multiple commits. From what I can see your changeset does at least two distinct things: (1) it introduces stats reporting in the reactor and (2) allows the reactor to spin before parking.

Ok. I feel this PR is ready for a proper review. I will split it.

HippoBaro:
> > > thread 'dma_file::test::file_fallocate_alocatee' panicked at 'fallocate failed: Operation not supported (os error 95), op: Pre-allocate space path /tmp/storage_media/file_fallocate_alocatee/testfile with fd 6', scipio/src/dma_file.rs:633:13
> > 
> > 
> > Reading your change from #65 it is to be expected. tmpfs doesn't support O_DIRECT nor `fallocate`.
> 
> Oh. Didn't know about O_DIRECT on tmpfs. However, the internet says it must support `fallocate`:
> 
> * https://lwn.net/Articles/467424/ tmpfs: add fallocate support from 2011
> * https://man7.org/linux/man-pages/man2/fallocate.2.html support for tmpfs since Linux 3.5

We use `fallocate` with `FALLOC_FL_ZERO_RANGE`: https://github.com/DataDog/scipio/blob/6de0fb57b187147d7e6d0574fb92afb4e1b3f7d8/scipio/src/dma_file.rs#L387
Unfortunately `fallocate` doesn't support that mode for `tmpfs`.

@glommer Should we create an issue for this? If I'm reading correctly, we could emulate the behavior of `FALLOC_FL_ZERO_RANGE` with a call to `fallocate` with mode `0` + a `memset` over the data range.

sitano:
Splitted

scipio/src/executor.rs
sitano: done

sitano: done

scipio/src/executor.rs
sitano: done

scipio/src/executor.rs
```diff
@@ -751,13 +765,25 @@ impl LocalExecutor {
...
             if !self.run_one_task_queue() {
                 self.parker.park();
             }
         })
     }
+
+    /// Polls f and measures its time.
+    fn measure_poll<T>(
+        stats: &mut ExecutorStats,
+        future: Pin<&mut impl Future<Output = T>>,
```
sitano, line 821: I am not sure about it to whoever will be reviewing this

glommer:
@HippoBaro I am not sure if we should emulate that on tmpfs. Depends on why we are using it for. The usage I have in mind is expanding the size of a file so it won't block on writes. But tmpfs wouldn't block anyway so by adding the memset step we're making writes slower for no reason.

scipio/src/timer/duration.rs
```diff
@@ -0,0 +1,324 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
```
glommer, line 1: I think the benefits of having this are not strong enough to outweight the confusion of users having to figure out the differences between various duration implementations.

In general we should stay away from reimplementing standard types without good reason. I'd prefer you keep everything as u64 until the very last step for presentation than doing this.

scipio/src/timer/time.rs
```diff
@@ -0,0 +1,101 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
```
glommer, line 1: Much like the previous patch, this one has to go too.

We had this conversation in private, but for the benefits of others:

`Instant::now()` does not take a mutex for no reason. In platforms where timekeeping is stable it doesn't take the mutex at all. And where it is not (something like) the mutex is needed anyway.

scipio/src/sys/rusage.rs
```diff
@@ -0,0 +1,123 @@
+#![allow(dead_code)]
```
glommer, line 1: Is it because there are fields in `ResourceUsage` that we don't read? It's best if we attach this to the struct, unless there is unused code all over.

scipio/src/sys/rusage.rs
```diff
@@ -0,0 +1,123 @@
...
+/// As far as impl From<libc::timeval> for Duration is not allowed.
+fn from_timeval(v: libc::timeval) -> Duration {
+    Duration::from_micros((v.tv_sec * 1_000_000 + v.tv_usec) as u64)
+}
+
+fn getrusage() -> libc::rusage {
+    use core::mem::MaybeUninit;
+    let mut s0 = MaybeUninit::<libc::rusage>::uninit();
+    let err = unsafe { libc::getrusage(libc::RUSAGE_THREAD, s0.as_mut_ptr()) };
+    if err != 0 {
```
glommer, line 119: assert_eq is better here. 

Btw, did you run `cargo clippy` ?

scipio/src/sys/rusage.rs
```diff
@@ -0,0 +1,123 @@
...
+        Self::from(getrusage())
+    }
+
+    pub fn runtime() -> Self {
+        let r = getrusage();
+        let mut u = ResourceUsage::new();
+
+        u.ru_utime = from_timeval(r.ru_utime).as_nanos() as u64;
+        u.ru_stime = from_timeval(r.ru_stime).as_nanos() as u64;
+        u.ru_nvcsw = r.ru_nvcsw as u64;
```
glommer, line 54: if those are all the fields we need, I am wondering whether it wouldn't be best if we just don't declare the others?

If we want to declare all of them for some reason, maybe it is best to keep the ones we use together in the struct so they  fall in the same cacheline

scipio/src/executor.rs
```diff
@@ -740,9 +745,18 @@ impl LocalExecutor {
...
+        let preempt = self.preempt_timer_duration();
+
+        let start = now(CLOCK_MONOTONIC);
+
+        let ru0 = ResourceUsage::runtime();
+        let mut stats = ExecutorStats::new();
 
         LOCAL_EX.set(self, || loop {
-            if let Poll::Ready(t) = future.as_mut().poll(cx) {
+            if let Poll::Ready(t) = LocalExecutor::measure_poll(&mut stats, future.as_mut(), cx) {
```
glommer, line 789: why are we only accounting when the future is ready ?
In a long running server this future will essentially never be ready and we'll always be pulling tasks from the task queue.

scipio/src/executor.rs
glommer: This is not the right approach.

As I said above, the main future that we poll here will essentially only be ready when the whole program is ready. In a long-living server, that might mean never. We need to account task queues, no futures.

You could try to measure *every* poll, but that's too expensive. So again, the right thing to do is to measure task queue time. Once we resort to measuring task queue time, there may be time unaccounted because it was run in the main future. For that we can add something like you are adding here. But it is a complement, not the whole thing.

scipio/src/executor.rs
```diff
@@ -1337,3 +1366,45 @@ fn test_no_spin() {
...
+    ex0.run(async { timer::sleep(dur).await });
+
+    let ex = LocalExecutorBuilder::new()
+        .spin_before_park(Duration::from_millis(10))
+        .make()
+        .unwrap();
+    let task = ex.spawn(async move { timer::sleep(dur).await });
+    ex.run(async { task.await });
+
+    let stats0 = &ex0.queues.borrow().stats;
```
glommer, line 1389: There should be a way to acquire the stats without looking into the queues.
The queues are an implementation detail, and from outside the executor should not be needed.

glommer (changes requested):

scipio/src/executor.rs
sitano: Ok, I moved to the simplest implementation possible for spinning only. This piece no longer exists.

scipio/src/timer/duration.rs
sitano: Wiped

scipio/src/timer/time.rs
sitano: Wiped

scipio/src/sys/rusage.rs
sitano: Wiped

scipio/src/sys/rusage.rs
sitano: Wiped

scipio/src/sys/rusage.rs
sitano: Wiped

scipio/src/executor.rs
sitano: Wiped

scipio/src/executor.rs
sitano: Wiped

sitano:
I reverted the patch to the most simplistic implementation without any statistics.

scipio/src/executor.rs
```diff
@@ -320,7 +320,12 @@ impl ExecutorQueues {
 /// ```
 #[derive(Debug)]
 pub struct LocalExecutorBuilder {
-    // The id of a CPU to bind the current (or yet to be created) thread
+    /// Preempt timeout. A value of zero means we are not interested in
+    /// installing a preempt timer. Tasks executing in the CPU right after this
+    /// will be able to check if the timer has elapsed and yield the CPU if that
+    /// is the case.
+    preempt_timer_duration: Duration,
```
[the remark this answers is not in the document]
sitano: You are right. Reworked a test into custom Latency queue.

scipio/src/executor.rs
```diff
@@ -761,10 +783,24 @@ impl LocalExecutor {
             // requests that are latency sensitive we want them out of the
             // ring ASAP (before we run the task queues). We will also use
             // the opportunity to install the timer.
-            let duration = self.preempt_timer_duration();
```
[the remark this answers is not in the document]
sitano: Yeah. Reverted.

sitano:
@glommer I have fixed the latest issues. However, I didn't have a chance to run tests.

glommer:
I tested your code and all tests pass.

I am making now a final round of comments. (which means if you don't see those comments by tomorrow then github is to blame!! ping me!)

scipio/src/executor.rs
```diff
@@ -362,6 +373,8 @@ impl LocalExecutorBuilder {
         let mut le = LocalExecutor::new(EXECUTOR_ID.fetch_add(1, Ordering::Relaxed));
         if let Some(cpu) = self.binding {
             le.bind_to_cpu(cpu)?;
+        } else {
```
glommer, line 376: I don't understand why this is being done only if there is no binding.
It's actually the contrary: we should only allow spinning at all if there is binding: if there is no binding you may have two executors in the same CPU, now fighting each other by spinning like crazy.

glommer (changes requested):

scipio/src/executor.rs
sitano: I misinterpret this:

> In our implementation, an executor may or may not be pinned to a CPU:
> Executors that are pinned should never poll.
> Executors that are pinned may or may not poll, depending on the configuration.

I still don't understand what it means.

sitano:
@glommer thank you for your patience. next turn.

sitano:
@glommer speaking of this

> When polling is employed, we need to keep statistics about how much time was spent polling to inform users about "real" CPU usage vs "seen-by-Linux" CPU usage.

If I understand the `executor.Run` pattern correctly it is expected to spin for a whole application lifetime. Then measuring just start and now timestamps must be enough for tracking total time. Is that correct?

Thus, would it be enough to keep `(start_timestamp, last_iteration_timestamp)` in the `ExecutorQueues` and update `last_iteration_timestamp` in `run_one_task_queue` before `match canditate` block?

`ExecutorQueues` picked because it's a central point that holds all the queues. So it makes sense to have overall stats.

scipio/src/executor.rs
glommer: Sorry Ivan, this is my bad.
Notice how both phrases start with "Executors that are pinned" so one of them has to be wrong =)
Executors that are *not* pinned should never poll.
Executors that are pinned may or may not poll, depending on the configuration.

glommer:
Hi Ivan,

Note that for most metrics, what we end up needing is a total over time that keeps increasing and a way to measure grab the current value. Metrics are usually exported every 15s or something like that.

How about we don't worry about that at all for now?
There's nothing special about the work that you are doing: we lack metrics across the board. Pretty much nothing has metrics at the moment. We'll have to come back later and add them and at that time we can add for the spinners too.

glommer (approved):

glommer:
I am force merging this. Circle CI seems stuck. I have run the tests manually on my machine.

If anything fails we can fix it later.

PR #54, opened 2020-09-12
Dma file tests

glommer:
### What does this PR do?

Cleans up the test directory that was crated for the test

### Motivation

Right now we delete the directory before starting the test, but leave the files be when the test ends.
It's a better practice to cleanup after ourselves.

### Related issues

Before we do that, let's move these tests to the `mod test` pattern that we have been using in other files

### Checklist

### DIFF
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -461,308 +461,312 @@ impl DmaFile {
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


HippoBaro (approved):

glommer:
@HippoBaro wth are you doing here, go drink some wine

Issue #32, opened 2020-09-07
Add helpers to create many executors with placement hints.


glommer:
Using the `hwloc` library, we are able to discover the topology of the box.

Our current API to create an executor takes a CPU number. But it would be nice if we could also have another, more expressive helper where we can express placement constraints and get that from `hwloc`.

Examples:
* "Give me two executors, placing them both in the same physical CPU, just in different hyperthreads"
* "Give me one executor in each socket in this box"

@penberg - you might find this interesting

qiujiangkun:
I'm working on my own project with similar issues. Hope it would give you some tips. Later on I'll be working on this as part of my job.
My project has to support both x86 and aarch64, both MacOS and linux.
`hwloc2-rs` does not compile on MacOS or aarch64 unfortunately, so I designed a CoreAllocator trait(only needs numbers). e.g a DummyAllocator that does not bind to cores, a SmartAllocator that allocates network related to 0 core, cpu intensive work to half of the remaining cores(I don't want hyperthread).  I also uses conditional compilation not to compile hwloc2-rs on Mac, to bind cores on aarch64 linux with cpu affinity, not to bind cores on MacOS, specify cores manually when I separate cores in grub kernal options(which makes cpu-cores not working properly)

PR #56, opened 2020-09-13
More self documenting types; LocalExecutor configuration

HippoBaro:
### What does this PR do?

> This PR contains three commits each introducing an independent type.
> Pick and choose!

This PR swaps some generic types for custom ones in the interest of self
documenting code.

The main (externally visible change) is that we currently use an
`Option<usize>` to optionally bind a `LocalExecutor` to a CPU. While
this is functionally fine, it does poorly in terms of API readability.
Newcomers reading examples or the documentation will likely wonder why
`None` is frequently passed as an argument to `LocalExecutor::new` and
variants.

After merging,

```rust
let local_ex = LocalExecutor::new(None);
let local_ex = LocalExecutor::new(Some(1));
```

becomes:

```rust
let local_ex = LocalExecutor::new(CpuBinding::AnyCpu);
let local_ex = LocalExecutor::new(CpuBinding::Cpu(1));
```

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -25,7 +25,8 @@ async fn server(conns: usize) {
     // necessarily spawn all executors at once running symmetrical code.
     let client_handle = LocalExecutorBuilder::new()
         .pin_to_cpu(2)
-        .spawn("client", move || async move {
+        .name(String::from("client"))
+        .spawn(move || async move {
             client(conns).await;
         })
         .unwrap();
@@ -82,7 +83,7 @@ fn main() -> Result<()> {
     // system configuration and most modern systems will balance it, but that it is
     // still common enough that it is worth excluding it in this benchmark
     let builder = LocalExecutorBuilder::new().pin_to_cpu(1);
-    let server_handle = builder.spawn("server", || async move {
+    let server_handle = builder.name(String::from("server")).spawn(|| async move {
         // If you try `top` during the execution of the first batch, you
         // will see that the CPUs should not be at 100%. A single connection will
         // not be enough to extract all the performance available in the cores.
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -25,7 +25,7 @@ async fn server(conns: usize) {
     // necessarily spawn all executors at once running symmetrical code.
     let client_handle = LocalExecutorBuilder::new()
         .pin_to_cpu(2)
-        .spawn_thread("client", move || async move {
+        .spawn("client", move || async move {
             client(conns).await;
         })
         .unwrap();
@@ -82,7 +82,7 @@ fn main() -> Result<()> {
     // system configuration and most modern systems will balance it, but that it is
     // still common enough that it is worth excluding it in this benchmark
     let builder = LocalExecutorBuilder::new().pin_to_cpu(1);
-    let server_handle = builder.spawn_thread("server", || async move {
+    let server_handle = builder.spawn("server", || async move {
         // If you try `top` during the execution of the first batch, you
         // will see that the CPUs should not be at 100%. A single connection will
         // not be enough to extract all the performance available in the cores.
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -25,7 +25,7 @@ async fn server(conns: usize) {
     // necessarily spawn all executors at once running symmetrical code.
     let client_handle = LocalExecutorBuilder::new()
         .pin_to_cpu(2)
-        .name(String::from("client"))
+        .name("client")
         .spawn(move || async move {
             client(conns).await;
         })
@@ -83,7 +83,7 @@ fn main() -> Result<()> {
     // system configuration and most modern systems will balance it, but that it is
     // still common enough that it is worth excluding it in this benchmark
     let builder = LocalExecutorBuilder::new().pin_to_cpu(1);
-    let server_handle = builder.name(String::from("server")).spawn(|| async move {
+    let server_handle = builder.name("server").spawn(|| async move {
         // If you try `top` during the execution of the first batch, you
         // will see that the CPUs should not be at 100%. A single connection will
         // not be enough to extract all the performance available in the cores.
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -5,7 +5,7 @@
 //
 use futures::future::join_all;
 use futures::io::{copy, AsyncReadExt, AsyncWriteExt};
-use scipio::{Async, Local, LocalExecutor};
+use scipio::{Async, Local, LocalExecutorBuilder};
 use std::io::Result;
 use std::net::{TcpListener, TcpStream, ToSocketAddrs};
 use std::rc::Rc;
@@ -23,10 +23,12 @@ async fn server(conns: usize) {
     // Not only this will guarantee that we are listening on the port already (so no need
     // for sleep or retry), but it also demonstrates how a more complex application may not
     // necessarily spawn all executors at once running symmetrical code.
-    let client_handle = LocalExecutor::spawn_executor("client", Some(2), move || async move {
-        client(conns).await;
-    })
-    .unwrap();
+    let client_handle = LocalExecutorBuilder::new()
+        .pin_to_cpu(2)
+        .spawn_thread("client", move || async move {
+            client(conns).await;
+        })
+        .unwrap();
 
     let mut servers = vec![];
     for _ in 0..conns {
@@ -79,7 +81,8 @@ fn main() -> Result<()> {
     // Skip CPU0 because that is commonly used to host interrupts. That depends on
     // system configuration and most modern systems will balance it, but that it is
     // still common enough that it is worth excluding it in this benchmark
-    let server_handle = LocalExecutor::spawn_executor("server", Some(1), || async move {
+    let builder = LocalExecutorBuilder::new().pin_to_cpu(1);
+    let server_handle = builder.spawn_thread("server", || async move {
         // If you try `top` during the execution of the first batch, you
         // will see that the CPUs should not be at 100%. A single connection will
         // not be enough to extract all the performance available in the cores.
--- a/scipio/src/executor.rs
+++ b/scipio/src/executor.rs
@@ -344,8 +344,8 @@ impl LocalExecutorBuilder {
 
     /// Names the thread-to-be. Currently the name is used for identification
     /// only in panic messages.
-    pub fn name(mut self, name: String) -> LocalExecutorBuilder {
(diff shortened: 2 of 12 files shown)


glommer:
@HippoBaro please coordinate with @sitano at https://github.com/DataDog/scipio/pull/53

He is about to introduce a new parameter to the executor, and I suggested that it should go on a struct.
It is orthogonal to your changes but y'all will conflict.

Maybe be a good citizen and if you reach agreement on what to do introduce the struct already.

scipio/src/sys/mod.rs
```diff
@@ -114,7 +120,7 @@ pub(crate) enum SourceType {
     Close,
     LinkRings(bool),
     Statx(CString, Box<RefCell<libc::statx>>),
-    Timeout(bool),
+    Timeout(TimeoutStatus),
```
glommer, line 129: I am about to push a change that will conflict heavily with this, and rewrites most of uring. (see https://github.com/DataDog/scipio/issues/52 - > the problem turned out to be quite nasty).

This is still a good change, but we'll conflict and what I am doing is a bit more delicate, so I'd like to defer this

glommer: Note that this is an internal type. The change is still good, but less crucial.

HippoBaro:
@sitano as this PR aims at changing types and function signatures, it makes sense to introduce the `LocalExecutor` configuration struct here. You can then rebase and add any extra configuration knob there for your work in #53. 

I am not settled on a name for this struct so feedback welcome :smile: 

Also, rust seems to prefer using builder APIs to allow for injecting configurations. An example is how the standard library defines [`std::thread::Builder`](https://doc.rust-lang.org/std/thread/struct.Builder.html) to allow things like:

```rust
let builder = thread::Builder::new()
                              .name("foo".into())
                              .stack_size(32 * 1024);

let handler = builder.spawn(|| {
    // thread code
}).unwrap();
```

At first glance this seems overkill but it is certainly future-proof and would probably we nice to have when we eventually implement features such as #32 .

glommer:
@HippoBaro I just pushed my work. As you will see, the boolean is gone from the timer.
That's the main thing I wanted to push first: if you rebase on top of that PR, you will probably drop the timer part altogether. And having the timer patch before I merge, will just introduce conflicts for no reason in a series that is already quite hard.


I like the builder idea. Let's take that to @sitano . Agree that it shouldn't block this here, then

glommer:
However maybe it should influence this PR of yours (the builder tdiscussion).

If we adopt the builder pattern then your first patch becomes unnecessary. It is fine to pass an unnamed `Option` if it is the argument to a function `.with_cpu_pinning`

sitano:
Builder pattern or Config or Config+Builder sounds great to me. 
As an example of how it can look like and be implemented I can offer:
- sled https://github.com/spacejam/sled/commit/ae5c298b0d469586149d0a578707fcfe7395c388
- or tokio::runtime https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/builder.rs

glommer:
Preferably I would like to have that before we publish on crates.io, so existing applications that get written with it don't have to change in the future.

Publication should happen this week if everything goes right.

@sitano if you want to take a stab at it and have the time please work on that first and then we can later rework the other pending issues on top of it.

sitano:
> Preferably I would like to have that before we publish on crates.io, so existing applications that get written with it don't have to change in the future.
> 
> Publication should happen this week if everything goes right.
> 
> @sitano if you want to take a stab at it and have the time please work on that first and then we can later rework the other pending issues on top of it.

@glommer you mean the Builder style configuration for Executors?

glommer:
yes.

This will affect API so best to get right from the start.

HippoBaro:
> @sitano as this PR aims at changing types and function signatures, it makes sense to introduce the LocalExecutor configuration struct here. You can then rebase and add any extra configuration knobs there for your work in #53.

@glommer as I said above I was going to work on that. @sitano feel free to give it a shot but let's just make sure we don't walk on each other's feet :)

sitano:
@HippoBaro cool! Then I will wait until you will do the Builder.

sitano:
@HippoBaro @glommer do you have any kind of a chat room to talk for coordination in RT? maybe Discord / Slack?

glommer:
Unfortunately not at this time.

sitano:
Maybe create one?

glommer:
At the current state of the project I believe this to be more of a hassle.

HippoBaro:
@sitano the new patch series implements the Builder pattern discussed above. I used the standard library `thread::Builder`  as a guide.

@glommer I also dropped the `Timeout` refactor to not interfere with your work in the reactor.

scipio/src/executor.rs
```diff
@@ -293,6 +293,50 @@ impl ExecutorQueues {
...
+impl LocalExecutorBuilder {
+    /// Generates the base configuration for spawning a [`LocalExecutor`], from which configuration
+    /// methods can be chained.
+    pub fn new() -> LocalExecutorBuilder {
+        LocalExecutorBuilder {}
+    }
+
+    /// Spawns a new [`LocalExecutor`] by taking ownership of the Builder, and returns an io::Result
+    /// to its JoinHandle.
+    pub fn spawn(&self) -> io::Result<LocalExecutor> {
```
glommer: shouldn't this be `self` instead of `self` if you are taking ownership of the builder ?

scipio/src/executor.rs
```diff
@@ -302,7 +346,7 @@ impl ExecutorQueues {
 /// ```
 /// use scipio::LocalExecutor;
 ///
-/// let local_ex = LocalExecutor::new(None).expect("failed to create local executor");
+/// let local_ex = LocalExecutor::spawn_default();
```
glommer, line 411: I do not like the name `spawn` for the version that does not create a new thread.

In all of rust when you `spawn` something you are given a `JoinHandle` to wait on. But not here. 
Let's go with `build`, `create` or anything else that does not mention spawn

scipio/src/executor.rs
glommer: I don't like the fact that this returns a `Result`. One of the biggest advantage of this builder route is that if you don't do anything fancy it can't possibly fail so you don't have to check for errors.

Your examples don't check for errors, but I am pretty sure they would generate warnings (maybe there is a way to error out on warnings)

examples/echo.rs
```diff
@@ -5,7 +5,7 @@
 //
 use futures::future::join_all;
 use futures::io::{copy, AsyncReadExt, AsyncWriteExt};
-use scipio::{Async, Local, LocalExecutor};
+use scipio::{Async, Local, LocalExecutorBuilder};
```
glommer, line 8: Hippo, how about we do `Local::Builder`, this way users only have to import `Local` ?

scipio/src/executor.rs
```diff
@@ -941,9 +940,9 @@ impl<T> Task<T> {
     ///
     /// # Examples
     /// ```
-    /// use scipio::{LocalExecutor, Local, Latency};
+    /// use scipio::{LocalExecutor, Local, Latency, LocalExecutorBuilder};
     ///
-    /// let ex = LocalExecutor::spawn_executor("example", None, || async move {
+    /// let ex = LocalExecutorBuilder::new().spawn_thread("example", || async move {
```
glommer, line 945: if we have a builder API, then I think we should pass the name to it as well. That would allow us to create unnamed executors, which is fine for many applications, and simpler.

glommer (changes requested):

glommer:
On top of the comments I left, this breaks `cargo doc --all-features`.

scipio/src/executor.rs
HippoBaro: We have to, because `spawn` can fail. All configuration methods on the builder are lazy: they never fail. If you pass an invalid value for CPU binding, `spawn` will fail.

examples/echo.rs
HippoBaro: We have established that `Local` is a module with convenient methods when inside an executor. Most of its functions fail outside of an executor.
It would be confusing to put the builder there. Also, the name wouldn't be self-explanatory.

Ideally, we would have `scipio::executor::Builder` just like rust has `std::thread::Builder`. However we collapse the `executor` module in `scipio` at the moment.

scipio/src/executor.rs
glommer: Right. But read that together with my previous comments about this being called spawn.

What we really need is a `create` function that doesn't allow any builder parameters.

This is similar to what `thread` does (except that they are indeed spawning, so theirs is called `spawn`.
You can get started really quickly by just calling `Local::create_executor()`, which doesn't ever return any errors.

For more complex scenarios you go to through the builder and that can return errors

examples/echo.rs
glommer: ok, fair enough.

scipio/src/executor.rs
HippoBaro: Alright I see what you mean. There is such a function: `LocalExecutor::make_default() -> LocalExecutor`. This is not configurable, uses the current thread and cannot fail.

The function above sits inside the builder so at this point some of the configuration the user provided may be faulty. 
For context this method as since been renamed `make()`. Another function `spawn()` exists too. It behaves the same way except that it also create a new thread to host the executor. It returns `io::Result<JoinHandle<()>>` as you'd expect.

scipio/src/executor.rs
HippoBaro: Agreed, new naming scheme:
* `make` variants returns the `LocalExecutor`
* `spawn` variants spawn threads and return `JoinHandle`s

examples/echo.rs
```diff
@@ -25,7 +25,8 @@ async fn server(conns: usize) {
     // necessarily spawn all executors at once running symmetrical code.
     let client_handle = LocalExecutorBuilder::new()
         .pin_to_cpu(2)
-        .spawn("client", move || async move {
+        .name(String::from("client"))
```
glommer, line 28: why can't you pass a `&str` to the name, and have the name function internally convert to a string ?

glommer:
Approved, but consider changing `name()` to accept a `&str`

glommer (approved):

sitano:
I am a bit worried that the created Executor does not have a ref to the configuration it was created with. For example, I can't read if the thread was pinned. It's not an issue right now, but maybe its also worth saving a Builder configuration into an Executor on make().

PR #57, opened 2020-09-13
shared: a cheaper version of Rc<RefCell<>>

glommer:
shared: a cheaper version of Rc<RefCell<>>

Daniel has been working to add linters to make sure that a borrow() on a
RefCell never crosses a suspension point.

Still, I've been thinking of ways to not pay the cost of the runtime check.
This proposal implements such a way: by allowing scoped access to the interior
of an Rc we should be able to guarantee that accesses are safe.

That is done through a new struct, Shared, which resembles an Rc. However
 it will not implement Deref, meaning we can't get a reference to it outside
 the scope of a closure passed to it.

 The closure that we expect is not async, meaning we cannot have a suspension
 point inside it, and because all spawn-like functions expect a 'static lifetime
 and the reference is only valid inside the closure we can't spawn either.

### DIFF
--- a/scipio/src/shared.rs
+++ b/scipio/src/shared.rs
@@ -1,0 +1,344 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
+// MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
+//
+use std::cell::UnsafeCell;
+use std::rc::{Rc, Weak};
+
+/// A shared pointer with contents that are safe for Scipio users to access.
+///
+/// It is similar to [`Rc`]<[`RefCell`]<T>>, except that it does not perform runtime checks on the
+/// inner cell, and the reference counted pointer does not implement [`Deref`].
+///
+/// Because Scipio does not use helper threads a particular shared pointer is always ever used by a
+/// single CPU, which provides natural serialization. There are no other users as long as it
+/// doesn't cross a suspension point.
+///
+/// Because [`Deref`] is not implemented the internal contents can only be accessed with the
+/// [`do_with`]. It is illegal to have any suspension points in code that is passed to [`do_with`]
+///
+/// [`Rc`]: https://doc.rust-lang.org/beta/std/rc/struct.Rc.html
+/// [`RefCell`]: https://doc.rust-lang.org/std/cell/struct.RefCell.html
+/// [`Deref`]: https://doc.rust-lang.org/std/ops/trait.Deref.html
+/// [`do_with`]: struct.Shared.html#method.do_with
+#[derive(Debug, Clone)]
+pub struct Shared<T> {
+    inner: Rc<UnsafeCell<T>>,
+}
+
+/// A weak pointer with contents that are safe for Scipio users to access.
+///
+/// It is similar to [`Weak`]<[`RefCell`]<T>>, except that it does not perform runtime checks on the
+/// inner cell, and the reference counted pointer does not implement [`Deref`]. It upgrades to a
+/// [`Shared`] and must be upgraded before using.
+///
+/// [`Weak`]: https://doc.rust-lang.org/beta/std/rc/struct.Weak.html
+/// [`RefCell`]: https://doc.rust-lang.org/std/cell/struct.RefCell.html
+/// [`Shared`]: struct.Shared.html
+/// [`Deref`]: https://doc.rust-lang.org/std/ops/trait.Deref.html
+#[derive(Debug, Clone)]
+pub struct WeakShared<T> {
+    inner: Weak<UnsafeCell<T>>,
+}
+
+impl<T> Default for WeakShared<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T> WeakShared<T> {
+    /// Constructs a new `WeakShared<T>`, without allocating any memory. Calling upgrade on the return value always gives [`None`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use scipio::WeakShared;
+    /// let empty: WeakShared<i64> = WeakShared::new();
+    /// assert!(empty.upgrade().is_none());
+    /// ```
+    ///
+    /// [`None`]: https://doc.rust-lang.org/beta/std/option/enum.Option.html#variant.None
+    pub fn new() -> WeakShared<T> {
+        WeakShared { inner: Weak::new() }
+    }
+
+    /// Attempts to upgrade the WeakShared pointer to a [`Shared`], delaying dropping of the inner value if successful.
+    ///
+    /// Returns [`None`] if the inner value has since been dropped or if this was created through
+    /// [`WeakShared::new`]
+    ///
+    /// # Examples
+    /// ```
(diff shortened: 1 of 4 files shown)


glommer:
@Daniel-B-Smith please take a look a this and let me know what you think.

I haven't bothered to write any docs or examples so far but you should be able to follow with the top level comments.

Daniel-B-Smith:
All of this comes with the obvious caveat that reasoning about `unsafe` is hard, so it's quite possible I missed something.

I think this is a good API and could be valuable even outside of Scipio. Since `Shared` is `!Send` it's impossible to call `do_with()` concurrently between threads. Since all access to the the inner mutable state is guarded by `do_with()`, it's impossible to do weird things across `await` suspension points because of the requirement that `R` be `Copy`. It would be nice to relax the `Copy` constraint by forbidding `Future` more directly ([this](https://users.rust-lang.org/t/a-macro-to-assert-that-a-type-does-not-implement-trait-bounds/31179) came up in a quick search). As an initial step, this API looks sound to me with the `Copy` constraint, and any weaker bound is going to require substantially more care.

glommer:
I agree that restricting just Future would be better. The macro you linked to seems like complete and total black magic so it'll take me a bit longer to understand, but it seems awesome. Indeed I am not fully happy with the restriction of !Copy

HippoBaro (dismissed):
CI has been updated since this PR was opened, please rebase on master :bowing_man:

Daniel-B-Smith:
Upon further review, I think the `Copy` / `!Future` restrictions aren't necessary. What we really care about is that there is no way to store the reference anywhere. We can still create a `Copy` `Future` with a (non-mut) reference, but the borrow checker won't let you hold on to the reference passed to you by `do_with`. Here's the borrow checker yelling at you for an explicit `Copy` future:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e755f873cbfdf9e5ba535073c932f02f

and here's the borrow checker yelling at you without the `Copy` restriction:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f098d1032e9bd044a315a625f3991e8b

glommer:
All right, I played with this a bit more yesterday.

Awesome about the `Copy`: That will make everything easier.
I was trying to think about ways to abuse this, so we can protect against, and there are two potential issues. However as I will explain I think they won't be problems in practice:

1. Someone could `spawn` a task inside `do_with`, and somehow pass the reference to it which would create a lifetime issue. 

I don't think this will be an issue in practice for two reasons:
  * a) we create a mutable reference to pass to the closure. If you would create a new mutable reference inside the `spawn` closure, that would be disallowed. If you try to get an immutable borrow, that would be disallowed too by the borrow checker rules.

 * b) Pretty much all `spawn` variants everywhere I looked (tokio, smol, core rust, whatever) will require a `'static` bound. If you remember we discussed relaxing that for `Local::local`, but playing with that in this context yesterday convinced me we should *not* relax it: even if you want to somehow track liveness between the spawn point and the `.await` point, nothing prevents a task from being `detach`ed. At which point `'static` really is the only option. So if we accept that `Local::local` and its variants will always accept a future with `'static` bound, we're safe against this.

2. There are other places very deep in Scipio where we the guarantee of "we're in the same CPU and won't preempt" won't hold: those are calls to `wake()`,`task::run()` and `task::schedule()`. That is because those go into the Reactor core and can change what is running. Borrows are not safe to cross those either, just like .await.

Unfortunately I don't know of a good way to *guarantee* that: nothing prevents you from calling one of those inside `do_with` because trait bounds are applied to the types passed to and returned from `do_with`. However the main reason for this API to exist is for scipio *users* to be able to consume it. Inside scipio we can just leave comments and do our best not to make those mistakes. Again, holding borrows across those points are also invalid and we can't check them (I think not even your clippy linter would check them) so it's not like we're creating a new problem.

Let me know what you think and I will move forward with introducing this officially this week

HippoBaro:
@glommer would it be possible to add a runtime assertion? We could set a local flag upon entering `do_with` and panic if it is set to true while in the reactor. I haven't thought this through however so I may be missing something.

glommer:
@HippoBaro while technically possible, a runtime assertion defeats the purpose of this API: if you are willing to pay the price of a runtime check, then you're at the same cost of a RefCell use.

Also very likely any check of "we're in the reactor" would involve TLS variables which would effectively make this more expensive than the RefCell.

glommer:
@Daniel-B-Smith `block_on` inside a living reactor is horrible, because it will essentially sleep making sure no other task queues run. As a matter of fact if there was any way to prevent that, we should!

The fact that we are not `!Send` is also baked deep down in Scipio's thread per core model (so far everything is `!Send`, when we have channels data structures that go over channel will be `Send`).

I like the idea of marking this as unstable for the first version. I will only consider scipio to be API-stable when we land on crates, which should hopefully happen very very soon

HippoBaro (approved):
+1 to mark this as an unstable API

scipio/src/shared.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// x.do_with(|x| assert_eq!(*x, 2) );
+    /// ```
+    pub fn do_with<F, R>(&self, func: F) -> R
+    where
+        F: FnOnce(&mut T) -> R,
+        R: Sized,
+    {
+        let k = unsafe { &mut *self.inner.get() };
+        func(k)
+    }
```
matklad, line 185: This API is fundamentally unsound due to re-entrancy:

```rust
#[test]
fn unsound() {
    let cell = Shared::new("hello world".to_string());
    cell.do_with(|r1| {
        let s = r1.as_str();
        cell.do_with(|r2| {
            *r2 = String::new();
            eprintln!("{:?}", s)
        })
    })
}
```

I *think* this fundamentally requires `RefCell`-like runtime flag to make it sound.

Rule-of-thumb: be extra careful if calling user-supplied code (`FnOnce`) near `unsafe`.

matklad: Minor addendum: idiomatic name would be just `with`: https://doc.rust-lang.org/stable/std/thread/struct.LocalKey.html#method.with

glommer: If this requires a `RefCell`-like runtime flag to work, then it is a pointless API since the whole idea of this is to experiment with ways to do away with the `RefCell` runtime check =)

If there is no way to prevent reentrancy, I agree that your example provides an example of something that could break this.

This leaves us with the other methods in `Shared`: instead of exposing `with` we can use that internally to update a value and return a value if the type implements `Copy`. (`swap` and `inner`)

I don't see a lot of upside in `inner` returning the value since it is already possible to do that without a `RefCell`. But 
`swap` still seems to me like a useful API as it allows you to mutate the contents in a controlled way (and it actually doesn't even need `Copy` as it is just moving the ownership)

So I think the direction I am going with this will probably be making `with` private and exposing just `swap` and `inner`.

glommer:
I am retiring this to avoid cluttering the list with PRs we won't work on.

The advantage of this interface if restricted to `Copy` types are dubious since a `Cell` would do the same job without the runtime `RefCell` costs

I think there is value in a similar interface that wraps around common types:
- For example, a SharedVec that allows one to insert elements into an array as well as consume an element safely

But the interface would likely be different for each collection anyway.

PR #63, opened 2020-09-15
source: move most of the source code the io_uring file.

glommer:
And make its methods fn instead of pub(crate). The code before
lived in mod.rs, closer to its definition point. However it does
some nasty unsafe things to account for the fact that io_uring is
an asynchronous interface, and we're associating a Source to each
request.

The functions that return references are all internal now, and the
ones that are pub(crate) are all returning copies of parts of the source
the external entities need to manipulate (type, raw fd, etc)

Also add a comment about why the stuff clippy doesn't like is fine

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -17,7 +17,7 @@ use std::task::Waker;
 use std::time::Duration;
 
 use crate::sys::posix_buffers::PosixDmaBuffer;
-use crate::sys::{mut_source, InnerSource, LinkStatus, PollableStatus, Source, SourceType};
+use crate::sys::{InnerSource, LinkStatus, PollableStatus, Source, SourceType};
 use crate::{IoRequirements, Latency};
 
 use uring_sys::IoRingOp;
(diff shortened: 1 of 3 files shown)


scipio/src/sys/uring.rs
```diff
@@ -489,6 +489,75 @@ impl UringCommon for PollRing {
...
+// and the user is free to do whatever it wants with it (as opposed to a scoped RefMut).
+//
+// We don't have too much of a choice because the Source accompanies a request in
+// the io_uring and we'd be dealing with raw pointers anyway. We want to make this as
+// safe as possible but will always have to rely on those weird tricks.
+//
+// This should only really be used within the io_uring code
+fn mut_source(source: &Rc<UnsafeCell<InnerSource>>) -> &mut InnerSource {
+    unsafe { &mut *source.get() }
+}
```
HippoBaro, line 510: This function could be inlined in `Source::inner`, this would limit the number of individual clippy `allow` attributes to one. Otherwise the above comment should also be added to  `Source::inner` too.

scipio/src/sys/uring.rs
```diff
@@ -489,6 +489,75 @@ impl UringCommon for PollRing {
...
+        self.inner().update_source_type(source_type)
+    }
+
+    pub(crate) fn extract_source_type(&mut self) -> SourceType {
+        self.inner().update_source_type(SourceType::Invalid)
+    }
+
+    pub(crate) fn wakers(&self) -> &RefCell<Wakers> {
+        &self.inner().wakers
+    }
```
HippoBaro, line 545: Your commit message states that:
```
The functions that return references are all internal now, and the
ones that are pub(crate) are all returning copies of parts of the source
the external entities need to manipulate (type, raw fd, etc)
```
But this function is public _and_ returns a reference.

HippoBaro (changes requested):

scipio/src/sys/uring.rs
glommer: right, but that's a reference to a RefCell, meaning access will happen through the usual borrows

It is still bad wording so I will change.

scipio/src/sys/uring.rs
glommer: we won't be able to get rid of the two clippy markings. The function is used many other times within uring. It is not used just to implement `Source::inner`. And because it is a method of `UnsafeCell<T>` and not `T` it has to be free-floating

scipio/src/sys/uring.rs
glommer: I went one step further and actually changed the code so that the reference is not returned.
It costs us an extra borrow_mut(), but it's better layering and better isolation so I think it is worth it

glommer: @HippoBaro please take a look at the end result

HippoBaro (approved):
LGTM!

PR #65, opened 2020-09-18
fix make_test_directories to support /tmp

sitano:
StorageMedia and TempFs can conflict if they target the same path.
This PR adds support for `env SCIPIO_TEST_POLLIO_ROOTDIR=/tmp cargo test`

### DIFF
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -506,6 +506,11 @@ mod test {
                 let mut dir = PathBuf::from(path);
                 std::assert!(dir.exists());
 
+                if std::env::temp_dir() == dir {
+                    dir.push("storage_media");
+                    std::fs::create_dir_all(&dir).unwrap();
+                }
+
                 dir.push(test_name);
                 let _ = std::fs::remove_dir_all(&dir);
                 std::fs::create_dir_all(&dir).unwrap();


glommer:
Ivan,

It is not fully clear to me why this fix is needed.

Is the issue that your `SCIPIO_TEST_POLLIO_ROOTDIR` environment variable points to `/tmp`? 
If that is the case, can't you just point it out to to somewhere else?

sitano:
yes, thats exactly what this PR do: append /media to /tmp if its /tmp. feel free to close if you don't like the fix. It was just very confusing to me to get all those errors.

glommer:
I think could do a better job here, but I am not sure if this is the right fix.
@HippoBaro , what do you say ?

@sitano am I right in understanding that your /tmp in this case is not on tmpfs?
Maybe the best thing to do in this case is to skip the tmpfs tests if we detect that /tmp is on hard media.

sitano:
@glommer as of yesterday /tmp was on tmpfs

sitano:
> I think could do a better job here, but I am not sure if this is the right fix.

For me the problem with `env SCIPIO_TEST_POLLIO_ROOTDIR=/tmp` was that both pieces of tests dir creation code try to erase create the same path.

HippoBaro:
`SCIPIO_TEST_POLLIO_ROOTDIR ` should be used to specify a test directory that supports poll io. tmpfs does not (that's the reason why we introduced this mechanism in the first place). I feel like in this case the fix is to set the envvar it to something else.

I agree that our unit test setup (irt file io with and w/o polling) is not great.

sitano:
Ok. I think it's proper then if I close it for now.

PR #67, opened 2020-09-18
do not use our own Error type for I/O operations

glommer:
io::Result is pretty much the standard everywhere else for errors resulting
from I/O operations.

I had introduced our own type earlier out of frustration for the fact that
io::Error wouldn't encode information that I considered valuable about location
of errors like the file path.

I had since discovered that this information can be added as an inner component
to io::Error. Doing that is superior to returning our own Error type because it
is easier to deal with in situations where you want to mix both. It is also
less surprising for users who are used to io::Error.

The Error struct is kept (now called ErrorEnhancer) as a convenience so we can
add the desired information to io::Error. But it is made pub(crate) so it cannot
be used outside Scipio.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -3,11 +3,10 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
-use crate::error::Error;
+use crate::error::ErrorEnhancer;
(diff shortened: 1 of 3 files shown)


HippoBaro (approved):
LGTM

PR #68, opened 2020-09-18
File stream

glommer:
### What does this PR do?

Adds Sequential access facilities for files, implementing AsyncRead and AsyncWrite

### Motivation

For users who do not do random access, this API will be more convenient, not to mention more familiar.

### Checklist


glommer:
@HippoBaro please take a look

scipio/src/file_stream.rs
```diff
@@ -0,0 +1,1338 @@
...
+    ///
+    /// let ex = LocalExecutor::make_default();
+    /// ex.run(async {
+    ///     let file = DmaFile::open("myfile.txt").await.unwrap();
+    ///     let mut reader = StreamReaderBuilder::new(file).build();
+    ///     assert_eq!(reader.current_pos(), 0);
+    ///     reader.skip(8);
+    ///     assert_eq!(reader.current_pos(), 8);
+    ///     reader.close().await.unwrap();
+    /// });
```
HippoBaro, line 458: This example does not demonstrate how to use `get_buffer_aligned `

HippoBaro (changes requested):
Not finished reviewing yet. Can you please break your last commit into two (one part reading, one part writing) in the meanwhile?

glommer:
@HippoBaro this is now split in two

scipio/src/file_stream.rs
glommer: I will add a commit fixing this

examples/defer.rs
```diff
@@ -0,0 +1,12 @@
...
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
+//
+use scipio::defer;
+fn main() {
+    defer! {
+        println!("This will print last!");
+    }
+    println!("This will print first");
+}
```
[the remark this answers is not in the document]
glommer: I added a commit with this.

scipio/src/file_stream.rs
```diff
@@ -615,12 +615,355 @@ impl Future for PrepareBuffer {
...
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut state = self.state.borrow_mut();
+        if let Some(err) = current_error!(state) {
+            return Poll::Ready(err);
+        }
+
+        let mut written = 0;
+        while written < buf.len() {
+            match state.buffer_queue.pop_front() {
```
HippoBaro, line 897: If I understand correctly, if my `StreamWriter` was configured with `n` write-behind buffers and I try to issue `n+1` write operations (assuming my writes are of the same buffer size), then the CPU will spin until a buffer is available (i.e until one of my previous writes has completed)?

Assuming this is correct, why not use our own async dequeue here?

scipio/src/file_stream.rs
```diff
@@ -909,4 +1269,64 @@ mod test {
...
+    });
+
+    file_stream_write_test!(write_close_twice, path, _k, filename, file, {
+        let mut writer = StreamWriterBuilder::new(file)
+            .with_buffer_size(128 << 10)
+            .build();
+
+        writer.close().await.unwrap();
+        expect_specific_error!(writer.close().await, "\"Bad file descriptor (os error 9)\"");
+    });
```
HippoBaro, line 1332: It would be nice to have some extra tests. Some ideas:
* assert error propagate well when trying to create a `StreamWriter` with a read-only file
* assert gentle failure when creating an invalid `StreamWriter` (`with_write_behind(0)` and/or `with_buffer_size(0)`)
* assert we fail when issuing multiple writes from different `StreamWriter`

HippoBaro (commented):
This is looking great!

scipio/src/file_stream.rs
HippoBaro: Those comments probably apply to the `StreamReader` to an extent as well.

scipio/src/file_stream.rs
glommer: That's note quite right.

First the whole point of this API is that you don't try to issue any operations. You just write bytes linearly to the file without worrying much about what the storage is doing. Writes are done to the storage as necessary.

But if you have already `n` in-flight buffers and need to write another, it doesn't spin. It just returns `Poll::Pending`, which tells the rust core to go do something else.

The AsyncDeque is a good way to implement this in contexts outside `poll`. But inside `poll` you just return a status saying whether or not you are ready.

scipio/src/file_stream.rs
glommer: The first two suggestions are good.
The Third, I fail to see the point. It is indeed possible, although weird, to use the same file for two `StreamWriter`. How could we possibly defend against that?

For readers it is more than that: you actually may want many stream readers into the same file.

glommer: About the second suggestion, though, I think those things should be considered hints and not hard values. If you pass zero the implementation should round them up to a reasonable number. Same thing with buffer sizes that are not powers of two.

However that still warrants a test.

scipio/src/file_stream.rs
```diff
@@ -667,7 +667,7 @@ impl StreamWriterBuilder {
     ///
     /// [`StreamWriter`]: struct.StreamWriter.html
     pub fn with_write_behind(mut self, write_behind: usize) -> Self {
-        self.write_behind = write_behind;
+        self.write_behind = std::cmp::max(write_behind, 1);
```
HippoBaro, line 671: This is a good idea. We should advertise than a value of `0` means "left to the system to decide" in the documentation.

scipio/src/file_stream.rs
```diff
@@ -682,6 +682,8 @@ impl StreamWriterBuilder {
     ///
     /// [`StreamWriter`]: struct.StreamWriter.html
     pub fn with_buffer_size(mut self, buffer_size: usize) -> Self {
+        let buffer_size = std::cmp::max(buffer_size, 1);
+        self.buffer_size = self.file.align_up(buffer_size as _ ) as usize;
```
HippoBaro, line 687: Same thing here, this is a sane behaviour and we should advertise it in the documentation.

HippoBaro (approved):

scipio/src/file_stream.rs
HippoBaro: For 3, I was referring to this:
https://github.com/DataDog/scipio/pull/68/commits/eb26095df5df79e4ccb9253648bc41d3c60b7a8a#diff-b7113f50e181897a934ae801f9253b34R719-R723

How do we trigger this in a unit test?

scipio/src/file_stream.rs
glommer: I don't think `0` means left the system to decide. Not calling this means left to the system to decide.

What we are doing here is just enforcing sane boundaries.

scipio/src/file_stream.rs
HippoBaro: Alright, thanks for the explanation :bow:

scipio/src/file_stream.rs
glommer: I don't think we can. That's an assert condition, it should not be possible unless we coded something wrong. Seems beyond the scope of a unit test since if I knew I had coded anything wrong, I would have fixed it ! =)

HippoBaro:
Before merging, please fix the typo in your commit message. It also looks like CI doesn't like your last commit.

glommer:
Which typo are you referring to ?

I will do a rebase and rustfmt pass before merging.

HippoBaro:
> Which typo are you referring to ?
> 
> I will do a rebase and rustfmt pass before merging.

a31ef80: 'test zero-sized red buffers, and read ahead = 0`

glommer:
That commit is temporary and will be folded into the previous ones.

glommer:
Pushed a rebased version. Empty diff from 9a0716f

PR #71, opened 2020-09-28
implement two new operations for the StreamWriter

glommer:
current_flushed_pos() will allow us to ask the writer how much data
is already flushed and made into the disk. New readers are guaranteed
to see this data except if there is a crash.

sync() is a crash-safe sync operation (in this implementation fdatasync)
that will be guaranteed to flush the data all the way down to the
underlying storage, and return which position is already safely stored.

### Checklist

### DIFF
--- a/scipio/src/file_stream.rs
+++ b/scipio/src/file_stream.rs
@@ -13,6 +13,7 @@ use futures::io::{AsyncRead, AsyncWrite};
 use futures::task::{Context, Poll};
 use futures_lite::Future;
 use std::cell::RefCell;
+use std::cmp::Reverse;
 use std::collections::{HashMap, VecDeque};
 use std::io;
 use std::pin::Pin;
@@ -713,6 +714,11 @@ struct StreamWriterState {
     pending: Vec<task::JoinHandle<(), ()>>,
     buffer_queue: VecDeque<Rc<DmaBuffer>>,
     file_pos: u64,
+    // this is so we track the last flushed pos. Buffers may return
+    // out of order, so we can't report them as flushed_pos yet. Store for
+    // later.
+    out_of_order_write_returns: Vec<u64>,
+    flushed_pos: u64,
     buffer_pos: usize,
     write_behind: usize,
     flush_on_close: bool,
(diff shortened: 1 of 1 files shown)


scipio/src/file_stream.rs
```diff
@@ -775,6 +785,27 @@ impl StreamWriterState {
         self.file_pos + self.buffer_pos as u64
     }
 
+    fn flushed_pos(&self) -> u64 {
+        self.flushed_pos
+    }
+
+    fn adjust_flushed_pos(&mut self, just_written: u64) {
+        self.out_of_order_write_returns.push(just_written);
+        self.out_of_order_write_returns.sort_by_key(|x| Reverse(*x));
```
HippoBaro, line 794: What is the runtime performance impact for this? Can we use binary search instead and insert directly at the right index?

scipio/src/file_stream.rs
```diff
@@ -793,6 +824,7 @@ impl StreamWriterState {
                 }
 
                 let mut state = state.borrow_mut();
+                state.adjust_flushed_pos(file_pos);
```
HippoBaro, line 827: Is there a way to make this lazy? I'm concerned about the cost of this for writers that will never care about this information.

HippoBaro (changes requested):
Please split your two ops as two commits.

scipio/src/file_stream.rs
glommer: That's a good point. I considered a binary search but gave up because sorting is simpler and this array is as big as the number of in-flight buffers. That's usually something like 10. Not worth fretting too much about, in my opinion.

Let me know if you agree / disagree.

scipio/src/file_stream.rs
glommer: Usually I am not too concerned about bookkeeping for writes: writes come in 128kB requests that take a lot of time to come back.

How would you do it lazily, though? I am open if you have a good suggestion.
The `file_pos` advances as we initiate the flush, so that is not good enough. We can keep an array of returned `file_pos` but if we don't trim it here it will keep growing which is worse: you win in time for the insert but lose in memory utilization.

scipio/src/file_stream.rs
HippoBaro: If this vectors stays small then sorting is fine, no objections.

glommer: Can you then move from changes requested to approved ?

HippoBaro: No, that's not part of the requested changes, what I want is for you to split your patch into two. See above :arrow_up:

glommer:
In this particular case, I don't believe the benefits of splitting would be justified.

PR #74, opened 2020-09-28
facility to allow testing if two files are the same

glommer:
file1 == file2 tests for object equality: it will return true if
the two files are the same file descriptors. However it is possible for:

- hard links to exist
- open a file more than once.

In which case they will represent the same on-storage entity but will be
represented by two different objects. This patch adds a facility that
allows us to test for that.

### DIFF
--- a/scipio/src/dma_file.rs
+++ b/scipio/src/dma_file.rs
@@ -191,6 +191,9 @@ pub struct DmaFile {
     path: Option<PathBuf>,
     o_direct_alignment: u64,
     pollable: PollableStatus,
+    inode: u64,
+    dev_major: u32,
+    dev_minor: u32,
 }
 
 impl DmaFile {
@@ -233,6 +236,9 @@ impl Default for DmaFile {
             path: None,
             o_direct_alignment: 4096,
             pollable: PollableStatus::Pollable,
+            inode: 0,
+            dev_major: 0,
+            dev_minor: 0,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


scipio/src/dma_file.rs
```diff
@@ -251,6 +257,31 @@ I will close it and turn a leak bug into a performance bug. Please investigate",
...
+    ///     let mut wfile = DmaFile::create("myfile.txt").await.unwrap();
+    ///     let mut rfile = DmaFile::open("myfile.txt").await.unwrap();
+    ///     // Different objects (OS file descriptors), so they will be different...
+    ///     assert!(wfile != rfile);
+    ///     // However they represent the same object.
+    ///     assert!(wfile.is_same(&rfile));
+    ///     wfile.close().await;
+    ///     rfile.close().await;
+    /// });
+    /// ```
```
HippoBaro, line 278: Good documentation but it would be befenicial to explain how doing this differs from comparing fds. Especially irt to symbolic links. Your existing documentation sorta does it already, but would gain from being more explicit.

HippoBaro (changes requested):
I like this addition. However I'd like to know your opinion on removing the trait `PartialEq` from `DmaFile`. My reasoning is that it is confusing (does it memcmp? does it compare fds? does it support links?, etc.). 

Let's follow the standard library in this case: https://doc.rust-lang.org/std/fs/struct.File.html#trait-implementations. It doesn't implement the trait either, probably for the same reasons. If users want to compare fds, they are publicly accessible.

scipio/src/dma_file.rs
glommer: I am open to removing PartialEq.

I can also add a mention to symlinks (this would return false for a symlink).

glommer: Both changes are now done.

HippoBaro (approved):

PR #75, opened 2020-09-28
need_preempt: allow users to use need_preempt() directly

glommer:
We have been enclosing need_preept() into a helper yield_if_needed()
that returns a future, allowing the user the convenience of .awaiting
directly on it.

However there are situations where we can't really .await, for example
if we have a borrow. For those, export need_preempt() directly. It
comes with an example of how to use each in the examples directory.


HippoBaro (approved):
Thanks for including an example! This is one of these APIs that really need one.

PR #82, opened 2020-10-02
Cleanup ErrorEnhancer

matklad:
### What does this PR do?

Minor cleanups.

### Motivation

### Checklist

### DIFF
--- a/scipio/src/error.rs
+++ b/scipio/src/error.rs
@@ -7,7 +7,7 @@ use std::fmt;
 use std::os::unix::io::RawFd;
 use std::path::PathBuf;
 
-/// Augments an io::Error with more information about what was happening
+/// Augments an `io::Error` with more information about what was happening
 /// and to which file when the error ocurred.
 pub(crate) struct ErrorEnhancer {
     pub(crate) inner: std::io::Error,
@@ -25,11 +25,11 @@ impl fmt::Debug for ErrorEnhancer {
 impl fmt::Display for ErrorEnhancer {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}, op: {}", self.inner, self.op)?;
-        if let Some(path) = &self.path.as_ref().and_then(|x| x.to_str()) {
-            write!(f, " path {}", path)?;
+        if let Some(path) = &self.path {
+            write!(f, " path {}", path.display())?;
         }
 
-        if let Some(fd) = &self.fd {
+        if let Some(fd) = self.fd {
             write!(f, " with fd {}", fd)?;
         }
         Ok(())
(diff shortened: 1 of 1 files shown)


glommer (approved):

glommer:
Hi @matklad 

Tests are broken as of today, and I bisected it to this commit (thanks for keeping everything bisectable!)

```
failures:
    io::file_stream::test::read_close_twice
    io::file_stream::test::read_wronly_file
    io::file_stream::test::write_with_readable_file
```

Those tests are checking for specific error formats so it makes sense. Reading through the commit is likely that the tests are just a bit too strict with quotes and are now breaking since you changed the format to `display`. It's more than fair to just patch the tests.

Unfortunately CircleCI does not have a new enough kernel so that we can run the integration tests so the CI jobs for the moment are only enforcing formatting , lint, etc.

I kind of hate that, but until that is fixed we're kind of doing an honor system thing 

Please send a fix, thanks

scipio/src/error.rs
```diff
@@ -38,6 +38,6 @@ impl fmt::Display for ErrorEnhancer {
 
 impl From<ErrorEnhancer> for std::io::Error {
     fn from(err: ErrorEnhancer) -> std::io::Error {
-        std::io::Error::new(err.inner.kind(), format!("{}", err.inner))
```
glommer, line 41: @matklad the problem is in this line.

The Enhanced Error mechanism depends on the io::Error having an extra string attached that is consumable through `into_inner`.

I will send a fix that reverts just this part.

glommer: All right, and while working on that I found out that the enhancer wasn't *really* doing what it was supposed to do anymore.

That's on me for having indirect unit tests for it but not specific, in-file unit tests.

I'll commit a fix.

matklad: Yeah, re-reading the code, I’ve also noticed that it doesn’t work quite as intended. I have it on my todo list to refactor this Code a bit, usually such error enhancing is done via an extension trait. See this for an example of the pattern: https://docs.rs/anyhow/1.0.33/anyhow/trait.Context.html.

glommer: thanks. I was not aware of this crate.
(Jesus, there are so many crates)

For now it's important that this keep working so the fixes I pushed should do, and in the future I am definitely open to do this in a more standardized way.

matklad: To be clear, I think we shouldn’t use anyhow crate; we might adapt the with_context pattern though, to replace both EnchancedError struct and the enchaced_try macro (yes, I do dislike macros :) )

The question of how the error type should look is also an interesting one! I think the current pattern with io::error might be the right one for scipio.

glommer: Agree about macros.
Sometimes I find them to be the right thing to do: I've been using them a lot in place of what would be async traits (even if you can use the async trait crate it forces an allocation on you and it's generally not great for now) and also been quite helpful in test code.

But so far your changes to convert some of them to functions are very welcome

PR #83, opened 2020-10-02
Fix Deque::push_back

matklad:
### What does this PR do?

Fix an obvious bug. This probably needs a test, but I don't have a motivation for adding that right now :-) 

### Checklist


glommer (approved):

glommer:
Indeed we're lacking tests for the deque.
It's on my todo list.

Thanks!

PR #84, opened 2020-10-03
Cleanup: replace a macro with an equivalent function

matklad:
### What does this PR do?

Strength reduction -- one doesn't need a macro where a function call would do.

### Motivation

I've noticed that scipio has a higher-than-average number of macros. I would like to push the code style to a less macro-heavy side :) I am not a fan of macros for two reasons:

* they make reading code more difficult. When I read plain code, I can get a general picture of what is happening, even if I am not familiar with specific helper functions called. When I read code inside the macro call, I have no idea about what is happening until I consult the declaration/docs of the macro
* macros break ide support in [interesing](https://github.com/matklad/proc-caesar) [ways](https://rust-analyzer.github.io/blog/2020/03/30/macros-vs-rename.html). While we should probably just poke those lazy IDE writers to fix their stuff, it wouldn't hurt to use less macros while IDEs are not yet there :D 

Now, I am not sure if macros in scipio are essential, but I have a hunch that many of them are gratuitous and can be replaced with functions and traits. Here's one specific I've found. 

### Checklist

### DIFF
--- a/scipio/src/io/file_stream.rs
+++ b/scipio/src/io/file_stream.rs
@@ -1376,21 +1376,20 @@ mod test {
         reader.close().await.unwrap();
     });
 
-    macro_rules! expect_specific_error {
-        ( $op:expr, $err:expr ) => {
-            match $op {
-                Ok(_) => panic!("should have failed"),
-                Err(x) => {
-                    let kind = x.kind();
-                    match kind {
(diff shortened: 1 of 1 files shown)


glommer (approved):
Agree that function is better, although for a test function it shouldn't matter much.

Still better is better than worse.

PR #85, opened 2020-10-03
Remove unused futures import

matklad:
### What does this PR do?

Removes 3rd party API in favor of stdlib one

### Additional Notes

I've noticed that scipio uses both `futures` and `futures-lite`. This seems unfortunate, I suggest sticking to the `-lite` version. 

### Checklist


glommer:
Thanks. This is likely because in the very early days of scipio I was using code from external crates that evolved into our first version...
Applying

glommer (approved):

PR #86, opened 2020-10-03
Use type-safe WaiterId more

matklad:
### What does this PR do?

Replaced untyped ints with already existing newtypede id

### Motivation

General code cleanup

### Checklist

### DIFF
--- a/scipio/src/semaphore.rs
+++ b/scipio/src/semaphore.rs
@@ -11,12 +11,12 @@ use std::io::{Error, ErrorKind, Result};
 use std::pin::Pin;
 use std::rc::Rc;
 
-#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


glommer (approved):
Good Change.

I think if you saw my first implementation of Semaphore you would refuse to speak to me ever again. If it was very C++ish and not at all Rusty. With time it evolved but I never came back to finish the deal.

Thanks a lot!

matklad:
>I think if you saw my first implementation of Semaphore you would refuse to speak to me ever again

Definitely not :) Rust idioms do take time to master, same as C++ idioms. Besides, as I've taught a couple of university courses, I have high tolerance for creative code :-)

PR #87, opened 2020-10-03
Optimize Semaphore::try_wake_one

matklad:
### What does this PR do?

Avoid double lookup in a hash-map by utilizing the [entry API](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry).

### Motivation

Duplicate hashmap lookups are just a smell

### Related issues

Builds on #86 

### Checklist

### DIFF
--- a/scipio/src/semaphore.rs
+++ b/scipio/src/semaphore.rs
@@ -260,13 +260,10 @@ impl Semaphore {
     /// });
     /// ```
     pub fn signal(&self, units: u64) {
-        self.state.borrow_mut().signal(units);
-        loop {
-            if let Some(waiter) = self.state.borrow_mut().try_wake_one() {
-                waiter.wake();
-            } else {
-                return;
-            }
+        let mut state = self.state.borrow_mut();
+        state.signal(units);
+        while let Some(waiter) = state.try_wake_one() {
+            waiter.wake();
         }
     }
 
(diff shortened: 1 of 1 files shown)


glommer:
The funny thing about this is that I was just telling @HippoBaro the other day that it was unfortunate that Rust didn't support an API that would avoid the double lookup like C++ had.

Turns out I just wasn't aware of it (Maybe it is just me, but `entry` isn't the most obvious name for it...)

Good change then, because that is something that I was complaining about anyway.

glommer:
There was a conflict with your previous PR to unify usage of `Future`.

It seemed trivial so I resolved it and as soon as CI comes back I'll merge this.

glommer (approved):

matklad:
>It seemed trivial so I resolved it and as soon as CI comes back I'll merge this.

Btw, can I advertise using https://bors.tech/ which implements [not rocket science](https://graydon2.dreamwidth.org/1597.html) rule? It helps in particular in such situations, where you can just r+ and let the robot wait for CI. 

>Turns out I just wasn't aware of it (Maybe it is just me, but entry isn't the most obvious name for it...)

Yeah, this is one of the things you have to just know. There's a [crazier version](https://docs.rs/hashbrown/0.9.1/hashbrown/struct.HashMap.html#method.raw_entry_mut) in third party library.

PR #88, opened 2020-10-03
Fix bug in sync_open

matklad:
### What does this PR do?

Fix a bug with std::path::Path handling, Rust paths don't include trailing zero by default. 

### Checklist

### DIFF
--- a/scipio/src/sys/mod.rs
+++ b/scipio/src/sys/mod.rs
@@ -59,16 +59,16 @@ pub(crate) fn fs_hint_extentsize(fd: RawFd, size: usize) -> nix::Result<i32> {
 }
 
 pub(crate) fn remove_file(path: &Path) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


glommer (approved):
Small venting: I hate that github does not allow comments on the commit message. I want to comment on something that you wrote on the commit message, the code is fine!!! (so doing it here:)

I don't think stack allocation makes sense here. Keep in mind that scipio is meant to be an asynchronous thread per core framework.

io_uring doesn't support mkdir and my preference would be to do it in the kernel at some point but if that's not possible, we'd have to have a syscall thread like seastar does. One way or another, I view the sync functions as ultimately temporary.

So sure, we could use a stack allocation for *them*, but we wouldn't be able to use a stack allocation for the asynchronous versions...

matklad:
Ah, indeed, this is a good point!

PR #89, opened 2020-10-03
Avoid double lookups in hash map

matklad:
### What does this PR do?

Code cleanup and avoiding premature pessimization

### Checklist

### DIFF
--- a/scipio/src/sys/sysfs.rs
+++ b/scipio/src/sys/sysfs.rs
@@ -48,11 +48,8 @@ macro_rules! block_property {
         DEV_MAP.with(|x| {
             let key = ($major, $minor);
             let mut map = x.borrow_mut();
-            if !map.contains_key(&key) {
-                map.insert(key, BlockDevice::new(key));
-            }
+            let bdev = map.entry(key).or_insert_with(|| BlockDevice::new(key));
 
-            let bdev = map.get(&key).unwrap();
             bdev.$property.clone()
         });
     };
--- a/scipio/src/sys/sysfs.rs
+++ b/scipio/src/sys/sysfs.rs
@@ -55,9 +55,11 @@ macro_rules! block_property {
     };
 }
 
-fn read_int<P: AsRef<Path> + std::fmt::Debug>(path: P) -> isize {
-    let data = read_to_string(&path).unwrap_or_else(|_| panic!("reading {:?}", path));
-    let contents = &data.trim_matches('\n');
+fn read_int<P: AsRef<Path>>(path: P) -> isize {
+    let path = path.as_ref();
+    let data =
+        read_to_string(path).unwrap_or_else(|err| panic!("reading {} ({})", path.display(), err));
+    let contents = data.trim_matches('\n');
     contents.parse::<isize>().unwrap()
 }
 


glommer:
@HippoBaro note that there are likely double-lookups everywhere since I didn't know about `entry`. To avoid putting all the burden on Aleksey let's see if we can carve out some time this week to look for some more.

scipio/src/sys/sysfs.rs
```diff
@@ -55,9 +55,11 @@ macro_rules! block_property {
     };
 }
 
-fn read_int<P: AsRef<Path> + std::fmt::Debug>(path: P) -> isize {
-    let data = read_to_string(&path).unwrap_or_else(|_| panic!("reading {:?}", path));
-    let contents = &data.trim_matches('\n');
+fn read_int<P: AsRef<Path>>(path: P) -> isize {
+    let path = path.as_ref();
+    let data =
+        read_to_string(path).unwrap_or_else(|err| panic!("reading {} ({})", path.display(), err));
```
glommer, line 61: I am not opposed to this cleanup, but can you explain why this is preferred to the old style ?

Overall my philoshopy on this is to leave all debate about style to machines. If cargo clippy and rustfmt didn't complain then having humans debate this is time better spent somewhere else.

However since you're clearly a lot rustier than we are I am wondering if there is something I missed here

glommer:
I am all in favor of the `entry` usage, which was probably the nicest found in a while for me, but left a comment on your cleanup.

Thanks!

scipio/src/sys/sysfs.rs
matklad: Preface: in terms of actual impact (magnitude) this is completely inconsequential and both versions are fine, really. In terms of style (sign), I feel pretty strongly that the new version is better. Ie, I am very confident that I like my version more, but it also doesn't matter at all.

So the biggest change here is in the signature of the function -- `P` is no longer required to be a `Debug`. This just "makes sense" to me, as if you already can get a `Path` via `AsRef`, you can just as well print it, and there's no need to ask for more. 

The second change is a change in the behavior: `path.display()` will produce `reading /etc/hosts`, while `{:?}` would give you `reading "/etc/hosts"` (with quotes). For user-facing messages, I think `Display` is usually a nicer choice. Besides, I've noticed that you didn't use `path.display` elsewhere, so it was a good opportunity to drop a useful Rust tidbit :0)

The third reason is a really tiny future-proofing. In a super production-polished code (like standard library), this would have been written as

```rust
pub fn read_int(path: impl AsRef<Path>) {
    fn inner(path: &Path) {
        ... actual impl
    }

    inner(path.as_ref())
}
```

This version would be faster to compile, as only the tiny wrapper would have to be monomorphised, and not the whole function. This is similar in spirit to [non-templated base pattern](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-nondependent) from C++.

glommer: Thanks for the explanation. The changes to the function signature to be less restrictive definitely makes sense to me.

Merging.

glommer (approved):

PR #91, opened 2020-10-04
Fix unsound aliasing in PosixDmaBuf

matklad:
### What does this PR do?

Require exclusive access when writing to a PosixDmaBuf

Fixes
https://github.com/DataDog/scipio/issues/79#issuecomment-702934315

Note that this also removes `is_nonoverlapping` assertion -- it is now
lifted into the type system, `&[u8]` and `&mut [u8]` can't overlap.

I also have a hunch that we don't actually *need* these functions at
all, as they can be implemented by calling `.as_bytes[_mut]` and
https://doc.rust-lang.org/stable/std/primitive.slice.html#method.copy_from_slice,
but that can be left to the follow-up work.

### DIFF
--- a/scipio/src/sys/posix_buffers.rs
+++ b/scipio/src/sys/posix_buffers.rs
@@ -20,20 +20,6 @@ pub struct PosixDmaBuffer {
     size: usize,
 }
 
-// Adapted from stdlib's source code, simplified because we only ever need u8.
-fn is_nonoverlapping(src: *const u8, dst: *const u8, size: usize) -> bool {
-    let src_usize = src as usize;
-    let dst_usize = dst as usize;
-    let diff = if src_usize > dst_usize {
-        src_usize - dst_usize
-    } else {
-        dst_usize - src_usize
-    };
-    // If the absolute distance between the ptrs is at least as big as the size of the buffer,
-    // they do not overlap.
-    diff >= size
-}
-
 impl PosixDmaBuffer {
     pub(crate) fn new(size: usize) -> Option<PosixDmaBuffer> {
         let data = aligned_alloc(size, 4 << 10) as *mut u8;
@@ -68,7 +54,7 @@ impl PosixDmaBuffer {
         self.size
     }
 
-    pub fn as_mut_ptr(&self) -> *mut u8 {
+    pub fn as_mut_ptr(&mut self) -> *mut u8 {
         unsafe { self.data.as_ptr().add(self.trim) }
     }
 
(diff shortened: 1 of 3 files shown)


matklad:
While fixing this I've noticed *yet another* soundness issue with DmaBuf -- it exposes uninitialized memory, which is UB. That might be a hard nut to crack, see https://github.com/rust-lang/rfcs/pull/2930 for the discussion on how to handle this in stdlib.

glommer:
Agree about not moving the checks to the type system if possible.

About those functions being needed, their existence inside the buffer API mostly serves to hide the buffer size manipulation. 

However now that I think more about it, if we require - as you suggest - that one grabs a mutable reference the buffer to copy data into it, then one should not be able to trim the buffer or add offsets. That means that if you had previously acquired a reference in the form of `.as_bytes` it should still be valid.

I honestly see no point in merging this if we agree that we'd be better off with a simpler API that moves that out of the `PosixBuffer` responsibility. I'd prefer that you add a patch on top of this removing those methods altogether. There are some users in the `StreamFile` that would have to be adjusted.

scipio/src/sys/posix_buffers.rs
```diff
@@ -82,25 +68,23 @@ impl PosixDmaBuffer {
...
 
-        let src_ptr = src.as_ptr() as *const u8;
-        assert_eq!(is_nonoverlapping(src_ptr, self.as_ptr(), len), true);
+        let src_ptr = src.as_ptr();
         unsafe { std::ptr::copy_nonoverlapping(src_ptr, self.as_mut_ptr().add(offset), len) }
         len
     }
 
-    pub fn memset(&self, value: u8) {
+    pub fn memset(&mut self, value: u8) {
```
glommer, line 87: @matklad this is pre-existing of course, but should have size asserts here, no ?

I am inclined towards merging this now even if we can simplify later because it provides a very clear benefit.

glommer (approved):

scipio/src/sys/posix_buffers.rs
glommer: Humm, I guess this one can live without the assert because the size is not passed by the user.

matklad: Exactly!

matklad: As for the simplification, I suspect we can do two things:

* just use global allocator instead of megalign: modern std api allow to specify alignment (no link: I am on mobile)
* reduce unsafe by leveraging [u8]::copy_from_slce

glommer: I sent a PR today to deal with your second suggestion. But look at it later, go enjoy your life ! Don't text and drink.

About the allocator, I am less hot on it: With seastar with had major issues with locks inside the allocator to the point we had to have our own.

Until I understand the real life issues of this better for rust I wasn't planning on doing any serious work around allocation.

PR #97, opened 2020-10-07
cleanup remove, rename

glommer:
As suggested by Aleksey:
- no need to to_owned() in rename anymore, as we don't keep a copy
  in the top-level version
- since there is no more name conflict, call them remove and rename
  and drop the _file suffix.

### DIFF
--- a/scipio/src/io/dma_file.rs
+++ b/scipio/src/io/dma_file.rs
@@ -409,14 +409,13 @@ impl DmaFile {
     ///
     /// Warning: synchronous operation, will block the reactor
     pub async fn rename<P: AsRef<Path>>(&mut self, new_path: P) -> io::Result<()> {
-        let new_path = new_path.as_ref().to_owned();
         let old_path = path_required!(self, "rename")?;
         enhanced_try!(
-            crate::io::rename_file(old_path, &new_path).await,
+            crate::io::rename(old_path, &new_path).await,
             "Renaming",
             self
         )?;
-        self.path = Some(new_path);
+        self.path = Some(new_path.as_ref().to_owned());
         Ok(())
     }
 
--- a/scipio/src/io/mod.rs
+++ b/scipio/src/io/mod.rs
@@ -86,17 +86,14 @@ use std::path::Path;
 /// rename an existing file.
 ///
 /// Warning: synchronous operation, will block the reactor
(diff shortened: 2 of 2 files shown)


scipio/src/io/mod.rs
```diff
@@ -86,17 +86,14 @@ use std::path::Path;
 /// rename an existing file.
 ///
 /// Warning: synchronous operation, will block the reactor
-pub async fn rename_file<P: AsRef<Path>>(old_path: P, new_path: P) -> io::Result<()> {
-    let new_path = new_path.as_ref().to_owned();
-    let old_path = old_path.as_ref().to_owned();
-
-    sys::rename_file(&old_path, &new_path)
+pub async fn rename<P: AsRef<Path>>(old_path: P, new_path: P) -> io::Result<()> {
```
matklad, line 89: For full generality, this should have two type parameters: https://doc.rust-lang.org/stable/std/fs/fn.rename.html

Otherwise, LGTM.

scipio/src/io/dma_file.rs
```diff
@@ -412,7 +412,7 @@ impl DmaFile {
         let new_path = new_path.as_ref().to_owned();
         let old_path = path_required!(self, "rename")?;
         enhanced_try!(
-            crate::io::rename_file(old_path, &new_path).await,
```
matklad, line 415: Hm, there’s an extra to_owned above still.

glommer: This one is correct in my view.
Note that right below this line we're storing new_path as a copy.

We could not do `.to_owned()` in the beginning and do it closer to the place where we store a copy, but that becomes a style issue.

scipio/src/io/mod.rs
glommer: I find this a bit weird but I am all for conformity.

Thanks for bringing that to my attention

scipio/src/io/dma_file.rs
glommer: well, this is really minor but if we move the `to_owned` closer to its user we save an allocation in the failure case.
so let's do that.

scipio/src/io/mod.rs
matklad: With one type parameter, two types must match exactly, for expelled, &str and &str or &Path and &Path. With two type parameters, you can mix different types, like &str and &Path.

So this is rather API flexibility than conformity.

PR #98, opened 2020-10-07
remove unnecessary use of unsafe in buffers

glommer:
After Aleksey's good work, there are still two usages of unsafe inside
DmaBuffer that could be avoided.

Aleksey suggests we could probably just use the standard function
copy_from_slice and do away with our version.

However our version is very convenient in the sense that it doesn't
require the slices to be of the same size and does size management
for us.

What we can do, is replace the uses of unsafe with copy_from slice.

The problem is that now the name is confusing because we have two
things with the same name that behave differently. So rename them
read_at and write_at. The _at suffix makes it clear that we are
dealing with an offset.

### DIFF
--- a/scipio/src/io/file_stream.rs
+++ b/scipio/src/io/file_stream.rs
@@ -212,7 +212,7 @@ impl StreamReaderState {
                 panic!("Buffer not found. But we should only call this function after we verified that all buffers exist");
             }
             Some(buffer) => {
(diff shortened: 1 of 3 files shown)


glommer:
@matklad let me know what you think . This should settle the buffer issue.

matklad (commented):
LGTM!

PR #101, opened 2020-10-09
Channel

glommer:
Depends on https://github.com/DataDog/scipio/pull/100.

Will merge that first.

### DIFF
--- a/scipio/src/channels/local_channel.rs
+++ b/scipio/src/channels/local_channel.rs
@@ -1,0 +1,564 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
+// MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
+//
+use crate::channels::ChannelCapacity;
+use futures_lite::stream::Stream;
+use std::cell::RefCell;
+use std::collections::VecDeque;
+use std::future::Future;
+use std::io;
+use std::pin::Pin;
+use std::rc::Rc;
+use std::task::{Context, Poll, Waker};
+
+#[derive(Debug)]
+/// This is the error that is returned from channel operations if something goes wrong.
+///
+/// It encapsulates an [`ErrorKind`], which will be different for each kind of error and
+/// also makes the item sent through the channel available through the public attribute `item`.
+///
+/// You can convert between this and [`io::Error`] (losing information on `item`) so the
+/// ? constructs around [`io::Error`] should all work. Another possible way of doing this
+/// is returning an [`io::Error`] and encapsulating `item` on its inner field. However that
+/// adds [`Send`] and [`Sync`] requirements plus an allocation to the inner field (or to a
+/// helper struct around the inner field) that we'd like to avoid.
+///
+/// [`ErrorKind`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html
+/// [`io::Error`]: https://doc.rust-lang.org/std/io/struct.Error.html
+/// [`Send`]: https://doc.rust-lang.org/std/marker/trait.Send.html
+/// [`Sync`]: https://doc.rust-lang.org/std/marker/trait.Sync.html
+pub struct ChannelError<T> {
+    kind: io::ErrorKind,
+    /// The `ChannelError` encapsulates the item we originally tried to send into the channel
+    /// in case you need to do something with it upon failure.
+    pub item: T,
+}
+
+impl<T> ChannelError<T> {
+    fn new(kind: io::ErrorKind, item: T) -> ChannelError<T> {
+        ChannelError { kind, item }
+    }
+}
+
+impl<T: std::fmt::Debug> From<ChannelError<T>> for io::Error {
+    fn from(error: ChannelError<T>) -> Self {
+        io::Error::new(error.kind, format!("item: {:?}", error.item))
+    }
+}
+
+#[derive(Debug)]
+/// Send endpoint to the `local_channel`
+pub struct LocalSender<T> {
+    channel: LocalChannel<T>,
+}
+
+#[derive(Debug)]
+/// Receive endpoint to the `local_channel`
+///
+/// The `LocalReceiver` provides an interface compatible with [`StreamExt`] and will
+/// keep yielding elements until the sender is destroyed.
+///
+/// # Examples
+///
+/// ```
+/// use scipio::{LocalExecutor, Local};
+/// use scipio::channels::local_channel;
+/// use futures_lite::StreamExt;
+///
+/// let ex = LocalExecutor::make_default();
+/// ex.run(async move {
+///     let (sender, mut receiver) = local_channel::new_unbounded();
+///
+///     let h = Local::local(async move {
+///         let sum = receiver.fold(0, |acc, x| acc + x).await;
+///         assert_eq!(sum, 45);
+///     }).detach();
+///
+///     for i in 0..10 {
+///         sender.try_send(i);
+///     }
+///     drop(sender);
+///     h.await;
+/// });
+/// ```
+///
+/// [`StreamExt`]: https://docs.rs/futures/0.3.6/futures/stream/trait.StreamExt.html
+pub struct LocalReceiver<T> {
+    channel: LocalChannel<T>,
+}
+
+#[derive(Debug)]
+struct State<T> {
+    capacity: ChannelCapacity,
+    channel: VecDeque<T>,
+    recv_waiters: Option<VecDeque<Waker>>,
+    send_waiters: Option<VecDeque<Waker>>,
+}
+
+#[derive(Debug, Clone)]
+struct LocalChannel<T> {
+    state: Rc<RefCell<State<T>>>,
+}
+
+impl<T> LocalChannel<T> {
+    #[allow(clippy::new_ret_no_self)]
+    fn new(capacity: ChannelCapacity) -> (LocalSender<T>, LocalReceiver<T>) {
+        let channel = match capacity {
+            ChannelCapacity::Unbounded => VecDeque::new(),
+            ChannelCapacity::Bounded(x) => VecDeque::with_capacity(x),
+        };
+
+        let state = Rc::new(RefCell::new(State {
+            capacity,
+            channel,
+            send_waiters: Some(VecDeque::new()),
+            recv_waiters: Some(VecDeque::new()),
+        }));
+
+        (
+            LocalSender {
+                channel: LocalChannel {
+                    state: state.clone(),
+                },
+            },
+            LocalReceiver {
+                channel: LocalChannel { state },
+            },
+        )
+    }
+
+    fn push(&self, item: T) -> Result<(), ChannelError<T>> {
+        let mut state = self.state.borrow_mut();
+        if state.recv_waiters.is_none() {
+            return Err(ChannelError::new(io::ErrorKind::BrokenPipe, item));
+        }
+
+        if let ChannelCapacity::Bounded(x) = state.capacity {
+            if state.channel.len() >= x {
+                return Err(ChannelError::new(io::ErrorKind::WouldBlock, item));
+            }
+        }
+
+        state.channel.push_back(item);
+        if let Some(w) = state.recv_waiters.as_mut().and_then(|x| x.pop_front()) {
+            drop(state);
+            w.wake();
+        }
+        Ok(())
+    }
+
+    fn is_full(&self) -> bool {
+        let state = self.state.borrow();
+        match state.capacity {
+            ChannelCapacity::Unbounded => false,
+            ChannelCapacity::Bounded(x) => state.channel.len() >= x,
+        }
+    }
+}
+
+/// Creates a new `local_channel` with unbounded capacity
+///
+/// # Examples
+/// ```
+/// use scipio::{LocalExecutor, Local};
+/// use scipio::channels::local_channel;
+/// use futures_lite::StreamExt;
+///
+/// let ex = LocalExecutor::make_default();
+/// ex.run(async move {
+///     let (sender, mut receiver) = local_channel::new_unbounded();
+///     let h = Local::local(async move {
+///         assert_eq!(receiver.next().await.unwrap(), 0);
+///     }).detach();
+///     sender.try_send(0);
+///     drop(sender);
+///     h.await;
+/// });
+/// ```
+pub fn new_unbounded<T>() -> (LocalSender<T>, LocalReceiver<T>) {
+    LocalChannel::new(ChannelCapacity::Unbounded)
+}
+
+/// Creates a new `local_channel` with capacity limited to the `size` argument
+///
+/// # Examples
+/// ```
+/// use scipio::{LocalExecutor, Local};
+/// use scipio::channels::local_channel;
+/// use futures_lite::StreamExt;
+///
+/// let ex = LocalExecutor::make_default();
+/// ex.run(async move {
+///     let (sender, mut receiver) = local_channel::new_bounded(1);
+///     assert_eq!(sender.is_full(), false);
+///     sender.try_send(0);
+///     assert_eq!(sender.is_full(), true);
+///     receiver.next().await.unwrap();
+///     assert_eq!(sender.is_full(), false);
+/// });
+/// ```
+pub fn new_bounded<T>(size: usize) -> (LocalSender<T>, LocalReceiver<T>) {
+    LocalChannel::new(ChannelCapacity::Bounded(size))
+}
+
+struct SendWaiter<T> {
+    channel: LocalChannel<T>,
+}
+
+impl<T> Future for SendWaiter<T> {
+    type Output = ();
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        if !self.channel.is_full() {
+            Poll::Ready(())
+        } else {
+            let mut state = self.channel.state.borrow_mut();
+            state
+                .send_waiters
+                .as_mut()
+                .unwrap()
+                .push_back(cx.waker().clone());
+            Poll::Pending
+        }
+    }
+}
+
+#[allow(clippy::len_without_is_empty)]
+impl<T> LocalSender<T> {
+    /// Sends data into this channel.
+    ///
+    /// It returns a [`ChannelError`] encapsulating a [`BrokenPipe`] if the receiver is destroyed.
+    /// It returns a [`ChannelError`] encapsulating a [`WouldBlock`] if this is a bounded channel that has no more capacity
+    ///
+    /// # Examples
+    /// ```
+    /// use scipio::{LocalExecutor, Local};
+    /// use scipio::channels::local_channel;
+    /// use futures_lite::StreamExt;
+    ///
+    /// let ex = LocalExecutor::make_default();
+    /// ex.run(async move {
+    ///     let (sender, mut receiver) = local_channel::new_bounded(1);
+    ///     sender.try_send(0);
+    ///     sender.try_send(0).unwrap_err(); // no more capacity
+    ///     receiver.next().await.unwrap(); // now we have capacity again
+    ///     drop(receiver); // but because the receiver is destroyed send will err
+    ///     sender.try_send(0).unwrap_err();
+    /// });
+    /// ```
+    ///
+    /// [`BrokenPipe`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.BrokenPipe
+    /// [`WouldBlock`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.WouldBlock
+    /// [`Other`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.Other
+    /// [`io::Error`]: https://doc.rust-lang.org/std/io/struct.Error.html
+    pub fn try_send(&self, item: T) -> Result<(), ChannelError<T>> {
+        self.channel.push(item)
+    }
+
+    /// Sends data into this channel when it is ready to receive it
+    ///
+    /// For an unbounded channel this is just a more expensive version of [`send`]. Prefer
+    /// to use [`send`] instead.
+    ///
+    /// For a bounded channel this will push to the channel when the channel is ready to
+    /// receive data.
+    ///
+    /// # Examples
+    /// ```
+    /// use scipio::{LocalExecutor, Local};
+    /// use scipio::channels::local_channel;
+    ///
+    /// let ex = LocalExecutor::make_default();
+    /// ex.run(async move {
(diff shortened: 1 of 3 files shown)


scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,586 @@
...
+        if let ChannelCapacity::Bounded(x) = state.capacity {
+            if state.channel.len() >= x {
+                return Err(io::Error::new(io::ErrorKind::Other, "Channel Full"));
+            }
+        }
+
+        state.channel.push_back(item);
+        if !state.recv_waiters.is_empty() {
+            let waiters = std::mem::replace(&mut state.recv_waiters, Vec::new());
+            drop(state);
```
matklad, line 174: Why do we need to drop state early here? I *think* that we can’t re-enter this function due to .wake()s, am I wrong?

If we don’t need to release “lock” early, then we can use drain to re-use allocation for the list of walkers.

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,586 @@
...
+    /// let ex = LocalExecutor::make_default();
+    /// ex.run(async move {
+    ///     let (sender, receiver) = LocalChannel::new_bounded(1);
+    ///     sender.send_eventually(0).await.unwrap();
+    ///     drop(receiver);
+    /// });
+    /// ```
+    ///
+    /// [`send`]: struct.LocalSender.html#method.send
+    pub async fn send_eventually(&self, item: T) -> io::Result<()> {
```
matklad, line 310: Idiomatic names here are send for blocking send and try_send for non-blocking send. I think the two best crates to study channel’s APIs are flume and crossbeam channel.

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,586 @@
...
+    ///     receiver.next().await.unwrap(); // now we have capacity again
+    ///     drop(receiver); // but because the receiver is destroyed send will err
+    ///     sender.send(0).unwrap_err();
+    /// });
+    /// ```
+    ///
+    /// [`BrokenPipe`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.BrokenPipe
+    /// [`Other`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.Other
+    /// [`io::Error`]: https://doc.rust-lang.org/std/io/struct.Error.html
+    pub fn send(&self, item: T) -> io::Result<()> {
```
matklad, line 284: Using io::Result here is problematic, because it won’t be possible to recover the item on error. Idiomatic Rust solution here is to return a type similar to Result<(), T>.

scipio/src/channels/local_channel.rs
matklad: Additional point of interest: there was a long discussion about the result of send. There are a regiments for the position that send should never fail, and that sending over a closed channel should be an error: https://github.com/crossbeam-rs/crossbeam/issues/314

scipio/src/channels/local_channel.rs
glommer: I understand `wake` to be reentrant. It may call the future's poll function immediately, which in turn may try to add to the waiters array again if it is not ready.

Is that not the case ?

scipio/src/channels/local_channel.rs
glommer: Thanks. In the case of an unbounded channel `try_send` would always succeed but I guess that's okay.

scipio/src/channels/local_channel.rs
glommer: right, that makes sense.

glommer:
In my view one of the nastiest classes of bugs in asynchronous systems is
missed notifications.

Take a condition variable as an example:
Let's say fiber A .waits() on it, and then fiber B notify() it. In 99.9% of
the cases that happen in this order and you are happy.
Then due to one small timing issue the notify() happens before the wait(),
you miss the notification, and good luck figuring out why.
(This is part of the reason I don't offer a condvar in Scipio, btw. I
essentially haven't figured out a way to make it immune to that without
essentially making it into a gate)

For some usages of a channel, a silent success over a closed channel would
eat up a notification to a similar effect.
Say you have a fiber that is generating data, and another one that is
responsible for writing it to disk. You send a request for flush/fsync over
the channel,
and that gets silently dropped because the receiver is gone. Granted, it's
not trivial to decide what to do, which is why I picked this example. But
it's certainly
better than not ever knowing this happens when you are trying to debug what
becomes essentially a rare data corruption on crash.

So given the above I do not agree that a send over a closed channel should
ever be silent.

To be clear, I do agree with the reasoning you made but a case like this is
a case of essentially like picking a politician for office.
Deep down you know they are all bad, you just have to choose which aligns
more with the set of issues that you care about.

matklad:
>To be clear, I do agree with the reasoning you made but a case like this is
a case of essentially like picking a politician for office.
Deep down you know they are all bad, you just have to choose which aligns
more with the set of issues that you care about.

+100500, I don't really think it's productive to debate this, I just wanted to give some extra context here. On thing I want to clarify though, is that the alternative proposal is *panicking* on send if the receiver is gone. I wouldn't call this silent; on the contrary, it elevates recoverable `Result` to non-recoverable, "this is a bug" panic.

>This is part of the reason I don't offer a condvar in Scipio, btw. I
essentially haven't figured out a way to make it immune to that without
essentially making it into a gate

Not sure if this fully covers all the footguns, but I quite like how std does notification's API. Condvar's [wait](https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html#method.wait) consumes a mutex guard, which prevents some logical races. 

Thread [parking](https://doc.rust-lang.org/stable/std/thread/fn.park.html) API just makes `unpark(); park()` sequence work in any relative order, making a sort-of level-triggered API.

scipio/src/channels/local_channel.rs
matklad: Good question, I actually don't know the answer, and quickly looking through docs didn't reveal one. I think ultimately that depends on the executor (but I don't know how scipo's one work), and I would naively expect no reentrancy -- otherwise, it seems like it is possible to get a stack overflow, if the executor runs a bunch of tasks in the single stack.

I guess the current impl is the conservative one, and we can always dodge the allocation problem by using a SmallVec.

matklad: Asked on urlo: https://users.rust-lang.org/t/reentrancy-guarantees-of-std-wake/49977

glommer: aha! so many crates. I didn't know about smallvec but it is *Exactly* what I wanted both here and in other places (like inside the reactor to collect the vector of wakers)

glommer: I see that the outcome of that discussion is that it is a bad idea to wake with the mutex (or refcell) taken so this seems like I was in the right track here.

glommer:
Hey, not all discussions have to be productive! For some we can settle on *fun* =)

Panic vs requiring `unwrap` to me falls into the idea of making Scipio less opinionated whenever possible. I just gave an example of usage where you'd prefer to panic but it is equally easy to imagine others where missing the notification is okay.

glommer:
I just added a second commit implementing most suggestions from @matklad to make review easier. I will stash it into a single commit once we're all in agreement

In particular:
- the original item is now returned on Error. I still am using the `io::Error` so signal reason for error by pairing both into a helper Error struct.
- method names are changed and now we have `try_send` and `send`

As we have discussed:
- we'll keep the need to `.unwrap()` to generate a panic.
- it is best indeed to release the `RefCell` before waking wakers.

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,622 @@
...
+
+        if let ChannelCapacity::Bounded(x) = state.capacity {
+            if state.channel.len() >= x {
+                return Err(ChannelError::new(io::ErrorKind::WouldBlock, item));
+            }
+        }
+
+        state.channel.push_back(item);
+        if !state.recv_waiters.is_empty() {
+            let waiters = std::mem::replace(&mut state.recv_waiters, Vec::new());
```
matklad, line 208: The shorter way to spell this would be 

```suggestion
            let waiters = std::mem::take(&mut state.recv_waiters);
```

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// let ex = LocalExecutor::make_default();
+    /// ex.run(async move {
+    ///     let (sender, receiver) = LocalChannel::new_bounded(1);
+    ///     assert_eq!(sender.has_room(), true);
+    ///     sender.try_send(0);
+    ///     assert_eq!(sender.has_room(), false);
+    ///     drop(receiver);
+    /// });
+    /// ```
+    pub fn has_room(&self) -> bool {
```
matklad, line 375: Alternative name would be is_full: https://docs.rs/flume/0.9.1/flume/struct.Sender.html#method.is_full

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,622 @@
...
+            ChannelCapacity::Bounded(x) => VecDeque::with_capacity(x),
+        };
+
+        let state = Rc::new(RefCell::new(State {
+            capacity,
+            channel,
+            send_waiters: Vec::new(),
+            recv_waiters: Vec::new(),
+            receiver_alive: true,
+            sender_alive: true,
```
matklad, line 179: I wonder if it makes sense to pair `waiters: Vec` and `alive: bool` into a single `waiters: Option<Vec>`?

That way, type system guarantees that you can‘t add waiter if the opposite side is closed.

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,622 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
+// MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
+//
+use crate::channels::ChannelCapacity;
+use futures::Stream;
```
matklad, line 7: I think here and elsewhere, we should stick to lite/core variants of futures, and remove dependency on futures crate eventually

https://docs.rs/futures-core/0.3.6/futures_core/stream/trait.Stream.html
https://docs.rs/futures-lite/1.11.1/futures_lite/stream/index.html

glommer: All good comments and I'll merge them all, thanks.

About this futures futures-lite, etc. I must admit that although I have liked your patch to do that change earlier and tried to follow suit, I am a bit confused with which is which.

Maybe it is because "futures" sounds so official =)
Both "futures" and "futures-lite" are listed as crates on Cargo and from what you wrote above lite seems to be one of the good ones?

I think the easiest way to solve this once and for all is to just remove the bad one from the list, fixing all the fallout -- as soon as I understand which it ought to be. I'll do that today so don't worry much about it.

matklad: Yeah, the situation here is confusing. My understanding of the lay of the land:

* `std::future` and `std::task` module provide the absolute bare bones future interface. They essentially just expose compiler hooks required to make the whole construction work. At the moment, they lack certain things which absolutely need to be in std, but are just not designed yet. Specifically, std lacks abstraction for streams and integration of futures and io. Stuff in std is interface types.
* `futures-core` -- these are missing types from std: `Stream`, `AsyncRead`, `AsyncWrite`. This is also an interface crate (it, it's types are your public API)
* `futures` crate re-exports `futures-core`, and provides various convenience methods and combinators. It's an implementation crate (ie, users of your crate are not aware if you are using futures or not, they are aware only of futures-core). `futures` is also a historical crate, the first version of futures predates async/await. As such, it went through a number of redesigns, and each redesign was compromised by the requirement of maintaining backwards compatibility. futures accumulated a lot of cruft.
* `futures-lite` plays exactly the same role as `futures`, but it has a benefit of starting from clean slate, which, as claimed in the readme, are substantional: https://github.com/stjepang/futures-lite.

My takeaways here:

* we definitely should not use *both* `futures` and `futures-lite`
* which one we use is not overly important, as they are impl details (and both bottom out to std/future-core for interface types)
* `futures-lite` seems to be a more compelling alternative. Caveat -- I am not a heavy futures user, so I can't speak from direct experience. However, stjepang is one of the best library designers I know, and I am comfortable just trusting their opinion without much thought :-)

glommer: futures-lite it is.
A lot of scipio itself was inspired by his work so I am totally cool with that.

So we don't need to list futures-core as a crate right ?(right now it is not).
I'll just go ahead later today and delete futures from the cargo list.

matklad:
You'll be looking for alternative to `join_all` fom futures. Here it is:
https://github.com/stjepang/futures-lite/issues/2#issuecomment-696141698

Daniel-B-Smith:
My only drive-by comment is that it would be nice to have a `len()` method on the receiver accessible for stats. When monitoring a data pipeline in a process, I've used the channel `len()` method to monitor where our bottlenecks are coming from. 

Though, I would also be more than open to alternative monitoring strategies.

scipio/src/channels/local_channel.rs
glommer: I played with this a bit and there is one major showstopper

Correct me if I am missing something, but compare those two:
https://docs.rs/futures/0.3.6/futures/stream/trait.StreamExt.html#method.for_each
https://docs.rs/futures-lite/1.11.1/futures_lite/stream/trait.StreamExt.html#method.for_each

The futures-lite version of `for_each` cannot take an async closure? That means I essentially cannot chain futures in parallel.

It also seems to lack an implementation of `for_each_concurrent` and `try_for_each_concurrent` which I consider essential.

glommer: This is a bit sad. futures-lite claims to be compatible with futures, but I tried to standardize into one of them - and tried in both directions, and none of the efforts were trivial.

The use of futures-lite seem to be coming mostly from the code we inherited originally from async-io, which makes sense as they are all from the same author.

Most of the things that I have added, unaware of the distinction between them, seemed to ended up being `futures` which hints that maybe I have an unconscious preference towards it.

The lack of `for_each_concurrent` and the fact that `for_each` don't seem to take async functions itself seem like a bit of a deal breaker to me as I've been using those two things heavily (again, unless I am missing something).

Eager to hear your thoughts in the morning but from what I could gather so far the most likely direction would be for us to standardize on `futures`

A lot of the `futures-lite` code is in networking.rs which I'd like to rewrite to use io_uring native anyway so I wouldn't necessarily bother converting at the moment

matklad: I think `for_each` case is a genuine API cleanup, as original for_each breaks orthogonality. The `then` function can be used for async processing:

```rust
stream.for_each(|x| async { stuff })
==>
stream.then(|x| async { stuff }).for_each(|x| ())
stream.then(|x| async { stuff }).collect::<()>()
```

`for_each_concurrent` is indeed missing: https://github.com/stjepang/futures-lite/issues/26. I *think* this is understandable, as API is a bit weird -- the `limit` parameter looks like it is from another layer of the abstraction. It doesn't look like a primitive combinator. 

It also shouldn't be hard to roll-our-own `for_each_concurrent` as a free-standing function.

I slightly lean towards using futures-lite and rolling our own combinators which don't exist there yet, but this is a weak opinion.

glommer: fair comment on the usage of `then`.
It is a weird name because it feels like a `map`, and when I was reading futures-lite code using `then` I somehow actually understood that it would apply that to the end result of the stream, fold-like. Some things make a bit more sense now.

But correct me if I am wrong: let's say our channels now implement `futures_lite::Stream` instead of `futures::Stream`
Doesn't that mean that users of the channel now have to use `futures_lite::StreamExt` instead of `futures::StreamExt`?

It is one thing for us to implement our own version of the concurrent functions, I could be persuaded of that. Another is to force that into users.

I actually kind of like `for_each_concurrent` (and it's try counterpart). I am biased because it reminds me a lot of Seastar's `parallel_for_each` which was by far seastar's most useful primitive, except that it had limit control which is something we wanted to add for a long time.

In the vast majority of cases futures are only useful when you are spamming a bunch of them in parallel. Say for instance you are scanning a directory and reading file headers: you want to do it with something like `try_for_each_concurrent`, not `try_for_each` to generate enough parallelism. The limit helps in making sure you don't blow up your memory (although in all fairness, an excellent API would express that in terms of resources and not arbitrary units)

matklad: > But correct me if I am wrong: let's say our channels now implement futures_lite::Stream instead of futures::Stream
Doesn't that mean that users of the channel now have to use futures_lite::StreamExt instead of futures::StreamExt?

So here's the trick: both `futures_lite::Stream` and `futures::Stream` are re-exports of `futures_core::Stream`:

* https://github.com/rust-lang/futures-rs/blob/67566c65f5bba15dfad9d1d77e16d919bc18875f/futures/src/lib.rs#L441
* https://github.com/stjepang/futures-lite/blob/39affa66a7170ea425861cdab9043b3b2ed17d87/src/stream.rs#L22

They are exactly the same trait! The `Ext` versions are different between the crates, and each `Stream` implements both `Ext`s. So futures vs futures-lite is purely an implementation decsision, the public API is futures_core. 

>I actually kind of like for_each_concurrent (and it's try counterpart).

I do as well. What I am unsure about is whether they belong to the futures abstraction layer, or to the framework (scipio). For example, one problem with `for_each_concurrent` is that the limit is per invocation. So, if you have 10 tasks each of which spawn for_each_concurrent with limit of 10, the total limit would be 100. In scipio, we can, for example (by now means I am claiming that this is actually a good idea) provide global limit on the parallelism, a-la make jobserver, so that *all* concurrent `for_each_concurrent`s execute at most 10 things.

glommer: fun discussion!!
I had issues converting the channel code in the examples to futures-lite. It seemed that `for_each` required an implementation of Iterator, not Stream but it was late. I will go back to it soon with this newfound knowledge of `then` and see if I can make that work.

If the `Stream` type is the same, I am more open to using -lite everywhere since the application could keep using `for_each_concurrent`. If you grep, you will notice that scipio itself doesn't use it, which makes sense as it is mostly providing primitives. But the application we are building with scipio uses that everywhere...

Going back to the fun part I do think that a limit belongs in the invocation. You can always add a global limit on top of that.

Before I found out about `try_for_each` I was emulating that with Scipio's Semaphore. It wouldn't be a bad idea to come up with a higher level API that processes a stream in parallel and gets passed a `ResourceManager` kind of object. 

A `ResourceManager` can be built on top of a `Semaphore` or multi-tiered `Semaphores` and could be generic enough to support many such use cases.

For instance, let's say I can only allow my file spawning to support 100 parallel files (because say, that's my fd limit) but at the same time only at most 2MB of memory (because that's how much buffer space I allocated for this)

Every file being processed would bump the `ResourceManager` accordingly as they use resources.

glommer:
I have pushed a new version that incorporates both of your suggestions (@matklad and @Daniel-B-Smith) 

Thank you both.

However for tests I still can't use `futures_lite::StreamExt`. It simply won't compile no matter what I do.
Because the vast majority of `futures` use is in tests, I am considering as a temporary measure removing `futures` from our dependencies but leaving it in `dev-dependencies` so we are at least not forced into conversion hell right now.

But @matklad if you could advise what would be needed to get those tests working with futures-lite, that I would be quite enlightening

glommer:
I managed to convert most tests to futures_lite by using `fold` instead of `for_each`, which is better for this anyway.
The test using `for_each_concurrent` has to stay on `futures` but I guess that is fine, as we want to make sure that those keep working too as users may use it.

Now, the problem lies on `producer_early_drop_receiver`, which uses `take`. (currently commented out). I just can't get that to work and the error messages make no sense to me.

glommer:
... which must be some old bug! It works if I update `futures_lite`. Brilliant

glommer:
Everything is using futures-lite now with the exception of the one test in which we test that `for_each_concurrent` works.

For this to compile, though ,we  need PR https://github.com/DataDog/scipio/pull/115 merged first as it bumps the version of futures-lite (while removing futures from Cargo)

glommer:
pushed a new version that should fix all comments here, plus the ones raised in https://github.com/DataDog/scipio/pull/111 (which included a copy of this, and some people commented there)

* `new` variants now at the top level
*  and as a result `LocalChannel` is no longer public
* only wake up one waker upon push or pop to avoid quadratic wake storms

matklad (approved):
LGTM!

Couple of "what if" alternatives:

* I wonder if we can remove `deque` module after merging this? It feels a bit confusing to me: it is in collections, but is actually a synchronisation mechanism. I think people just provide channel interface instead of deques in rust typically. 

* for the code organization I personally have a mildly-strong preference to putting all `pub` things on top: https://github.com/rust-analyzer/rust-analyzer/blob/0c67edc0f7ebb7e6a265d4a925bc368e460cd8cb/docs/dev/style.md#order-of-items. I often just read source code as documentation (issuing "go to definiion" in the editor is faster than opening the browser), so I like when the prefix of file can be read as API docs.

glommer:
The Deque can go.

A bit of history, In our Scipio-based internal application I was using it to implement a write-behind / read-ahead mechanism similar to what we now have in the `StreamReader` and `StreamWriter` controlling in-flight buffers. When the code matured and I moved it to Scipio using the standard Future trait made more sense and we implemented poll. Therefore I didn't find much use for the `pop_front().await` pattern of the Deque.

It is still not the same as the channel, because technically you could implement urgency into the `Deque` by pushing to the front but we can always enhance the channel when it comes to it. I don't mind seeing it go.

Speaking of enhancing the channel, I am considering implementing a bidirectional channel too (should be easy with just two channels playing the role of both lanes). If you look at the controller code I am doing that manually which may mean there is room for such abstraction. Thoughts?

About the code organization: as I said before already while our community is small we can have those discussions but I do hope it grows! When it does, programmers tend to naturally bite each other too much about style and I always hated that. So I found it very refreshing that Rust has clippy and rustfmt which allowed me to essentially come with the the "if they don't complain, everything goes" policy. That means I will certainly not oppose it if you do it this way but I'd be wary of enforcing such a rule.

Now in my personal opinion, it does sound like a good rule. I am wondering if we couldn't get Clippy to shout about it ?

matklad:
> but I'd be wary of enforcing such a rule.

Agree, the only rule we should enforce is "CI is green". If some guideline can't be automatically checked, it's futile to enforce it. Though, having non-enforceable/non-enforced guidelines is still useful to steer new code and refactorings in the right direction.

>I am wondering if we couldn't get Clippy to shout about it ?

I don't think so. Well, in theory we can contribute a lint for that, but I wouldn't want to use this -- this is a pretty nuanced guideline, and I fear would have a fair amount of false positives if enforced by a dumb robot :)

glommer:
Hey, robots have feelings too.
Calling them dumb is how you get Skynet.

PR #103, opened 2020-10-11
Check formatting during `cargo test`

matklad:
This is easier for contributors as you don't have to wait for CI to
find out that you need to reformat the code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@
 version: 2
 
 jobs:
-  cargobuild:
+  cargotest:
     docker:
       - image: circleci/rust:1.46.0
     steps:
@@ -20,6 +20,10 @@ jobs:
       - run:
           name: Build all targets
           command: cargo build --all --all-targets
+      - run:
+          name: Run the tests
+          # CI machine has a pretty old kernel, so run only tidy tests
+          command: cargo test --test tidy
       - save_cache:
           paths:
             - /usr/local/cargo/registry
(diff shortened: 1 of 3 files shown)


matklad:
I don't have a super-strong option here, but this the pattern used in [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/blob/b7596d248382aa88b81da4dd6c7598993bc4b787/xtask/tests/tidy.rs), and it works out nicely. Generally, I prefer to put as little as possible into the CI scripts :)

glommer:
I agree, CI should be the last resort mechanism. Myself included, it has happened that sometimes one pushes code that we think is perfectly ok but then there is this one file that didn't get formatted, or this last minute `git add` that broke formatting, etc.

 However the CI run is still nice to have in case a contributor (even ourselves sometimes) forget to run `cargo test`

It is good practice to do it, but it is also good practice to run rustfmt and yet we forget.

So I welcome the code to add this to `cargo test`, but would like to keep the CI run.

matklad:
I mean, CI runs `cargo test` as well, no? So we get strictly more testing then

glommer:
we will at some point, and then I'd have no objections.
but right now circleCi doesn't support new enough kernels to run io_uring =(

matklad:
Ouch, didn't expect that. 

I've added a job to run only tidy tests to the CI

glommer:
yeah  that sucks. You might remember the commit I pointed out that you broke a while ago: that's exactly because we unfortunately can't run all tests yet so I'm relying on manually doing it. This will change with time, so it's not too bad, just annoying.

I will ask @HippoBaro to take a look at this since he is the only mainly handle all the CI work

glommer:
Closing in favor of #468

PR #104, opened 2020-10-11
Fix use after free during cancellation

matklad:
### What does this PR do?

Fix use after free due to overly eager `Source::drop` (see commit message for the details). 

### Additional Notes

@glommer I am much less confident in this PR than usual. It would be nice to be able to actually trigger this condition in tests, but this doesn't seem to easy -- we need two ~~parallel syscalls~~ concurrent io_uring requests, one of which finishes fast, and the other doesn't finish. I wonder if we should add some kind of "mock" / "strace" capability, so that it is easy to fake syscalls going via uring. Might be useful for fault injection as well. 

### Checklist

### DIFF
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -1010,11 +1010,14 @@ mod tests {
             (source, op)
         }
 
+        let (fast, op) = timeout_source(50);
+        queue_standard_request!(reactor, &fast, op);
+
         let (slow, op) = timeout_source(150);
         queue_standard_request!(reactor, &slow, op);
 
-        let (fast, op) = timeout_source(50);
-        queue_standard_request!(reactor, &fast, op);
+        let (lethargic, op) = timeout_source(300);
+        queue_standard_request!(reactor, &lethargic, op);
 
         let start = Instant::now();
         let mut wakers = Vec::new();
@@ -1022,9 +1025,11 @@ mod tests {
         let elapsed_ms = start.elapsed().as_millis();
         assert!(50 <= elapsed_ms && elapsed_ms < 100);
 
+        drop(slow); // Cancel this one.
+
         let mut wakers = Vec::new();
         reactor.wait(&mut wakers, None, None).unwrap();
         let elapsed_ms = start.elapsed().as_millis();
-        assert!(150 <= elapsed_ms && elapsed_ms < 200);
+        assert!(300 <= elapsed_ms && elapsed_ms < 350);
     }
 }
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -209,13 +209,18 @@ where
             return Some(());
         }
 
-        // Will be None if it was cancelled while in-flight.
-        if let Some(src) = consume_source(value.user_data()) {
+        let src = consume_source(value.user_data());
+
+        let result = value.result();
+        let was_cancelled =
+            matches!(&result, Err(err) if err.raw_os_error() == Some(libc::ECANCELED));
+
+        if !was_cancelled {
             let source = mut_source(&src);
 
             if try_process(source).is_none() {
                 let mut w = source.wakers.borrow_mut();
-                w.result = Some(value.result());
+                w.result = Some(result);
                 wakers.append(&mut w.waiters);
             }
         }
@@ -258,36 +263,19 @@ fn peek_source(id: u64) -> Rc<UnsafeCell<InnerSource>> {
     })
 }
 
-fn consume_source(id: u64) -> Option<Rc<UnsafeCell<InnerSource>>> {
+fn consume_source(id: u64) -> Rc<UnsafeCell<InnerSource>> {
     SOURCE_MAP.with(|x| {
         let mut map = x.borrow_mut();
-        if let Some(source) = map.map.remove(&id) {
-            let mut s = mut_source(&source);
-            s.id = None;
-            s.queue = None;
-            return Some(source);
-        }
-        None
+        let source = map.map.remove(&id).unwrap();
+        let mut s = mut_source(&source);
+        s.id = None;
+        s.queue = None;
+        source
     })
 }
 
 pub(crate) fn cancel_source(id: u64, queue: ReactorQueue) {
-    // This may happen during teardown of the system itself, in which
-    // case the thread local may be gone. That's fine because in that case
-    // the ring is gone
-    SOURCE_MAP
-        .try_with(|x| {
-            let mut map = x.borrow_mut();
-            // This is synchronous so if we were passed an id, it exists.
-            // Otherwise we panic. consume_source is different because it runs
-            // asynchronously when the ring returns. If we don't find and Id here
-            // that means that someone called drop() inside the io_uring poller which
-            // is totally illegal.
-            map.map.remove(&id).unwrap();
-            let mut q = queue.borrow_mut();
-            q.cancel_request(id);
-        })
-        .ok();
+    queue.borrow_mut().cancel_request(id);
 }
 
 #[derive(Debug)]
@@ -311,7 +299,14 @@ impl UringQueueState {
         match found {
             Some(idx) => {
                 self.submissions.remove(idx);
+                // We never submitted the request, so it's safe to consume
+                // source here -- kernel didn't see our buffers.
+                consume_source(id);
             }
+            // We are cancelling this request, but it is already submitted.
+            // This means that the kernel might be using the buffers right
+            // now, so we delay `consume_source` until we consume the
+            // corresponding event from the completion queue.
             None => self.cancellations.push_back(UringDescriptor {
                 args: UringOpDescriptor::Cancel(id),
                 fd: -1,
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -237,7 +237,7 @@ struct SourceMap {
 impl SourceMap {
     fn new() -> RefCell<Self> {
         RefCell::new(Self {
-            id: 0,
+            id: 1,
             map: HashMap::new(),
         })
     }


glommer:
I am all for error injection btw and that is one of the things I wanted to have eventually in scipio!
I wanted for instance to randomly fail some requests, especially with conditions like `ENOSPC` and etc. The problem with that is that it can't be *that* random: no real device will throw ENOSPC and then return correctly the very next request.

bad blocks, however, is quite random. So a good error injection mechanism takes time and effort to do well and we're not there yet. But.... I digress!!

With the caveat that I am responding without having read the code yet:
The beauty of thread per core is that parallel syscalls are not possible. At least not in the sense that would matter here as far as I understand your concern. a **syscall** is one call to `io_uring_enter`. It contains many requests.

Different executors are the only way to run things in different cores. And running things in different cores is the only way to  parallel syscalls, but they will be issued to different rings, as we have one ring per executor (Well, three rings - latency, poll, main: but they are serialized in an executor)

io_uring will issue all its `requests` inside a ring in parallel. That's a different matter. If we have any race it would come from this.

scipio/src/sys/uring.rs
```diff
@@ -215,15 +215,13 @@ where
...
-        // Will be None if it was cancelled while in-flight.
-        if let Some(src) = consume_source(value.user_data()) {
-            let source = mut_source(&src);
-
-            if try_process(source).is_none() {
-                let mut w = source.wakers.borrow_mut();
-                w.result = Some(value.result());
-                wakers.append(&mut w.waiters);
-            }
+        let src = consume_source(value.user_data());
```
glommer, line 218: In general I like the idea of pairing operations. It is easier to get it right and more robust.

However note that both those operations are not the same (we'll now discuss whether or not that matters, but...)

Prior to your patch we would only push a waker if the operation was *not* cancelled, because `try_process` was inside the `consume_source` branch.

Now we may be queueing a waker for an operation that is cancelled which may have other nasty side effects too.

Whether or not the operation was cancelled is encoded in the `result`, so we can still have it both ways: we can require consume, the way you do, but test for the `result` before calling `try_process`.

glommer:
Ok.

I think the code is going in a good direction and I agree that testing this is a PITA. When I was working on this code I found out a bunch of cases that were clearly wrong, but I could only tell that by carefully positioned `println` statements, which sucks.

Error injection helps but it is still not enough: let's say you use a source that is freed but not yet reallocated: that's wrong, but may not trigger any bugs. In fact, that's precisely what I saw and was referring to above.

The best way to deal with this in my view is either some kind of poisining, or better yet, ASAN.

I played a bit with ASAN when working with this, but didn't make too much progress for two reasons:
* it requires the nightly compiler. I would be opened to having things working on the nightly compiler without `requiring` it and it seems like asan could fall into this category, so that's not a deal breaker
* pretty much all tests triggered asan in some inexplicable way. And by all I mean, seriously, even `spawn_without_executor()` failed and there's clearly nothing wrong with that one.

The second point led me to believe that there is probably something that asan doesn't quite like with the implementation of `task`, which I unapologetically copied almost verbatim from `async-task` (except removing  the `Send` requirements), but that's still mostly a hunch. One way or another, ASAN is not very useful if it always fail =(

But let's say we do have ASAN working:

You don't need parallel syscalls to trigger what you describe you just need cancellations happening with operations still in flight. And there is a very easy way to trigger those, with timers.

Timers are coalesced but they are coalesced in the higher up layer, so we could still test them in the uring layer without coalescing. Without timer coalescing, you can fire a bunch of timers for various expiration times and selectively or randomly cancel some of them while you keep allocating more. If you are using your freelist implementation, and poison the freelist on free, you should be able to trigger bugs here.

scipio/src/sys/uring.rs
matklad: Hm, yeah, agree that this was an unintended change in the behavior, I've put the old behavior back in place.

However, it does seem surprising and lost-wakeupy to me. Intuitively, I would assume that, when we cancel an op, we need to wake everyone waiting on op and notify them that their waiting is now futile. But, even if that reasoning is true, it should be dealt with in a separate PR.

scipio/src/sys/uring.rs
```diff
@@ -317,12 +298,21 @@ impl UringQueueState {
...
+                // source here -- kernel didn't see our buffers.
+                consume_source(id);
+            }
+            None => {
+                self.cancellations.push_back(UringDescriptor {
+                    args: UringOpDescriptor::Cancel(id),
+                    fd: -1,
+                    user_data: 0,
+                });
+                // We are cancelling this request, but it is already submitted.
```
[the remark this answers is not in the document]
matklad: Heh, that was obvious when I was writing the code: the comment is in the place of `consume_source(id)` which isn't there :) Agree that this is confusing, moved comment above.

scipio/src/sys/uring.rs
glommer: You may be right and I think a great way of demonstrating that would be through testing those conditions specifically.
The reason this is done this way is essentially driven by the fact that cancellations right now are mostly coming from cancelled tasks in which case there is no one around to receive the wake.

Whether they will be ignored or leak in that situation is speculative for now. I am putting a lot of faith in my new friend miri to help us figure that out as we add more tests for those cases

matklad:
@glommer I still have trouble with exercising this code via tests.

Here are things I tried:

1. Using public API timers

```
    #[test]
    fn fast_slow_timer() {
        test_executor!(async move {
            let now = Instant::now();
            let fast = Timer::new(Duration::from_millis(200));
            let slow = Timer::new(Duration::from_millis(328));
            future::race(fast, slow).await;
            let elapsed_ms = now.elapsed().as_millis();
            assert!(200 <= elapsed_ms && elapsed_ms < 400);
        });
    }
```

This doesn't work because timers are coalesced.

2.   Manually creating reactor and submitting two timers:

```
    #[test]
    fn fast_slow_timer_ll() {
        let reactor = Reactor::new().unwrap();

        let slow_micros = 100_000;
        let fast_micros = 50_000;

        let slow = &Source::new(
            IoRequirements::default(),
            -1,
            SourceType::Timeout(Some(slow_micros)),
        );
        queue_standard_request!(reactor, slow, UringOpDescriptor::Timeout(slow_micros));

        let fast = &Source::new(
            IoRequirements::default(),
            -1,
            SourceType::Timeout(Some(fast_micros)),
        );
        queue_standard_request!(reactor, fast, UringOpDescriptor::Timeout(fast_micros));

        let now = Instant::now();
        future::block_on(future::race(slow.collect_rw(), fast.collect_rw())).unwrap();
        let elapsed_micros = now.elapsed().as_micros() as u64;

        assert!(slow_micros <= elapsed_micros && elapsed_micros < fast_micros);
    }
```

This one doesn't work because no one actually drives the reactor. Is there some existing test I can copy-paste setup code from?

glommer:
Unfortunately no. It's no accident that we don't really have tests for the io_uring module directly.

I can try to add some simple ones that you could at least use as inspiration. I am not 100% sure I'll be able to do it this week, though (setting expectations...) as I am fighting a deadline.

If you want to work with just some pointers for now, you are right that indeed the reactor needs to be driven manually.

It would probably be easier if you don't use futures.
At least some leaks should show this way, as we may try to reuse `Source`s.
You would not be calling `collect_rw` in that case, but rather just `Reactor::wait` directly and processing the output.

I think it will be quite the challenge to drive the reactor automatically at this level of abstraction

matklad:
Huh, I am observing something really weird. It looks like uring itself does some sort of time coalescing? I armed two timers, and they fired simultaneously:

```diff
diff --git a/scipio/src/parking.rs b/scipio/src/parking.rs
index 7e211f5..62ec64e 100644
--- a/scipio/src/parking.rs
+++ b/scipio/src/parking.rs
@@ -255,7 +255,7 @@ impl Reactor {
         }
     }
 
-    fn new_source(&self, raw: RawFd, stype: SourceType) -> Source {
+    pub(crate) fn new_source(&self, raw: RawFd, stype: SourceType) -> Source {
         let ioreq = self.current_io_requirements.borrow();
         sys::Source::new(*ioreq, raw, stype)
     }
diff --git a/scipio/src/sys/uring.rs b/scipio/src/sys/uring.rs
index cedd1e5..154f677 100644
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -210,6 +210,11 @@ where
     F: FnOnce(&mut InnerSource) -> Option<()>,
 {
     if let Some(value) = cqe {
+        eprintln!(
+            "process_one_event({:?}, {:?})",
+            value.user_data(),
+            value.result()
+        );
         // No user data is POLL_REMOVE or CANCEL, we won't process.
         if value.user_data() == 0 {
             return Some(());
@@ -219,7 +224,7 @@ where
 
         let result = value.result();
         let was_cancelled =
-            matches!(&result, Err(err) if err.raw_os_error() == Some(libc::ECANCELED));
+            false && matches!(&result, Err(err) if err.raw_os_error() == Some(libc::ECANCELED));
 
         if !was_cancelled {
             let source = mut_source(&src);
@@ -255,6 +260,12 @@ fn add_source(source: &Source, queue: ReactorQueue) -> u64 {
     SOURCE_MAP.with(|x| {
         let mut map = x.borrow_mut();
         let id = map.id;
+        eprintln!(
+            "add_source({:?}, {:?})",
+            id,
+            mut_source(&source.inner).source_type
+        );
+
         map.id += 1;
         map.map.insert(id, source.inner.clone());
         source.update_reactor_info(id, queue.clone());
@@ -332,7 +343,6 @@ trait UringCommon {
 
     fn add_to_submission_queue(&mut self, source: &Source, descriptor: UringOpDescriptor) {
         let id = add_source(source, self.submission_queue());
-
         let q = self.submission_queue();
         let mut queue = q.borrow_mut();
         queue.submissions.push_back(UringDescriptor {
@@ -377,6 +387,7 @@ trait UringCommon {
             }
             completed += 1;
         }
+        eprintln!("completed = {:?}", completed);
         completed
     }
 
@@ -754,13 +765,13 @@ macro_rules! flush_rings {
 }
 
 macro_rules! queue_request_into_ring {
-    ($ring:expr, $source:ident, $op:expr) => {{
+    ($ring:expr, $source:expr, $op:expr) => {{
         $ring.borrow_mut().add_to_submission_queue($source, $op)
     }};
 }
 
 macro_rules! queue_storage_io_request {
-    ($self:expr, $source:ident, $op:expr) => {{
+    ($self:expr, $source:expr, $op:expr) => {{
         let pollable = match $source.source_type() {
             SourceType::DmaRead(p, _) => p,
             SourceType::DmaWrite(p) => p,
@@ -774,7 +785,7 @@ macro_rules! queue_storage_io_request {
 }
 
 macro_rules! queue_standard_request {
-    ($self:expr, $source:ident, $op:expr) => {{
+    ($self:expr, $source:expr, $op:expr) => {{
         match $source.latency_req() {
             Latency::NotImportant => queue_request_into_ring!($self.main_ring, $source, $op),
             Latency::Matters(_) => queue_request_into_ring!($self.latency_ring, $source, $op),
@@ -985,3 +996,40 @@ impl Reactor {
 impl Drop for Reactor {
     fn drop(&mut self) {}
 }
+
+#[cfg(test)]
+mod tests {
+    use std::time::Instant;
+
+    use futures_lite::future;
+
+    use super::*;
+
+    #[test]
+    fn fast_slow_timer_ll() {
+        let reactor = Reactor::new().unwrap();
+
+        let slow_micros = 100_0000000;
+        let fast_micros = 50_000;
+
+        let slow = Source::new(
+            IoRequirements::default(),
+            -1,
+            SourceType::Timeout(Some(slow_micros)),
+        );
+        queue_standard_request!(reactor, &slow, UringOpDescriptor::Timeout(slow_micros));
+
+        let fast = Source::new(
+            IoRequirements::default(),
+            -1,
+            SourceType::Timeout(Some(fast_micros)),
+        );
+        queue_standard_request!(reactor, &fast, UringOpDescriptor::Timeout(fast_micros));
+
+        let now = Instant::now();
+        let mut wakers = Vec::new();
+        reactor.wait(&mut wakers, None, None).unwrap();
+        let elapsed_ms = now.elapsed().as_millis();
+        println!("done in {:?}", now.elapsed());
+    }
+}
```

printout

```
running 1 test
add_source(0, Timeout(Some(1000000000)))
add_source(1, Timeout(Some(50000)))
completed = 0
completed = 0
completed = 0
completed = 0
completed = 0
completed = 0
add_source(2, LinkRings(Linked))
completed = 0
completed = 0
process_one_event(0, Err(Os { code: 62, kind: Other, message: "Timer expired" }))
process_one_event(1, Err(Os { code: 62, kind: Other, message: "Timer expired" }))
completed = 2  # <- we reaped both completion in one ring turn, how can this be? 
done in 50.243032ms
test sys::uring::tests::fast_slow_timer_ll ... ok
```

matklad:
Lol, the reason I am seeing this weirdness is *another* use after free:

```
            UringOpDescriptor::Timeout(micros) => {
                let d = Duration::from_micros(micros);
                let timeout: _ = uring_sys::__kernel_timespec {
                    tv_sec: d.as_secs() as _,
                    tv_nsec: d.subsec_nanos() as _,
                };
                sqe.prep_timeout(&timeout);
            }
```

We are passing *address* of the timespec to the kernel. The timespeck lives on the stack. The actual memory of the timespec is thrown away after function return -- that's a local variable! And it so happens that, because we call this function twice, we overwrite the the first timeout with the second one! Isn't system programming wonderful? 

The following "fix" confirms that this is indeed the problem:

```
            UringOpDescriptor::Timeout(micros) => {
                let d = Duration::from_micros(micros);

                let timeout: _ = Box::new(uring_sys::__kernel_timespec {
                    tv_sec: d.as_secs() as _,
                    tv_nsec: d.subsec_nanos() as _,
                });
                sqe.prep_timeout(&*Box::leak(timeout));
            }
```

glommer:
yeah, this s** is hard...
Very good catch!

I do wonder if the better solution wouldn't be to add the timespec structure into the `Timeout` enum variant instead of boxing it at that point.

Please consider sending a fix for that separately

matklad:
Sumbitted #116

matklad:
Should be ready for review now!

I've extended smoke test to do the cancellation. I don't think it's possible to verify that we are not UAFing here, even with sanitizers -- it's the kernel which would use the freed memory, and we can't poison kernel from the user space. 

I've added some printfs though to convince myself that the thing works as I expect it to:

```diff
diff --git a/scipio/src/sys/uring.rs b/scipio/src/sys/uring.rs
index d1e453c..49960fa 100644
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -214,6 +214,11 @@ where
         let result = value.result();
         let was_cancelled =
             matches!(&result, Err(err) if err.raw_os_error() == Some(libc::ECANCELED));
+        eprintln!(
+            "was_cancelled({:?}) = {:?}",
+            value.user_data(),
+            was_cancelled
+        );
 
         if !was_cancelled {
             let source = mut_source(&src);
@@ -249,6 +254,11 @@ fn add_source(source: &Source, queue: ReactorQueue) -> u64 {
     SOURCE_MAP.with(|x| {
         let mut map = x.borrow_mut();
         let id = map.id;
+        eprintln!(
+            "add_source({:?}, {:?})",
+            id,
+            mut_source(&source.inner).source_type
+        );
         map.id += 1;
         map.map.insert(id, source.inner.clone());
         source.update_reactor_info(id, queue.clone());
@@ -267,6 +277,11 @@ fn consume_source(id: u64) -> Rc<UnsafeCell<InnerSource>> {
     SOURCE_MAP.with(|x| {
         let mut map = x.borrow_mut();
         let source = map.map.remove(&id).unwrap();
+        eprintln!(
+            "consume_source({:?}, {:?})",
+            id,
+            mut_source(&source).source_type
+        );
         let mut s = mut_source(&source);
         s.id = None;
         s.queue = None;
@@ -987,6 +1002,12 @@ impl Drop for Reactor {
     fn drop(&mut self) {}
 }
 
+impl Drop for InnerSource {
+    fn drop(&mut self) {
+        eprintln!("Dropping {:?}", self.id)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::time::Instant;
@@ -1019,14 +1040,17 @@ mod tests {
         let (lethargic, op) = timeout_source(300);
         queue_standard_request!(reactor, &lethargic, op);
 
+        eprintln!("will wait1");
         let start = Instant::now();
         let mut wakers = Vec::new();
         reactor.wait(&mut wakers, None, None).unwrap();
         let elapsed_ms = start.elapsed().as_millis();
         assert!(50 <= elapsed_ms && elapsed_ms < 100);
 
+        eprintln!("will drop");
         drop(slow); // Cancel this one.
 
+        eprintln!("will wait2");
         let mut wakers = Vec::new();
         reactor.wait(&mut wakers, None, None).unwrap();
         let elapsed_ms = start.elapsed().as_millis();
```

Naturally, it **didn't** work as expected, due to a minor bug, which I fixed in the last comment. 

Final printout for me is

```
unning 1 test
add_source(1, Timeout(None, 50ms))
add_source(2, Timeout(None, 150ms))
add_source(3, Timeout(None, 300ms))
will wait1
add_source(4, LinkRings(Linked))
consume_source(1, Timeout(None, 50ms))
was_cancelled(1) = false
will drop
will wait2
consume_source(2, Timeout(None, 150ms))
was_cancelled(2) = true
Dropping None
consume_source(3, Timeout(None, 300ms))
was_cancelled(3) = false
Dropping None
Dropping None
Dropping None
Dropping None
test sys::uring::tests::timeout_smoke_test ... ok
```

I am wondering, should we `log::debug!` all the things perhaps?

glommer:
This is clear bug an improvement so I'll review and merge soon.

I think it should be possible to trigger a userspace issue with this that we could catch with sanitizers if we keep allocating sources and freeing it incorrectly in a loop: at some point we ought to do something ASAN would catch as a result.

matklad:
Hm, yeah, you are correct, we can force the thing to blow up with a stress-test! Though, I'd rather not do it in this PR :)

glommer:
I'm on the fence about debug messages. You only see them with nocapture anyway, so in my experience they'll just be ignored
(I just patched a lack of close today in the file stream tests that was there from day 1: we print about it, but we never see it in the test output!)

When tests do fail it is helpful to have messages saying what went wrong, so I am certainly not opposed. Just doubtful about how useful they are.

matklad:
>I'm on the fence about debug messages.

Heh, interesting! In my experience, I've also find `log::debug!` to be something which you *in theory* should use to debug similar problem, but, in practice, log statements are both:

* never present at the exact point you need them
* present at some unrelated points with verbose outputs

so manually adding `eprintln`s works better than leveraging the `log`/`tracing` ecosystem.

glommer:
Agree on the stress test.

BTW totally unrelated to this PR but since we are here, I feel like this is almost ready to go into crates - big thanks to you.
I'm just debating with myself whether I should at least have a Buffered I/O file interface merged first and the controllers / channel or if we are good as is.

On the one hand it's good to have a first 0.1 version out there so people can consume and generate more feedback, on the other hand feedback will only come if it is minimally useful.

We'll have to copy the uring crates code temporarily, because I think it will be a while before all the fixes that I sent there make into their respective crates. All is merged, so at some point we can shave that too.

Happy to hear your thoughts (and happy to take the discussion somewhere else too so we don't keep hijacking this lol)

scipio/src/sys/uring.rs
```diff
@@ -237,7 +237,7 @@ struct SourceMap {
 impl SourceMap {
     fn new() -> RefCell<Self> {
         RefCell::new(Self {
-            id: 0,
+            id: 1,
```
glommer, line 240: good catch, no idea how this didn't cause issues before.

scipio/src/sys/uring.rs
```diff
@@ -210,12 +210,19 @@ where
         }
 
         let src = consume_source(value.user_data());
-        let source = mut_source(&src);
 
-        if try_process(source).is_none() {
-            let mut w = source.wakers.borrow_mut();
-            w.result = Some(value.result());
-            wakers.append(&mut w.waiters);
+        let result = value.result();
```
glommer, line 214: can you fold this into the previous patch so we can keep everything as bisectable as possible?

Aside from that all good to go, thanks again!

glommer (approved):

matklad:
Hm, so I would be very wary of signaling general availability until all `unsafe` blocks are audited to not actually causing memory unsafety via public safe API. Given my rate of finding UBs so far, I don't think we are quite ready yet :D 

I think making a release, while clearly communicating that scipio may UB, eat laundry, or summon nasal daemons, should be fine. For example, an `0.1.0-alpha1` version would work. OTOH, it's not that hard to just depend on the git repo...

But maybe the point is moot -- you can't release on crates.io with git dependencies, so we need to upstream all io_uring-sys changes first and get a new release. I have a hunch that that might take some time.

scipio/src/sys/uring.rs
matklad: Done!

glommer:
One of my biggest concerns is that while we don't release someone may take the name.
Plus there is at least one user (see issue 1) that has requested this on crates.

I'm totally fine calling it alpha in the version and communicating the status clearly.

glommer:
The uring bits would have to be copied into the repo. I have an experimental patch that does this, I just haven't pushed it yet.

matklad:
you can publish an 0.0.1 empt crate under the name

glommer:
merged =)

good point on the empty crate, but if we are going through the trouble, why not put something minimally useful there?

matklad:
:thumbsup:, as long as the minimally useful thing is clearly demarcated as
not production ready.

Putting a minimally useful thing, which is broken, and which is not
explicitly advertised as broken, can be a net negative :)

PR #106, opened 2020-10-12
React spin

### LINKED ISSUE #76: For spinning executors, have a version of `poll` that doesn't mess with timers.
For executors that are configured to `spin`, we just `continue` the loop if we decide to keep spinning meaning we will go into `poll_io` to gather I/O.

This code is **correct** but it could be *more efficient*:
 - `poll_io` takes a duration parameter and it will reset the preempt timer for that duration. That means that we will have to do that timer dance every time we call `poll_io`.

If we are just spinning we don't have to worry about the preempt timer. Plus, if the natural result of spinning is that we give up and go to sleep after the spinning period ends, we will want to sleep with the timer disabled anyway.

So, it would be very helpful if the `Reactor` provided a new companion version of `poll_io` and `park`, such that:
 - when we detect that we *Start* spinning, (IOW, there is a preempt timer registered), we disable it and while we are in spin mode we don't register it again.
- when we *stop* spinning (IOW, events are generated from the poll), then we register the preempt timer again.

Technically this could be doing just with logic inside `poll_io`, but it would probably be cleaner to have a specialized version.

sitano:
Closes #76 by replacing `park()` with `park(spin: Duration)`.

This patch separates `ReactorLock::react` code paths into 2 flows: stepping through the `uring` queue and sleeping:

- it adds support for spinning directly to the `Reactor::wait`.
- it separates `Reactor::wait` into `wait` and `step`
- it stores wakers vec in Parker::Inner.

Spinning occurs right before sleeping (parking) with canceled or rearmed timer for the next timer occurrence.

### DIFF
--- a/scipio/src/executor.rs
+++ b/scipio/src/executor.rs
@@ -842,10 +842,6 @@ impl LocalExecutor {
         let waker = waker_fn(|| {});
         let cx = &mut Context::from_waker(&waker);
 
-        let spin_before_park = self.spin_before_park().unwrap_or_default();
-        let spin = spin_before_park.as_nanos() > 0;
-        let mut spin_since: Option<Instant> = None;
-
         LOCAL_EX.set(self, || loop {
             if let Poll::Ready(t) = future.as_mut().poll(cx) {
                 break t;
@@ -856,24 +852,9 @@ impl LocalExecutor {
             // requests that are latency sensitive we want them out of the
             // ring ASAP (before we run the task queues). We will also use
             // the opportunity to install the timer.
-            let duration = self.preempt_timer_duration();
-            self.parker.poll_io(duration);
+            self.parker.poll_io(self.preempt_timer_duration());
             if !self.run_one_task_queue() {
-                if spin {
-                    if let Some(t) = spin_since {
-                        if t.elapsed() < spin_before_park {
-                            continue;
-                        }
-                        spin_since = None
-                    } else {
-                        spin_since = Some(Instant::now());
-                        continue;
-                    }
-                }
-
-                self.parker.park();
-            } else {
-                spin_since = None
+                self.parker.park(self.spin_before_park());
             }
         })
     }
--- a/scipio/src/parking.rs
+++ b/scipio/src/parking.rs
@@ -63,8 +63,8 @@ impl Parker {
     }
 
     /// Blocks until notified and then goes back into unnotified state.
-    pub(crate) fn park(&self) {
-        self.inner.park(None);
+    pub(crate) fn park(&self, spin: Option<Duration>) {
+        self.inner.park(spin);
     }
 
     /// Performs non-sleepable pool and install a preempt timeout into the
@@ -73,7 +73,7 @@ impl Parker {
     /// will be able to check if the timer has elapsed and yield the CPU if that
     /// is the case.
     pub(crate) fn poll_io(&self, timeout: Duration) {
-        self.inner.park(Some(timeout));
+        self.inner.poll(Some(timeout));
     }
 }
 
@@ -96,12 +96,14 @@ impl fmt::Debug for Parker {
 struct Inner {}
 
 impl Inner {
-    fn park(&self, timeout: Option<Duration>) -> bool {
-        // If the timeout is zero, then there is no need to actually block.
-        // Process available I/O events.
-        let reactor_lock = Reactor::get().lock();
-        let _ = reactor_lock.react(timeout);
-        false
+    /// If the timeout is zero, then there is no need to actually block.
+    /// Process available I/O events.
+    fn poll(&self, timeout: Option<Duration>) {
+        let _ = Reactor::get().lock().react(timeout, false);
+    }
+
+    fn park(&self, spin: Option<Duration>) {
+        let _ = Reactor::get().lock().react(spin, true);
     }
 }
 
@@ -404,7 +406,7 @@ struct ReactorLock<'a> {
 
 impl ReactorLock<'_> {
     /// Processes new events, blocking until the first event or the timeout.
-    fn react(self, timeout: Option<Duration>) -> io::Result<()> {
+    fn react(self, timeout: Option<Duration>, spin: bool) -> io::Result<()> {
         // FIXME: there must be a way to avoid this allocation
         // Indeed it just showed in a profiler. We can cap the number of
         // cqes produced, but this is used for timers as well. Need to
@@ -415,9 +417,13 @@ impl ReactorLock<'_> {
         let next_timer = self.reactor.process_timers(&mut wakers);
 
         // Block on I/O events.
-        let ores = match timeout {
-            Some(preempt) => self.reactor.sys.step(&mut wakers, preempt),
-            None => self.reactor.sys.wait(&mut wakers, next_timer),
+        let ores = if spin {
+            self.reactor.sys.wait(&mut wakers, next_timer, timeout)
+        } else {
+            match timeout {
+                Some(preempt) => self.reactor.sys.step(&mut wakers, preempt),
+                None => self.reactor.sys.wait(&mut wakers, next_timer, None),
+            }
         };
 
         let res = match ores {
--- a/scipio/src/parking.rs
+++ b/scipio/src/parking.rs
@@ -415,7 +415,12 @@ impl ReactorLock<'_> {
         let next_timer = self.reactor.process_timers(&mut wakers);
 
         // Block on I/O events.
-        let res = match self.reactor.sys.wait(&mut wakers, timeout, next_timer) {
+        let ores = match timeout {
+            Some(preempt) => self.reactor.sys.step(&mut wakers, preempt),
+            None => self.reactor.sys.wait(&mut wakers, next_timer),
+        };
+
+        let res = match ores {
             // We slept, so don't wait for the next loop to process timers
             Ok(true) => {
                 self.reactor.process_timers(&mut wakers);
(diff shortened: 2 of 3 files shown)


scipio/src/parking.rs
```diff
@@ -93,15 +95,23 @@ impl fmt::Debug for Parker {
     }
 }
 
-struct Inner {}
+struct Inner {
+    wakers: RefCell<Vec<Waker>>,
```
sitano, line 99: @glommer you want to use SmallVec instead of Vec? what capacity do you expect it to have on stack then?

scipio/src/sys/uring.rs
```diff
@@ -944,7 +935,7 @@ impl Reactor {
         if should_sleep {
             // We are about to go to sleep. It's ok to sleep, but if there
             // is a timer set, we need to make sure we wake up to handle it.
-            if let Some(dur) = timer_expiration {
+            if let Some(dur) = preempt {
                 let mut src = self.timeout_src.borrow_mut();
                 lat_ring.cancel_preempt_timer(&mut src);
```
sitano, line 940: @glommer what if `should_sleep &= wakers.is_empty();` will evaluate to false in original code? then the preemption timer would not be rearmed. is it fine? it is safe now when we always have `poll_io(preempt)` before calling `run_one_task` because poll_io will rearm the timer, but aside of this, is it fine that it can left preempt timer not rearmed?
Maybe it is fine to cancel preempt timer at all before parking (sleeping) and I can do else cancel here for spinning case?

scipio/src/sys/uring.rs
```diff
@@ -960,9 +951,21 @@ impl Reactor {
...
+        }
+
+        if should_sleep {
+            if let Some(dur) = spin {
+                let start = Instant::now();
+                while start.elapsed() < dur {
+                    consume_rings!(into wakers; lat_ring, poll_ring, main_ring);
+
+                    if !wakers.is_empty() {
+                        return Ok(true);
```
sitano, line 963: @glommer I am not sure if I need to rearm preempt timer before exiting here

glommer:
Hi Ivan!

It hadn't occur to me that we could just add spinning support to Reactor::wait, but I must say I really like the idea! 
I'll think a bit more about whether or not there is any disadvantage that I am missing... but overall it will allow us to keep stats gathering on the reactor very clean.

Looking at the actual implementation now.

scipio/src/parking.rs
glommer: There are two options in my view.

The beautiful thing to keep in mind is this:

Note that we don't *have* to fetch all events from the ring buffer. Let's assume that reacting to each event takes 10us to process. Not the time to fetch the event, the time the user spends reacting to the event. For instance if you read something from the network, you will do something with it and send a response back, imagine that this think time takes on average 10us. 

That means that if you fetch 10,000 events, that is 100ms which is the maximum preempt timer setting.
What that means is that there is no situation on earth where it is ever worth it to read more than 10k events.

Events that are not consume can sit in the io_uring and we'll consume them next time.

seastar btw does something similar: Except for Seastar I used 128 as a hard limit because the preempt timer was less than 1ms.

Even for us, though, 10k is absurdly high: the io_uring itself only has a limited amount of buffer space. It will never ever generate 10k events in one row. We keep timers separately so it needs a bit more extra space, but still... even a 1k maximum would be enough.

Assuming every waker is 8 bytes ( a function pointer), that means 8kB stack space which is absolutely reasonable for the Reactor.

What to do with this info, though? There are. two choices:
* SmallVec with room for 1k or even 512 entries and if we ever have to read more, it will go on heap
* Our own data structure that has a maximum capacity for 1k or even 512 entries (backed by a static array) and has a `is_full` method. The io_uring implementation is changed such that before consuming a cqe we test for `is_full`: if we can't add more wakers that's fine, we just stop processing.

The second is a lot more work, I think, because the changes run deeper. But it would never trigger an allocation so it is better technically.

The first is simpler and to be honest, in practice I very very much doubt we'll ever have more than 1k wakers returning in one batch. We can probably just use SmallVec, which simplifies coding, and add a stat to the `ExecutorStats`, counting the number of times we went over the limit. Why complicate something before it is an issue?

I'll leave it up to you. To be honest, using a standard Vec works as well, so long as you reserve enough space for 1k entries in the beginning with `with_capacity`.

scipio/src/sys/uring.rs
glommer: The preempt timer only needs to be rearmed if there is an active task queue.
If we arrive at `wait` with no timeout, that means that the Executor wants us to sleep because there are no more active task queues.

The only thing that can change that, is if we generated wakers while we were processing the cancellations and remaining pending flushes (or if we have anything in the poll ring, because the poll ring can't sleep).

The timer we rearm here, is no the preempt timer:

```rust
            // We are about to go to sleep. It's ok to sleep, but if there
            if let Some(dur) = timer_expiration {
```

Those are user timers. Let's say for instance you, as a user, asked for this:
```rust
Timer::new(Duration::from_secs(3600))
```

and then we go to sleep right after.


We can't sleep forever: we have to rearm the timer for 1h from now otherwise you will miss it.
The fact that the preempt timer path is reused here is just an implementation detail so we can coalesce those things.

scipio/src/executor.rs
```diff
@@ -856,24 +852,9 @@ impl LocalExecutor {
...
-                    } else {
-                        spin_since = Some(Instant::now());
-                        continue;
-                    }
-                }
-
-                self.parker.park();
-            } else {
-                spin_since = None
+                self.parker.park(self.spin_before_park());
```
glommer, line 857: I love this! Very good.

scipio/src/parking.rs
```diff
@@ -58,13 +58,15 @@ impl Parker {
     pub(crate) fn new() -> Parker {
         // Ensure `Reactor` is initialized now to prevent it from being initialized in `Drop`.
         Parker {
-            inner: Rc::new(Inner {}),
+            inner: Rc::new(Inner {
+                wakers: RefCell::new(Vec::new()),
```
glommer, line 62: See my extensive comment about this.
The vector is ultimately fine, but please initialize it with `with_capacity`.

scipio/src/parking.rs
```diff
@@ -93,15 +95,23 @@ impl fmt::Debug for Parker {
...
-        // If the timeout is zero, then there is no need to actually block.
-        // Process available I/O events.
-        let reactor_lock = Reactor::get().lock();
-        let _ = reactor_lock.react(timeout);
-        false
+    /// If the timeout is zero, then there is no need to actually block.
+    /// Process available I/O events.
+    fn poll(&self, timeout: Option<Duration>) {
+        let _ = Reactor::get()
+            .lock()
```
glommer, line 107: if you are down for a cleanup, the lock stuff in the reactor can go.
It is an artifact of async-io as well.
It may simplify your code: you may not need the refcell on the vector if you dispose of it.

scipio/src/sys/uring.rs
```diff
@@ -914,23 +914,14 @@ impl Reactor {
     pub(crate) fn wait(
         &self,
         wakers: &mut Vec<Waker>,
-        timeout: Option<Duration>,
-        timer_expiration: Option<Duration>,
+        preempt: Option<Duration>,
+        spin: Option<Duration>,
```
glommer, line 918: This feels wrong. The fact that `timeout` and `timer_expiration` were separate here used to tell us if we are ok to go to sleep: we could sleep with no timeout, but we'd have to set a timer for `timer_expiration` if there was any.

(this whole thing is underdocumented, which is my fault).

Now, how will you be able to tell that you can sleep ?
Spin could signal that, but you still need to tell the no spin case from the spin so it works with both kinds of executors

scipio/src/parking.rs
```diff
@@ -404,21 +414,29 @@ struct ReactorLock<'a> {
...
+        spin: bool,
+    ) -> io::Result<()> {
         // Process ready timers.
-        let next_timer = self.reactor.process_timers(&mut wakers);
+        let next_timer = self.reactor.process_timers(wakers);
 
         // Block on I/O events.
-        let res = match self.reactor.sys.wait(&mut wakers, timeout, next_timer) {
+        let ores = if spin {
+            self.reactor.sys.wait(wakers, next_timer, timeout)
```
glommer, line 428: this definitely feels wrong... 
You are now passing the preempt timer as the `spin` parameter.

scipio/src/sys/uring.rs
glommer: I don't know, but it got a bit hard to follow.
I think it will be more productive if you rework this a bit with your newfound understanding of what the preempt timer is given the recent explanations.

I think separating the sleepable and non-sleepable parts of wait is good (but then wait should be renamed to `poll`, or something else that doesn't give the idea that this is waiting on anything)

* For the poll part, we have to register a timer if there is a user timer specified.
* In the case where we sleep (or spun for a while, and stopped sleeping), the preempt timer should be registered when we return. Imagine if you spin for 1ms, and your preempt timer was 2ms before the spin: you already stole 1ms of this poor task queue! So only at the end.

The cleanest way is maybe for the `poll` function to never deal with the preempt timer, only the user timer.
The sleep function would always deal with the preempt timer, never the user timer.

glommer (changes requested):

glommer:
I had to do almost everything needed for this for the TcpStreams. With that infrastructure in place, this was very simple to solve with a small extension so I did that.

Thanks for the effort Ivan!

sitano:
@glommer sorry Glauber. I am buried under the weights of my work.

glommer:
don't worry Ivan!

PR #108, opened 2020-10-13
Improvements to the Scipio scheduler

glommer:
### What does this PR do?

Improves the scheduler so that we are able to run with a higher resolution (nanos instead of micros). This will make us a bit slower but I argue that the cost of the extra branch is worth paying for generality and should be negligible if the tasks
are big enough (which they should)

Once that is done we fix a bug with yield and add a test case to make sure the bug won't resurface

### Checklist

### DIFF
--- a/scipio/src/executor.rs
+++ b/scipio/src/executor.rs
@@ -190,19 +190,14 @@ impl TaskQueue {
         }
     }
 
-    fn account_vruntime(&mut self, delta: Duration) -> u64 {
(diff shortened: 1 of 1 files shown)


glommer:
I have hard reason to believe that this would fix the regression that @matklad saw in https://github.com/DataDog/scipio/pull/107 so I am fast tracking this despite no reviews.

We can always revert or add follow up fixes later.

matklad:
Confirm that this fixes the test for me

PR #109, opened 2020-10-13
Reduce code duplication between DmaFile and Directory

matklad:
### What does this PR do?

Cleanup discussed in #102 

### Checklist

### DIFF
--- a/scipio/src/io/dma_file.rs
+++ b/scipio/src/io/dma_file.rs
@@ -3,6 +3,7 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
+use crate::error::ErrorEnhancer;
 use crate::io::read_result::ReadResult;
 use crate::parking::Reactor;
 use crate::sys;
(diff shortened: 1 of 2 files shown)


scipio/src/io/dma_file.rs
```diff
@@ -462,6 +434,79 @@ impl DmaFile {
     }
 }
 
+/// A wrapper over `std::fs::File` which carries a path (for better error
+/// messages) and prints a warning if closed synchronously.
+#[derive(Debug)]
+struct EnhancedFile {
```
matklad, line 440: I am not thrilled about the name tbh. I *think* I would like `ScipioFile` more, but it seems that we generally don't use "Scipio" prefix anywhere else?

glommer: `ScipioFile` is fine as long as it doesn't get confusing.

`Directory` is well, a directory.
`DmaFile` allows for dma operations (O_DIRECT) randomly issued at direct positions.
`StreamReader` and `StreamWriter` allow for linear access (currently only over O_DIRECT).
I have a queued patchset for `BufferedFile` which will use buffered I/O for the simple applications that don't want to deal with O_DIRECT.

So I wouldn't be opposed to `ScipioFile` as something common to *all* of those.

Closing seems like something that will be shared between `Directory`, `BufferedFile` and `DmaFile`. It's fair to say that the streams are not *really* files, but rather abstractions on top of file.

So maybe?

glommer:
Good change and I could merge it as is.

Please consider folding your last patch into the first so it is called `ScipioFile` from the beginning.
If you post a fast forward with an empty diff from b9c25f6 I'll merge

matklad:
Rebased! (In general, I am very pro-rebase and skilled with magit, so feel free to request history cleanups)

glommer:
Shh, don't let the linux folks hear it or you'd end up doing kernel work

PR #111, opened 2020-10-14
Scipio controllers: deadline queue

glommer:
### What does this PR do?

This PR adds Scipio's first controller. A controller is an entity that can dynamically and automatically adjust shares of a task queue lower or higher so that the tasks in that task queue proceed accordingly to some user-defined criteria.

The first of such controllers is the DeadlineQueue: It is a version of a work queue, in which each element is given a deadline. The goal of the controller is to minimize the impact of running the work queue by reducing shares as much as possible so long as each element finishes within its deadline

It comes with an interactive example you can play with!

### Motivation

Process controllers were used extensively in Scylla to control background processes such as buffer flushes and LSM compactions, making sure that the database had stable latencies throughout its operation. However, those were done
at the application level (Scylla), as opposed to the framework level (seastar)

The challenge I set myself to solve for Scipio was to do this in a generic enough fashion so that it can live in the library and
be easily consumable by application developers.

This means a considerable amount of effort will be poured into wrapping up the controllers in a series of APIs that makes sense for general purpose use.

The first such API is the DeadlineQueue: a workqueue-like structure that queues items implementing a particular trait. (DeadlineSource)

Items in that queue tell the controller how much work they are supposed to do (in user-defined units) and how how much work they have done already. Items execute sequentially. The controller does the rest.

### DIFF
--- a/examples/deadline_writer.rs
+++ b/examples/deadline_writer.rs
@@ -1,0 +1,208 @@
+use ansi_term::{Colour, Style};
+use scipio::controllers::{DeadlineQueue, DeadlineSource};
+use scipio::{Latency, Local, LocalExecutorBuilder, Shares, Task, TaskQueueHandle};
+use std::cell::Cell;
+use std::future::Future;
+use std::io;
+use std::pin::Pin;
+use std::rc::Rc;
+use std::time::{Duration, Instant};
+
+fn burn_cpu(dur: Duration) {
+    let now = Instant::now();
+    while now.elapsed() < dur {}
+}
+
+struct IntWriter {
+    deadline: Duration,
+    start: Instant,
+    count_target: usize,
+    count: Cell<usize>,
+    next_print: Cell<Duration>,
+    count_at_last_print: Cell<usize>,
+
+    last_tq_runtime: Cell<Duration>,
+    last_ex_runtime: Cell<Duration>,
+}
+
+impl IntWriter {
+    fn new(count_target: usize, deadline: Duration) -> Rc<IntWriter> {
+        Rc::new(IntWriter {
+            start: Instant::now(),
+            deadline,
+            count_target,
+            count: Cell::new(0),
+            next_print: Cell::new(Duration::from_secs(1)),
+            count_at_last_print: Cell::new(0),
+            last_tq_runtime: Cell::new(Duration::from_nanos(0)),
+            last_ex_runtime: Cell::new(Duration::from_nanos(0)),
+        })
+    }
+
+    async fn write_int(&self) -> Duration {
+        let my_handle = Local::current_task_queue();
+
+        loop {
+            let me = self.count.get();
+            let elapsed = self.start.elapsed();
+            if me >= self.count_target {
+                return elapsed;
+            }
+            self.count.set(me + 1);
+            if elapsed > self.next_print.get() {
+                let tq_stats = Local::task_queue_stats(my_handle).unwrap();
+
+                let tq_runtime = tq_stats.runtime();
+                let tq_delta = tq_runtime - self.last_tq_runtime.get();
+                let ex_runtime = Local::executor_stats().total_runtime();
+                let ex_delta = ex_runtime - self.last_ex_runtime.get();
+
+                let ratio = self.count.get() as f64 / self.count_target as f64 * 100.0;
+                let intratio = self.count.get() - self.count_at_last_print.get();
+
+                let cpuratio = 100.0 * tq_delta.as_secs_f64() / ex_delta.as_secs_f64();
+
+                println!(
+                    "{}: Wrote {} ({}%), {:.0} int/s, scheduler shares: {} , {:.2} % CPU",
+                    Colour::Blue.paint(format!("{}s", elapsed.as_secs())),
+                    self.count.get(),
+                    Style::new().bold().paint(format!("{:.0}", ratio)),
+                    intratio,
+                    Style::new()
+                        .bold()
+                        .paint(tq_stats.current_shares().to_string()),
+                    cpuratio
+                );
+                self.next_print
+                    .set(self.next_print.get() + Duration::from_secs(1));
+                self.count_at_last_print.set(self.count.get());
+                self.last_tq_runtime.set(tq_runtime);
+                self.last_ex_runtime.set(ex_runtime);
+            }
+
+            burn_cpu(Duration::from_micros(500));
+            Local::later().await;
+        }
+    }
+}
+
+impl DeadlineSource for IntWriter {
+    type Output = Duration;
+
+    fn expected_duration(&self) -> Duration {
+        self.deadline
+    }
+
+    fn action(&self) -> Pin<Box<dyn Future<Output = Duration> + '_>> {
+        Box::pin(self.write_int())
+    }
+
+    fn total_units(&self) -> u64 {
+        self.count_target as _
+    }
+
+    fn processed_units(&self) -> u64 {
+        self.count.get() as _
+    }
+}
+
+fn competing_cpu_hog(
+    stop: Rc<Cell<bool>>,
+    cpuhog_tq: TaskQueueHandle,
+) -> scipio::task::JoinHandle<(), ()> {
+    Local::local_into(
+        async move {
+            while !stop.get() {
+                burn_cpu(Duration::from_micros(500));
+                Local::later().await;
+            }
+        },
+        cpuhog_tq,
+    )
+    .unwrap()
+    .detach()
+}
+
+async fn static_writer(how_many: usize, shares: usize, cpuhog_tq: TaskQueueHandle) -> Duration {
+    let name = format!("shares-{}", shares);
+    let tq = Local::create_task_queue(Shares::Static(shares), Latency::NotImportant, &name);
+
+    let stop = Rc::new(Cell::new(false));
+    let hog = competing_cpu_hog(stop.clone(), cpuhog_tq);
+
+    let writer = Task::local_into(
+        async move {
+            // Last parameter is bogus outside the queue, but we're just reusing the same writer
+            let test = IntWriter::new(how_many, Duration::from_secs(0));
+            test.write_int().await
+        },
+        tq,
+    )
+    .unwrap()
+    .detach();
+
+    let res = writer.await.unwrap();
+    stop.set(true);
+    hog.await.unwrap();
+    res
+}
+
+// FIXME: this is just much simpler, but in a standard scipio program you should avoid blocking.
+// This has actually some nasty effects in that all the blocking time is accounted in this task
+// queue. As soon as we have a good buffered file implementation we should come here and fix this
+// so nobody thinks lowly of us.
+fn blocking_read_int() -> Result<usize, <usize as std::str::FromStr>::Err> {
+    let mut buffer = String::new();
+    io::stdin().read_line(&mut buffer).unwrap();
+    let buf = buffer.trim();
+    buf.parse::<usize>()
+}
+
+fn main() {
+    let handle = LocalExecutorBuilder::new().pin_to_cpu(0)
+        .spawn(|| async move {
+
+            let cpuhog_tq = Local::create_task_queue(Shares::Static(1000), Latency::NotImportant, "cpuhog");
+
+            println!("{}", Style::new().bold().paint("Welcome to the Deadline Writer example"));
+            println!("In this example we will write a sequence of integers to a variable, busy looping for 500us after each write");
+            println!("While we do that, another CPU hog will be running constantly in a different TaskQueue");
+            println!("For {} results, this test is pinned to your CPU0. Make sure nothing else of significance is running there. You should be able to see it at 100% at all times!",
+                     Style::new().bold().paint("best"));
+
+            println!("\n\nPlease tell me how many integers you would like to write");
+            let to_write = blocking_read_int().unwrap();
+            println!("Ok, now let's write {} integers with both the writer and the CPU hog having the same priority", Colour::Blue.paint(to_write.to_string()));
+            let dur = static_writer(to_write, 1000, cpuhog_tq).await;
+            println!("Finished writing in {}", Colour::Green.paint(format!("{:#.0?}", dur)));
+            println!("This was using {} shares, and short of reducing the priority of the CPU hog. {}",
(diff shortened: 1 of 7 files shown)


glommer:
Heads up: coming up with unit tests for controllers is *HARD* but quitters never win, and winners never quit.
I'll do my best to test at least some boundary conditions.

In the meantime the example should be super nice to play with and is what I have been mostly using for test purposes.

examples/deadline_writer.rs
```diff
@@ -0,0 +1,212 @@
...
+                    deadline.push_work(test).await.unwrap()
+                }}).detach();
+
+                Local::later().await;
+
+                let stop = Rc::new(Cell::new(false));
+                let hog = competing_cpu_hog(stop.clone(), cpuhog_tq);
+                let dur = writer.await.unwrap();
+                println!("Finished writing in {}", Colour::Green.paint(format!("{:#.2?}", dur)));
+                stop.replace(true);
```
matklad, line 201: Is there any particular reason to use `.replace` over `.set`? This feels slightly unusual.

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// What type is returned by the [`action`] method
+    ///
+    /// [`action`]: trait.DeadlineSource.html#method.action
+    type Output;
+
+    /// Returns a [`Duration`] indicating when we would like this operation to complete.
+    ///
+    /// It is calculated from the point of Queueing, not from the point in which the operation
+    /// starts.
+    fn deadline(&self) -> Duration;
```
matklad, line 36: Without reading the code closely, the signature feels weird. I'd expect `fn deadline(&self) -> Instant`

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+    ///
+    /// For example, you can buffer all updates and just inform that processed_units == total_units
+    /// at the end of the process, but then the controller would be a step function.
+    ///
+    /// [`total_units`]: trait.DeadlineSource.html#method.total_units
+    fn processed_units(&self) -> u64;
+}
+
+impl<T> Debug for dyn DeadlineSource<Output = T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
```
matklad, line 77: Minor style nit, but I find the following way to write this to be aesthetically nicest:
```suggestion
impl<T> fmt::Debug for dyn DeadlineSource<Output = T> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+            adjustment_period,
+            last_error: Cell::new(0.0),
+            min_shares: Cell::new(1),
+        }
+    }
+
+    fn admit(&self, source: Rc<dyn DeadlineSource<Output = T>>) -> io::Result<()> {
+        let expiration = Instant::now()
+            .checked_add(source.deadline())
+            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, ""))?;
```
matklad, line 255:
```suggestion
            .ok_or_else(|| io::Error::from(io::ErrorKind::InvalidInput))?;
```

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+            if adjustment_period < Duration::from_millis(100) {
+                Latency::Matters(adjustment_period)
+            } else {
+                Latency::NotImportant
+            }
+        };
+
+        let tq = Local::create_task_queue(Shares::Dynamic(queue.clone()), lat, name);
+        let (sender, receiver): (LocalSender<QueueItem<T>>, LocalReceiver<QueueItem<T>>) =
+            LocalChannel::new_bounded(1);
```
matklad, line 369: channels APIs usually make the ctor function top-level. That way, you'll have a nice place to put a type parameter once: `local_channel::new_bounded::<QueueItem<T>>(1)`

examples/deadline_writer.rs
glommer: no, just force of habit as sometimes I do want the value back then end up always using replace from muscle memory. Will change.

scipio/src/controllers/deadline_queue.rs
glommer: right, so maybe the name needs to be changed.
This represents how long the operation is supposed to run for - we use that to calculate speed later so this has to be a duration.

When we enqueue it we add that to now() to obtain an instant, which is then placed in the queue collection.

Maybe just call it `duration` ?

scipio/src/controllers/deadline_queue.rs
glommer: agree

scipio/src/controllers/deadline_queue.rs
glommer: so just a free function without an associated struct?
I'll change that in the channel PR.

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,586 @@
...
+/// [`next`]: https://docs.rs/futures/0.3.6/futures/stream/trait.StreamExt.html#method.next
+/// [`StreamExt`]: https://docs.rs/futures/0.3.6/futures/stream/trait.StreamExt.html
+pub struct LocalChannel<T> {
+    state: Rc<RefCell<State<T>>>,
+}
+
+impl<T> LocalChannel<T> {
+    fn new(capacity: ChannelCapacity) -> (LocalSender<T>, LocalReceiver<T>) {
+        let channel = match capacity {
+            ChannelCapacity::Unbounded => VecDeque::new(),
```
duarten, line 169: Is there a good use case for unbounded channels?

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,586 @@
...
+        }
+
+        state.channel.push_back(item);
+        if !state.recv_waiters.is_empty() {
+            let waiters = std::mem::replace(&mut state.recv_waiters, Vec::new());
+            drop(state);
+
+            for w in waiters {
+                w.wake();
+            }
```
duarten, line 213: Shouldn't it wake just one?

scipio/src/channels/local_channel.rs
```diff
@@ -0,0 +1,586 @@
...
+        }
+    }
+}
```
duarten, line 394: Can a `LocalSender` be dropped while there are still `send_waiters`? Something like calling `send_eventually` without awaiting the result.

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+impl<T> PriorityBump<T> {
+    fn new(queue: Rc<InnerQueue<T>>) -> PriorityBump<T> {
+        queue.min_shares.replace(250);
+        PriorityBump { queue }
+    }
+}
+
+impl<T> Drop for PriorityBump<T> {
+    fn drop(&mut self) {
+        self.queue.min_shares.replace(1);
```
duarten, line 111: I guess no one will do 

```ts
let pb = PriorityBump::new(q)
let pb2 = PriorityBump::new(pb.queue)
drop(pb) // back to 1
```

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+        self.accumulated_error.replace(acc);
+        let delta_error = error - self.last_error.get();
+        self.last_error.replace(error);
+
+        // How did we pick our constants:
+        //  * As with any stable PI controller we want the bulk of our gain to come from P.
+        //  * As we normalize the maximum error to 1 we know that the gain should be at most 1000
+        //  * physically, Ki can be expressed as Kp / Tau where Tau is a time constant, roughly
+        //    equivalent to how many periods need to pass for the integral term to generate the
+        //    same gain as the proportional term. We set that too 6 so the controller is not too
```
duarten, line 214: s/too/to

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+    //    be comparable as we accumulate artifacts of the delta_t calculation (as we'll
+    //    never in practice keep delta_t constant)
+    //
+    //  The way we'll solve this is by expressing an alternate error E(T) which is
+    //  essentially the integral of e(t) in time:
+    //
+    //    E(t) = 1 - units_processed / units_expected.
+    //
+    //  That is easy to compute, as it is essentially the total count of units for
+    //  all items in the queue, both processed an expected.
```
duarten, line 167: s/an/and

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+        self.results.pop_front().await?
+    }
+}
+
+#[derive(Debug)]
+/// Scipio's scheduler is based on [`Shares`]: the more shares, the more resources the task
+/// will receive.
+///
+/// There are situations however in which we don't want shares to grow too high: for instance,
+/// a background process that is flushing a file to storage. If it were to run a full speed,
```
duarten, line 285: s/a/at

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+///   between the application of the control decision and its effects being seen. In our case, the
+///   scheduler may not schedule us for a long time, especially if the shares are low.
+/// * Control loops work better the faster and smoother the response is. Let's use an example flushing a
+///   file: you may be moving data to the file internal buffers but they are not *flushed* to the
+///   media. When the data is finally flushed a giant bubble is inserted into the control loop. The
+///   characteristics of the system will radically change. Contract that for instance with O_DIRECT
+///   files, where writing to the file means writing to the media: smooth and fast feedback!
+///
+///   The moral of the story is:
+///    * do not use this with buffered files or other buffered sinks where where the real physical response
```
duarten, line 316: s/where/

duarten (commented):
very cool :)

scipio/src/channels/local_channel.rs
matklad: Non-DAG communication graphs (actor systems) need at least some channels to be unbounded, otherwise they deadlock: https://trio.discourse.group/t/sizing-the-channel-deadlock-freedom-vs-back-pressure/311

glommer: More importantly, this is not an opinion that I would like Scipio to have.
Other crates for channels have this which indicates that some people have uses for unbounded channels.


Scipio should strive for being generic and making many use cases possible.

scipio/src/channels/local_channel.rs
glommer: That's a very good question. It's clearly not wrong to wake them all since they'll just be put to sleep.
Intuitively it feels to me that if we are pushing and consuming a single item, a single notification would be enough and you are right.

I need to convince myself that a lost notification is not possible in the parallel executors, though.

scipio/src/controllers/deadline_queue.rs
glommer: I would consider this a bug.
Notice how you can't even specify the priority you want in the API, so this is not like a verbose flag that you can pass many.

scipio/src/channels/local_channel.rs
glommer: I spent some time thinking about this.

I don't think it should be possible:
- Rust futures are lazy, so there is no such thing as "not waiting on them". If you don't .await they don't execute.
- For that situation to trigger you have to spawn a new detached task and have the `LocalSender` and the send operation running in different tasks
- But for you to pass that into a spawn-like method you need an `Rc` pointer: spawn-like methods require static lifetime for all references passed to it.
- Which means if you drop at this point you'd be dropping the reference count. The senders would keep the refcount up and prevent an early drop.

scipio/src/channels/local_channel.rs
duarten: Not totally clear on what the parallel executors are, but in general, a _check - signal sleep - redo check - sleep_ pattern works well when there's parallelism.

scipio/src/channels/local_channel.rs
duarten: It could have that opinion, similar to how Go has it. Unbounded queues are usually anathema for latency-sensitive solutions (unbounded queues do solve deadlocks in CSP systems, in a similar way that a Big Lock can solve data races).

glommer: you are right about the latency issue but I would like scipio to be useful to build latency sensitive applications, but *not only* to build latency sensitive applications.

Sometimes the complexity of the application is elsewhere.

scipio/src/channels/local_channel.rs
glommer: Actually I think we should be waking just one here. Otherwise imagine a blocked channel with 1,000,000 waiters that gets 1,000,000 elements pushed: it will call `wake` a lot as it becomes quadratic.

scipio/src/controllers/deadline_queue.rs
glommer: Called it `expected_duration`

glommer:
rebased. All the dependencies are now in, so this is just the controller code.

glommer:
forced pushed a version:
* added unit tests, especially for the boundary conditions
* improved the example
* changed the API so this is easily usable behind `Rc` through interior mutability (no `&mut self`) for `push`.

scipio/src/controllers/deadline_queue.rs
```diff
@@ -0,0 +1,637 @@
...
+        }
+    }
+
+    fn admit(&self, source: Rc<dyn DeadlineSource<Output = T>>) -> io::Result<()> {
+        let expiration = Instant::now()
+            .checked_add(source.expected_duration())
+            .ok_or_else(|| io::Error::from(io::ErrorKind::InvalidInput))?;
+        expiration
+            .checked_duration_since(self.last_admitted.get())
+            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, ""))?;
```
matklad, line 256:
```suggestion
            .ok_or_else(|| io::Error::from(io::ErrorKind::InvalidInput))?;
```

PR #112, opened 2020-10-14
relax restrictions for naming executors.

glommer:
Requiring a static string is good for saving an allocation but essentially
means we can't dynamically name executors, like format!("my-executor-{}", 1)

If we require just an str instead, without the static bound and then copy
the name internally we can lift this restriction. That is a lot more flexible.

### DIFF
--- a/scipio/src/executor.rs
+++ b/scipio/src/executor.rs
@@ -120,7 +120,7 @@ struct TaskQueue {
     runtime: u64,
     vruntime: u64,
     io_requirements: IoRequirements,
-    name: &'static str,
+    name: String,
     index: usize,             // so we can easily produce a handle
     last_adjustment: Instant, // for dynamic shares classes
     yielded: bool,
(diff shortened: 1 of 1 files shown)


scipio/src/executor.rs
```diff
@@ -150,7 +150,7 @@ impl Eq for TaskQueue {}
 impl TaskQueue {
     fn new<F>(
         index: usize,
-        name: &'static str,
+        name: &str,
```
matklad, line 153: Mostly a style nit, but here and elsewhere I'd take a `String` as an argument, rather then implicitly allocating: https://github.com/rust-analyzer/rust-analyzer/blob/dedfaa384408b266c461fea45fee936989bb151b/docs/dev/style.md#push-allocations-to-the-call-site

glommer: this makes sense, actually.

Will change, thanks!

glommer: changed to use `Into<String>` which solves the allocation problem without forcing an API change.

PR #115, opened 2020-10-15
standardize on futures_lite

glommer:
This patch removes "futures" from Cargo.toml, meaning it won't be possible
anymore to use it by accident in our implementations.

There are still a lot of users in test code, so for now we'll leave it
in dev-dependencies so we don't need to rush to convert.

Outside of testing, really the only contentious use was in file_stream.

### DIFF
--- a/scipio/src/io/file_stream.rs
+++ b/scipio/src/io/file_stream.rs
@@ -10,8 +10,8 @@ use crate::sys::DmaBuffer;
 use crate::task;
 use crate::Local;
 use core::task::Waker;
-use futures::future::join_all;
-use futures::io::{AsyncRead, AsyncWrite};
+use futures_lite::io::{AsyncRead, AsyncWrite};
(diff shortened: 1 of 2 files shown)


scipio/src/io/file_stream.rs
```diff
@@ -310,7 +310,12 @@ impl StreamReader {
         let mut state = self.state.borrow_mut();
         let handles = state.cancel_all_in_flight();
         drop(state);
-        join_all(handles).await;
+        let to_cancel = handles.len();
+        let cancelled = stream::iter(handles)
+            .then(|f| async { f.await })
+            .fold(0, |acc, _| acc + 1)
+            .await;
```
matklad, line 317:
```suggestion
        let cancelled = stream::iter(handles)
            .then(|f| f)
            .count()
            .await;
```

glommer:
merged Aleksey's suggestion and fixed cargofmt

PR #116, opened 2020-10-15
Fix use after free in timers

matklad:
### What does this PR do?

    We give the kernel a *pointer* to a timespec struct. Keeping this
    struct on the stack leads to a textbook use-after-free.
    
    Instead, lets stash timespec into Source, the same thing we do for all
    other resources shared with the kernel.
    
    We also introduce a helper `TimeSpec64` struct for more idiomatic
    API (notably to make `derive(Debug)` work on Source).

### DIFF
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -7,7 +7,6 @@ use nix::poll::PollFlags;
 use rlimit::Resource;
 use std::cell::{RefCell, UnsafeCell};
 use std::collections::{HashMap, VecDeque};
-use std::convert::TryInto;
 use std::ffi::CStr;
 use std::io;
 use std::io::{Error, ErrorKind};
@@ -17,7 +16,7 @@ use std::task::Waker;
 use std::time::Duration;
 
 use crate::sys::posix_buffers::PosixDmaBuffer;
-use crate::sys::{InnerSource, LinkStatus, PollableStatus, Source, SourceType};
+use crate::sys::{InnerSource, LinkStatus, PollableStatus, Source, SourceType, TimeSpec64};
 use crate::{IoRequirements, Latency};
 
 use uring_sys::IoRingOp;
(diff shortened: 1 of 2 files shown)


scipio/src/sys/uring.rs
```diff
@@ -670,22 +667,22 @@ impl UringCommon for SleepableRing {
...
+                    *status = LinkStatus::Freestanding;
                     Some(())
                 }
                 SourceType::LinkRings(LinkStatus::Freestanding) => {
                     panic!("Impossible to have an event firing like this");
                 }
-                SourceType::Timeout(Some(_)) => {
-                    source.source_type = SourceType::Timeout(None);
+                SourceType::Timeout(id @ Some(_), _) => {
+                    *id = None;
```
glommer, line 679: what kind of dark magic is this?
I am not ware of @

scipio/src/sys/uring.rs
```diff
@@ -990,3 +991,45 @@ impl Reactor {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use std::time::Instant;
+
+    use super::*;
+
+    #[test]
+    fn timout_smoke_test() {
```
glommer, line 1002:
```suggestion
    fn timeout_smoke_test() {
```

glommer (approved):

scipio/src/sys/uring.rs
matklad: Yeah, `@` is the least used sigil in the whole Rust (well, depending on the style, `ref` might be even rare). 

https://doc.rust-lang.org/reference/patterns.html#identifier-patterns

It allows to bind a thing to an identifier, and *also* to match the thing. 

I've used it here to avoid overwriting the whole `SourceType` -- we rely on `SourceType` addresses being stable, so surgically fixing specific bits of SourceType felt more reliable than replacing it wholesale.

scipio/src/sys/uring.rs
matklad: Fixed!

scipio/src/sys/uring.rs
glommer: Have you no love for your soul? 
If you linger too long in the shadows, you may become the shadow!

(j/k it's all cool. I knew about ref, did not know about that one).

PR #118, opened 2020-10-16
Buffered random io

glommer:
### What does this PR do?

Implements the first part of Buffered I/O support: random buffered I/O
After this is nailed, we'll provide stream interfaces like we do for DmaFile.

### Motivation

Not everyone is writing a data-intensive application.

### DIFF
--- a/scipio/src/io/buffered_file.rs
+++ b/scipio/src/io/buffered_file.rs
@@ -7,7 +7,7 @@
 use crate::io::scipio_file::ScipioFile;
 use crate::parking::Reactor;
 use std::io;
-use std::os::unix::io::{AsRawFd, RawFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::path::Path;
 
 /// Constructs a file that is backed by the operating system page cache
@@ -19,6 +19,14 @@ pub struct BufferedFile {
 impl AsRawFd for BufferedFile {
     fn as_raw_fd(&self) -> RawFd {
         self.file.as_raw_fd()
+    }
+}
+
+impl FromRawFd for BufferedFile {
+    unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        BufferedFile {
+            file: ScipioFile::from_raw_fd(fd),
+        }
     }
 }
 
--- a/scipio/src/io/buffered_file.rs
+++ b/scipio/src/io/buffered_file.rs
@@ -173,6 +173,10 @@ impl BufferedFile {
     pub async fn close(self) -> io::Result<()> {
         self.file.close().await
     }
+
+    pub(crate) fn path(&self) -> &Path {
+        self.file.path.as_ref().unwrap().as_path()
+    }
 }
 
 #[cfg(test)]
--- a/scipio/src/io/buffered_file.rs
+++ b/scipio/src/io/buffered_file.rs
@@ -1,0 +1,286 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
+// MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
+//
+
+use crate::io::scipio_file::ScipioFile;
+use crate::parking::Reactor;
+use std::io;
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::path::Path;
+
+/// Constructs a file that is backed by the operating system page cache
+#[derive(Debug)]
+pub struct BufferedFile {
+    file: ScipioFile,
+}
+
+impl AsRawFd for BufferedFile {
+    fn as_raw_fd(&self) -> RawFd {
+        self.file.as_raw_fd()
+    }
+}
+
+impl BufferedFile {
+    /// Returns true if the BufferedFile represent the same file on the underlying device.
+    ///
+    /// Files are considered to be the same if they live in the same file system and
+    /// have the same Linux inode. Note that based on this rule a symlink is *not*
+    /// considered to be the same file.
+    ///
+    /// Files will be considered to be the same if:
+    /// * A file is opened multiple times (different file descriptors, but same file!)
+    /// * they are hard links.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use scipio::LocalExecutor;
+    /// use scipio::io::BufferedFile;
+    /// use std::os::unix::io::AsRawFd;
+    ///
+    /// let ex = LocalExecutor::make_default();
+    /// ex.run(async {
+    ///     let mut wfile = BufferedFile::create("myfile.txt").await.unwrap();
+    ///     let mut rfile = BufferedFile::open("myfile.txt").await.unwrap();
+    ///     // Different objects (OS file descriptors), so they will be different...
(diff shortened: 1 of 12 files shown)


glommer:
Streams are now available too.

The API is as similar as possible to the Direct I/O stream and they implement `AsyncRead`, `AsyncBufRead`, `AsyncWrite` and `AsyncSeek`

scipio/src/io/mod.rs
```diff
@@ -88,7 +148,45 @@ pub async fn remove<P: AsRef<Path>>(path: P) -> io::Result<()> {
...
+};
 pub use self::read_result::ReadResult;
 pub use crate::sys::DmaBuffer;
+
+/// Allows asynchronously reading from the standard input
+pub fn stdin() -> StreamReader {
+    use std::os::unix::io::FromRawFd;
+
+    unsafe { StreamReaderBuilder::new(BufferedFile::from_raw_fd(0)).build() }
+}
```
matklad, line 168: Hm, wouldn't `drop(stdin()); drop(stdin())` close fd 0 twice? I am also not sure how that would interract with stdlib, which also would like to own 0, 1 and 2

scipio/src/io/buffered_file_stream.rs
```diff
@@ -0,0 +1,734 @@
...
+            },
+            Some(source) => Poll::Ready(self.consume_flush_result(source)),
+        }
+    }
+
+    // We want our own close to be called so we can mem::forget the underlying file
+    #[allow(unreachable_code)]
+    fn poll_close(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        panic!("Should never be called");
+        Poll::Ready(Ok(()))
```
matklad, line 476:
```suggestion
```

scipio/src/io/mod.rs
glommer: You are right about dropping. Good catch.
Ultimately I don't think we should be in the business of closing 0 , 1 and 2.
I'll see if I can come up with a version of those files that don't close on drop.

As a matter of fact maybe the streams should never close, and we should push that to the `BufferedFile`. Closing is already a bit annoying right now because some of the AsyncBufExt will consume the stream, meaning it won't be able to close it... 

I'll think about it a bit more, thanks!

Note sure what stdlib does with 0 1 and 2, but if we don't close them I think this shouldn't matter. It is always legal to dispatch I/O in an open file system. The kernel doesn't really understand any of the userspace ownership...

glommer: The cleanest way to handle this would be something akin to `mem::forget` for the file descriptor if we are using stdio fds, but `mem::forget` consumes the object so we can't keep using it. The easiest way them seem to be just special casing 0 1 and 2 inside `ScipioFile`'s `drop`

glommer: (which can't be done, because drop has a mutable reference to self as a parameter so we can't decide not to drop the underlying file without a `mem::replace`) or similar. Rust drives me crazy sometimes.

It'd be a bit of code duplication, but the best could maybe be to implement a whole new `Stdin` type that deals directly with the `RawFd`

glommer:
I yanked `stdout` and `stderr`. To be honest, I don't find them terribly useful (explanation in the commit log)
Stdin is implemented as its own class. There is a little bit of code repetition but nothing crazy since most of the logic is already in the `Buffer`. It then doesn't suffer from the `close` issue because we only ever deal with the RawFd directly.

glommer:
Implemented Seek for the write stream too.
Feature-wise, I feel this is ready to go.

scipio/src/io/buffered_file.rs
```diff
@@ -0,0 +1,278 @@
...
+    ///
+    /// Note that this differs from [`DmaFile`]'s read APIs: that reflects the
+    /// fact that buffered reads need no specific alignment and io_uring will not
+    /// be able to use its own pre-allocated buffers for it anyway.
+    ///
+    /// [`DmaFile`]: struct.DmaFile.html
+    pub async fn read_at(&self, buf: &mut [u8], pos: u64) -> io::Result<usize> {
+        let source = Reactor::get().read_buffered(self.as_raw_fd(), buf, pos);
+        enhanced_try!(source.collect_rw().await, "Reading", self.file)
+    }
```
matklad, line 113: Hm....

I think this API is fundamentally unsound. If the user cancells this `read_at` operation (by dropping the future), wouldn't consider `buf` mutably borrowed anymore, but the kernel would happily write into it, until we *await* for cancellation. And we can't await for cancellation in Drop because drop is sync, and because that would  be unsound anyway due to mem::forget or rc cycles. 

See https://without.boats/blog/io-uring/ and https://without.boats/blog/ringbahn/ for some discussion of this problem.

I think the fix would be to incur a copy -- we should use our owned DMA buffer internally, and copy the data to a user-provided buf when we are done. 

(note -- this is cursory glance at the code, maybe I am mistaken here)

glommer: I am seriously unhappy about the amount of copies that are forced in the read path by the APIs. There are some in the `AsyncReadBuf` too... 
But you are probably right. In which case it would probably better to have an API more similar to the DmaFile, where we allocate our own buffers. They would just be simple buffers instead of the DmaBuffer with their alignment guarantees.

matklad: >But you are probably right. In which case it would probably better to have an API more similar to the DmaFile, where we allocate our own buffers. They would just be simple buffers instead of the DmaBuffer with their alignment guarantees.

Yeah, agree. I think the simplest possible fix here is to just use the following API 

```rust
pub async fn read_at(&self, pos: u64, len: usize) -> io::Result<Vec<u8>> {
```

Long term, we should replace `Vec<u8>` with some dedicated type, which represents a buffer from a buffer pool owned by scipio.

scipio/src/io/mod.rs
glommer: This is now done

scipio/src/io/buffered_file.rs
glommer: This is a bit worse, I think.
I can't convince myself write wouldn't have the same problem.

But that also means that the current interface for `write_dma` is unsound.
We're taking a `DmaBuffer` reference as a parameter there. The best solution is probably to just take ownership of the buffer and store it in the `Source`. We keep the `Source` alive throughout the whole operation, so that's safe.

And for our `write_at` here in the `BufferedFile`, same thing: pass the ownership of a vector, not a reference...

Ok, this will be a bit of a big structural change. In the end both APIs will be very similar, which is good news.

I still want to keep them separate as APIs because whether or not you are using Direct I/O is a big decision, shouldn't be swept under the hood. But common implementations will help.

glommer:
Reworked the series on top of https://github.com/DataDog/scipio/pull/128, which changes the `DmaFile` API for consistency.
`BufferedFile` follows suit.

Other than that, not a lot new.

glommer:
* reworked on top of https://github.com/DataDog/scipio/pull/128
* Also based on #128, changed `write_at` buffered API so that we don't do buffer reuse (a price to pay for simplicity)
* also a lot of conflicts with Aleksey cleanups at the uring level. Fixed all

glommer:
Merging, as the API is largely similar to the now already extensively reviewed Dma API.

Still happy to address issues later should they arise!

PR #121, opened 2020-10-18
Remove UnsafeCell from Source

matklad:
### What does this PR do?

Replaces Source's UnsafeCell with more interior mutability

### Motivation

Getting rid of the most questionable unsafe block in scipip

### Additional Notes

I did just a mechanical API safe-isation, it is nominally safe, but is still prone to panics due to refmut borrows. I *think*  this might be improvable by either cleverer use of `Cell`s, or by replacing the whole source with a more well abstracted syncronization primitive -- Source looks a whole lot like Channel, with the main difference being in custom drop. 

### Checklist

### DIFF
--- a/scipio/src/sys/mod.rs
+++ b/scipio/src/sys/mod.rs
@@ -3,7 +3,7 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
-use std::cell::{RefCell, UnsafeCell};
+use std::cell::{Cell, RefCell};
 use std::ffi::CString;
 use std::mem::ManuallyDrop;
 use std::net::{Shutdown, TcpStream};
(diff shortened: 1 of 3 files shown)


scipio/src/sys/mod.rs
```diff
@@ -183,40 +183,46 @@ impl Wakers {
...
 
-    pub(crate) source_type: SourceType,
+    pub(crate) source_type: RefCell<SourceType>,
 
     io_requirements: IoRequirements,
 
-    id: Option<SourceId>,
+    id: Cell<Option<SourceId>>,
+
+    queue: Cell<Option<ReactorQueue>>,
```
matklad, line 200: @glommer this is a fun one -- you can use `Cell` with non-Copy types, it's a useful advanced tool to have in a toolbox.

glommer: which is surprising, especially given that `Option` is not `Clone`

glommer:
Agree that `mut_source` is pretty nasty.
In the early days of this code I didn't have the companion `source_map` at all, and relied just on the pointer being stored
in the `user_data` (which is what a respectable C++ person would do)

As the `source_map` came to be and we moved more towards indexes, this is the logical next step -> good work.

If I understand correctly, you are just moving the interior mutability pattern to the fields that really need it, and as `SourceMap` will translate between id and the rc pointer and that can be made safe.

matklad:
> If I understand correctly, you are just moving the interior mutability pattern to the fields that really need it, and as SourceMap will translate between id and the rc pointer and that can be made safe.

Indeed, this is just mechanical leveraging of interior mutability. Though, i use Cells somewhat more actively then RefCells, as they have a better API.

glommer:
Agree about `Cell` and I have been naturally using them more and more.
But I had no idea they could be used for complex, non-copy types.

matklad:
This is ready for review now. I am much happier with the code in terms of safety, but I don't quite like the overall shape of it.

glommer:
Thanks. I will take a look as soon.

What are you unhappy about?

matklad:
@glommer see the last commit in #131, which tries to cleanup things by introducing an explicit state machine for Source state

scipio/src/sys/uring.rs
```diff
@@ -831,7 +795,7 @@ impl Reactor {
     }
 
     pub(crate) fn statx(&self, source: &Source) {
-        let op = match source.source_type() {
+        let op = match &*source.source_type() {
```
glommer, line 798: I trust you to know what you're doing rust wise but I don't understand this. Would like to understand a bit better before pressing the button.

`source_type()` returns a `Ref` from the immutable borrow.
Aren't you getting a naked reference to the interior of the cell by doing `&*` ? I thought this was not possible?

But even if it is possible, wouldn't that live for just as long as the `Ref`?

matklad: Taking reference to the interior is possible, Ref implements Deref. Moreover, it is essential, otherwise you won‘t be able to call any methods at all on values behind RefCells.

The reference would indeed leave only as long as Ref, but this is enough in this case: rust has temporary lifetime extension rules, much like C++, but you don‘t generally need to understand them in details -- the worst case is that the code fails to compile, and you need to introduce a temporary.

Note that this is exactly the same API as for mutex guards. There, one also often uses &* incantation to summon a short-lived reference to a data protected by a guard.

glommer: Right, so what is not possible is to generate a reference that escapes the lifetime of the borrow.

Correct?
My first instinct looking at this code is that the lifetime of the borrow would be not the entire match statement but just the first part `match <this_part>`

But I agree that as far as lifetimes go, if it compiled it is correct.

matklad: Yes, that is correct.

I think all temporaries are live until the end of the full expression (the whole `match` In this case), unless the are extended further by something like `let x = &“hello“.to_string();`. But this is a part of Rust I don’t know and just rely on the compiler to tell me when I am wrong.

PR #128, opened 2020-10-19
rework of the DmaFile API

glommer:
### What does this PR do?

Changes the DmaWrite::write_dma API so it takes ownership of the buffer instead of taking a reference.

Also in preparation for the Buffered File API, rename `write_dma => write_at`, `read_dma => read_at` for maximum consistency.

### DIFF
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -179,7 +179,7 @@ where
                 sqe.prep_read(op.fd, buf.as_bytes_mut(), pos);
                 let src = peek_source(from_user_data(op.user_data));
 
-                if let SourceType::DmaRead(_, slot) = &mut *src.source_type.borrow_mut() {
+                if let SourceType::Read(_, slot) = &mut *src.source_type.borrow_mut() {
                     *slot = Some(buf);
                 } else {
                     panic!("Expected DmaRead source type");
@@ -769,10 +769,14 @@ impl Reactor {
         self.queue_standard_request(source, UringOpDescriptor::PollAdd(flags));
     }
 
-    pub(crate) fn write_dma(&self, source: &Source, buf: &DmaBuffer, pos: u64) {
-        //        let op = UringOpDescriptor::WriteFixed(buf.as_ptr() as *const u8, buf.len(), pos, buf.slabidx);
-        let op = UringOpDescriptor::WriteFixed(buf.as_ptr() as *const u8, buf.len(), pos, 0);
-        self.queue_storage_io_request(source, op);
+    pub(crate) fn write_dma(&self, source: &Source, pos: u64) {
+        match &*source.source_type() {
+            SourceType::Write(_, buf) => {
+                let op = UringOpDescriptor::WriteFixed(buf.as_ptr(), buf.len(), pos, 0);
+                self.queue_storage_io_request(source, op);
+            }
+            x => panic!("Unexpected source type for write: {:?}", x),
+        }
     }
 
     pub(crate) fn read_dma(&self, source: &Source, pos: u64, size: usize) {
@@ -913,7 +917,7 @@ impl Reactor {
 
     fn queue_storage_io_request(&self, source: &Source, op: UringOpDescriptor) {
         let pollable = match &*source.source_type() {
-            SourceType::DmaRead(p, _) | SourceType::DmaWrite(p) => *p,
+            SourceType::Read(p, _) | SourceType::Write(p, _) => *p,
             _ => panic!("SourceType should declare if it supports poll operations"),
         };
         match pollable {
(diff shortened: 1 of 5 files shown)


glommer:
I thought long and hard about this and decided to forego buffer reuse.

* It is dangerous in the wrong hands to begin with, because if you trim the buffer you cannot safely expand it again.
* it complicates the API and it becomes the only function not to return `io::Result`.

Avoiding allocations is important but it is a lot more important for reads where the buffers can be quite small. Small writes are very rare so let's keep it simple and safe.

scipio/src/sys/uring.rs
```diff
@@ -769,10 +769,15 @@ impl Reactor {
...
 
-    pub(crate) fn write_dma(&self, source: &Source, buf: &DmaBuffer, pos: u64) {
-        //        let op = UringOpDescriptor::WriteFixed(buf.as_ptr() as *const u8, buf.len(), pos, buf.slabidx);
-        let op = UringOpDescriptor::WriteFixed(buf.as_ptr() as *const u8, buf.len(), pos, 0);
-        self.queue_storage_io_request(source, op);
+    pub(crate) fn write_dma(&self, source: &Source, pos: u64) {
+        match &*source.source_type() {
+            SourceType::Write(_, buf) => {
+                let op =
+                    UringOpDescriptor::WriteFixed(buf.as_ptr() as *const u8, buf.len(), pos, 0);
```
matklad, line 776:
```suggestion
                    UringOpDescriptor::WriteFixed(buf.as_ptr(), buf.len(), pos, 0);
```

I think this is a no-op cast?

matklad:
LGTM, but I have a philosophical concern.

Am i correct that this fixes potential UAF? The same problem as in the initial buffered API: cancelling write without awaiting for cancellation would cause the compiler to think that the users Code is the sole owner of buffer, while in fact it is being read from by the kernel?

If this is indeed the case, I slightly worry that this PR *doesn‘t* use unsafe keyword in implementation. It would be better if our internal uring API was annatoted with unsafe in the right places, so that, when writing higher-level code like DmaFile, the unsafe on the call side stood out to mark „here be dragons, make sure not to release this buffer prematurely“ bits.

glommer:
Not sure I understand your concern - although my goal is not to add unsafe anywhere, but rather make this safe.

The buffer is being passed to `DmaFile` now, who takes ownership. So we are the owner of the buffer, not the user.
The buffer is written into the Source structure, and we keep the `Source` alive until completion or cancellation.

Where exactly do you see the UAF ?

glommer:
* Folded Aleksey's suggestion into the first patch and fixed the fmt breakage that resulted.
* patch description no longer mention buffer reuse, since we stopped doing that.

I am pressing the button, but happy to keep the safety discussion going

matklad:
To be clear, the current API is sound (well, at lest I don't see any obvious holes). The *previous* one was unsound (scipio's user could have written a safe code triggering UAF). 

What I am trying to say is this. We can have two approaches to safety in scipio.

*First*, we can guarantee only that scipio's public API is safe and sound, and internally do whatever, as long as user-visible surface is OK.

*Second*, we can designate a safety boundary *within* scipio (say, around the sys module), and require that all the unsafety lives inside this smaller boundary.

In the first case, potentially any code change in scipio can lead to undefined behavior. In the second case, only changes inside the smaller boundary or unsafe blocks need to be scrutinized for safety. 

Second case, if we pull it off, feels very attracive -- the less code to verify for safety the better! Moreover, it makes veryfing the safety of public API much easier as well -- if large public API is build using only safe code on top of small, but trusted unsafe boundary, we only need to very the small boundary, and the rest we get because safety is compositional. 

This PR demonstrates that we are in the first case though. This PR didn't change any `unsafe` blocks inside scipio, yet it changed the public API from being unsound to being sound. That means that there isn't safety boundary inside scipio, and every code change needs to be carefully examined for memory safety issues. We can't just apply the usual Rust wildecard of "I haven't touched unsafe blocks, so, while my code may have bugs, it doesn't have memory safety issues".

glommer:
👍 

I see what you mean.
I think we could definitely aim for the second option. It seems doable.
If you are not:
 - dealing with io_uring
 - dealing with buffers

I see no reason why we'd ever be unsafe. And also incidentally, all of that lives around sys already anyway. There are also syscalls: but first, they are inside sys too, and second they are temporary. We'll either wait for uring to provide them or implement it in the kernel ourselves...

We haven't started this way, so it begs the question of how much return of investment we get from doing it now. 

One thing that we could do, which is unrelated to unsafe, is just to require that everything that goes into the io_uring is `'static'. That certainly was not my original intention as I wanted to use references as much as possible, but I was not aware of the cancellation issue. Now that I am, we don't ever want to pass a reference anywhere...

PR #129, opened 2020-10-21
enhance local channel receiver with a recv function

glommer:
Right now the local channel receiver can only be consumed through a stream.

The stream interface itself is convenient, as it can be used to fetch a single
element with next().

However, next(), try_next() and friends in futures_lite take a mutable reference
to self.

This is not very convenient when we are dealing with immutable references.
Because the receiver implementation is already entirely done with interior
mutability, we can easily expose a recv() interface.

Although in some situations we could put the receiver itself in a Cell
or Refcell, this is really not convenient:

* It is a rule of safety (which we should soon enforce with clippy), that
  you must never .await with a borrow held. Because everything in the same
  executor is executing in a single CPU, it is always safe to borrow() or
  borrow_mut() -> unless it crosses an .await point. As the result of
  .next() has to be .awaited, that breaks.

* A cell won't work well since the LocalReceiver is not Copy, and also not
  long lived: if we were to .take() it from the Cell (which is apparently
  legal even for non-Copy types), we'd have to leave the Cell empty while
  we are consuming.

The good thing about the Deque implementation which we had was that
pop_front() was &self, not &mut self. In fact as I try to change my
application to use channels instead of the Deque, I hit this one particular
issue every time.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/channels/local_channel.rs
+++ b/scipio/src/channels/local_channel.rs
@@ -5,6 +5,7 @@
 //
 use crate::channels::ChannelCapacity;
 use futures_lite::stream::Stream;
+use futures_lite::FutureExt;
 use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::future::Future;
@@ -207,6 +208,10 @@ struct SendWaiter<T> {
     channel: LocalChannel<T>,
 }
 
+struct RecvWaiter<T> {
+    channel: LocalChannel<T>,
+}
+
 impl<T> Future for SendWaiter<T> {
     type Output = ();
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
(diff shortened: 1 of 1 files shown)


scipio/src/channels/local_channel.rs
```diff
@@ -207,6 +208,10 @@ struct SendWaiter<T> {
     channel: LocalChannel<T>,
 }
 
+struct RecvWaiter<T> {
+    channel: LocalChannel<T>,
+}
+
```
matklad, line 214: These waiter structs look a bit silly to me. I wonder if we should leverage `poll_fn` here?

It works for SendWaiter, but doesn't for RecvWaiter, becuse it is used twice:

https://github.com/matklad/scipio/compare/4e21fddab2303ab58ba35c3f8bd873f6aefeedc4..a97e2b498dfef67e78dd6d938d6731746c33758d

glommer: I've seen `poll_fn` before but wasn't terribly familiar with it.
I have read its docs again as a refresher and indeed seems like it could be used here.

I'll take another pass with this.

Thanks!

glommer:
Because both sender and receiver could use the poll_fn version, I decided to leave this patch untouched and add a cleanup pass that converts both as a patch on top.

I will not fold as I believe it will be clear this way.

PR #131, opened 2020-10-21
Strongly-typed state transition for source

matklad:
### What does this PR do?

Refactoring, encoding Source state transitions as an enum.

@glommer this currently fails the test, because `preempt_timer` functions actually do violate state transition assumptions I think are reasonable to make. Why these functions are "hand-codeed", instead of using the standard code-path of `queue_request_into_ring` and `consume_source`? Why do we need to track SourceId/user_data via `SourceType::Timeout.0` for timeouts, but not for other requests? 

Is there perhaps some high-level irregularity with this preempt_timer I am missing? 

### DIFF
--- a/scipio/src/sys/mod.rs
+++ b/scipio/src/sys/mod.rs
@@ -3,7 +3,7 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
-use std::cell::{Cell, RefCell};
+use std::cell::RefCell;
 use std::convert::TryFrom;
 use std::ffi::CString;
 use std::mem::ManuallyDrop;
@@ -185,54 +185,40 @@ impl From<Duration> for TimeSpec64 {
     }
 }
 
-/// Tasks interested in events on a source.
-#[derive(Debug)]
-pub(crate) struct Wakers {
-    /// Raw result of the operation.
-    pub(crate) result: Option<io::Result<usize>>,
-
-    /// Tasks waiting for the next event.
-    pub(crate) waiters: Vec<Waker>,
-}
-
-impl Wakers {
-    pub(crate) fn new() -> Self {
-        Wakers {
-            result: None,
-            waiters: Vec::new(),
-        }
-    }
-}
-
 /// A registered source of I/O events.
+#[derive(Debug)]
 pub struct InnerSource {
     /// Raw file descriptor on Unix platforms.
     raw: RawFd,
-
-    /// Tasks interested in events on this source.
-    wakers: RefCell<Wakers>,
 
     source_type: RefCell<SourceType>,
 
     io_requirements: IoRequirements,
 
-    enqueued: Cell<Option<EnqueuedSource>>,
+    state: RefCell<SourceState>,
 }
 
-pub struct EnqueuedSource {
-    pub(crate) id: SourceId,
-    pub(crate) queue: ReactorQueue,
+#[derive(Debug)]
+enum SourceState {
+    /// The `Source` starts in this state.
+    NotQueued,
+    /// The source was queued into the ring.
+    Queued(QueueInfo),
+    /// The source was completed with the specified result.
+    Ready {
+        /// Raw result of the operation.
+        result: io::Result<usize>,
+    },
+    /// The source finishes in this state after the result is consumed.
+    Consumed,
 }
 
-impl fmt::Debug for InnerSource {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("InnerSource")
-            .field("raw", &self.raw)
-            .field("wakers", &self.wakers)
-            .field("source_type", &self.source_type)
-            .field("io_requirements", &self.io_requirements)
-            .finish()
-    }
+#[derive(Debug)]
+struct QueueInfo {
+    id: SourceId,
+    queue: ReactorQueue,
+    /// Tasks waiting for the next event.
+    waiters: Vec<Waker>,
 }
 
 #[derive(Debug)]
@@ -246,10 +232,9 @@ impl Source {
         Source {
             inner: Rc::new(InnerSource {
                 raw,
-                wakers: RefCell::new(Wakers::new()),
                 source_type: RefCell::new(source_type),
                 io_requirements: ioreq,
-                enqueued: Cell::new(None),
+                state: RefCell::new(SourceState::NotQueued),
             }),
         }
     }
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -8,12 +8,12 @@ use rlimit::Resource;
 use std::cell::{Cell, Ref, RefCell};
 use std::collections::VecDeque;
 use std::ffi::CStr;
-use std::io;
 use std::io::{Error, ErrorKind};
 use std::os::unix::io::RawFd;
 use std::rc::Rc;
 use std::task::Waker;
 use std::time::Duration;
+use std::{io, mem};
 
 use crate::free_list::{FreeList, Idx};
 use crate::sys::posix_buffers::PosixDmaBuffer;
@@ -22,7 +22,7 @@ use crate::{IoRequirements, Latency};
 
 use uring_sys::IoRingOp;
 
-use super::{EnqueuedSource, TimeSpec64};
+use super::{QueueInfo, SourceState, TimeSpec64};
 
 type DmaBuffer = PosixDmaBuffer;
 
@@ -220,16 +220,18 @@ where
             return Some(());
         }
 
-        let src = consume_source(from_user_data(value.user_data()));
+        let (src, mut queue_info) = consume_source(from_user_data(value.user_data()));
 
         let result = value.result();
         let was_cancelled =
             matches!(&result, Err(err) if err.raw_os_error() == Some(libc::ECANCELED));
 
         if !was_cancelled && try_process(&*src).is_none() {
-            let mut w = src.wakers.borrow_mut();
-            w.result = Some(result);
-            wakers.append(&mut w.waiters);
+            let new_state = SourceState::Ready { result };
+            let old_state = src.transition(new_state);
+            assert!(matches!(old_state, SourceState::NotQueued));
+
+            wakers.append(&mut queue_info.waiters);
         }
         return Some(());
     }
@@ -251,10 +253,15 @@ fn add_source(source: &Source, queue: ReactorQueue) -> SourceId {
     SOURCE_MAP.with(|x| {
         let item = source.inner.clone();
         let id = x.borrow_mut().alloc(item);
-        source.inner.enqueued.set(Some(EnqueuedSource {
+
+        let new_state = SourceState::Queued(QueueInfo {
             id,
-            queue: queue.clone(),
-        }));
+            queue,
+            waiters: Vec::new(),
+        });
+        let old_state = source.inner.transition(new_state);
+        assert!(matches!(old_state, SourceState::NotQueued));
+
         id
     })
 }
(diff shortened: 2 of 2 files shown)


glommer:
Historically speaking, keep in mind that early on we didn't have indexes because my heart was still consumed by C++ish thoughts. The user_data for the io_uring structure would take the address of a source structure and it was very helpful to have a stable address for the preempt timer. Now with the indexes we could definitely have stable indexes (which we do) and relax a bit some of the other restrictions.

But "why do we need stable indexes?" is an interesting question:

* The way the preempt timer works is that it has to be reset every time we are going to go back to the run queues to run tasks. Before we leave `wait`, we must make sure that the preempt timer is not only queued but actually sent to io_uring.
However notice that the previous preempt timer may not have fired yet:
 - let's say you have set the preempt timer for 1ms. Then you ran tasks for 400us and is back to `wait`. You don't want to give the next batch of tasks just 600us to run. That would be unfair! So you need to cancel the existing preempt timer, and rearm it for 1ms again (or whatever the `preempt_timer_duration()` happens to be now, it is dynamically calculated).  You also can't just rearm a new one and let the old one fire and ignore it, because by firing it will take you off the CPU...

* When we go to sleep, however, we must not have the preempt timer set. We must cancel it and reap the results (otherwise we'll wake up from sleep when it fires).

Because of the above, we need to keep track of the preempt timer state at all times. If it fired already, I have to arm it. But if it didn't fire, I have to cancel it and rearm it. If we are going to sleep and it is fired, nothing to do. If we are going to sleep and it is armed we need to cancel and reap.

So you need a state machine of sorts for the preempt timer itself. It is undoubtedly easier to do so with a stable preempt timer address/index.

Is it the only way? No. If you manage to do it without complicating the preempt timer logic too much it's very welcome.

But notice from your own experience how easy the preempt timer does break! Also a lot of breakages are hard to catch because they only affect for how long a task queue runs, but wouldn't cause any crash or etc.

matklad:
Hm, interesting! Can we actually re-arm existing timer in place? The timespec argument in iouring is a pointer, is it OK that we mutate data pointed by it from the user-space, while the kernel has another op using it queued (with the cancellation queued as well)?  I would think that this should be fine because the kernel most probably copies the data from that pointer when we submit events (and we have a guarantee that the previous event is submitted). But I am also not completely convinced that this doesn’t suffer from the problem, similar to the one I’ve fixed in timers recently.

An alternative approach would be to cancel old timer and arm a new one.

glommer:
Best of my knowledge we cannot rearm the timer in place.
The kernel does copy the timespec so although it is passed a pointer, the pointer only serves to tell the kernel where to copy from (Because it is a complex type)

We can mutate our source structure just fine, but it won't do anything until we cancel and submit another one.

matklad:
> Best of my knowledge we cannot rearm the timer in place.

Did you intend to use „cannot“ here? The rest of the message seems like an explanation for why we *can* do it.

glommer:
Humm, no? Maybe we mean different things by "rearm in place".

I understood you meant "by just changing the value under that pointer, the timer will be rearmed". That is not the case.

And I can't see how we would rearm any other way:
There are only two io_uring timer operations: add and remove.
There is no rearm.

We could, as I said in my first explanation, arm a new timer and leave the old one in place. But we don't want to do that. Once the old timer fires it would take us out of the CPU.

matklad:
Seems like we indeed mean different things. I guess I could better rephrase it better: 

Can we re-use the same source to add a new timer, before we await for the cancellation of the old one?

This is what the current code does, and I wonder if it is ok. Based on 

> The kernel does copy the timespec so although it is passed a pointer, the pointer only serves to tell the kernel where to copy from (Because it is a complex type)

This should be ok.

matklad:
> Historically speaking, keep in mind that early on we didn't have indexes because my heart was still consumed by C++ish thoughts. 

Unrelated to the current PR, but one interesting alternative setup of the data structures would be to replace RCs with indexes altogether. The Source could contain *just* a SourceId, and lookupnthe source in a SourceMap for all relevant operations. I generally would prefer that architecture, if not for the problem that rusts thread locals are slow: https://matklad.github.io/2020/10/03/fast-thread-locals-in-rust.html

glommer:
Right, we can use the same source. We don't have to, but this has so far made dealing with the rearming easier.

glommer:
Right, indeed working with TLS in rust has been a bit disappointing.

If you see the comments in `need_preempt()` at parking.rs, definitely losing some cycles there that are very precious.

But I fail to see how using indexes is related to TLS. The reason we use the sourcemap is that we may have to manipulate those operations from outside the reactor, in places where a reactor context is hard to obtain.

matklad:
> But I fail to see how using indexes is related to TLS

Right, they are not inherently dependent on each other, if we can thread a reference to reactor everywhere, we won’t need TLS. Philosophically speaking, it would be cool if reactor contained a bunch of vectors for various resources, instead of using Rc. That way, answering „how much of each thing here is ?“ becomes trivial, and we are also immune to Rc cycles. But this just excited hand-waving at this point.

matklad:
I guess this is ready for review.

I am not completely convinced that this is a good idea.

On the one hand, this is a text-book example of the right way to do it: we replace C++-style default-initialized fields which not always make sense with a well-typed enum, which specifies exactly which data is and is not meaningful at any given point in the Source's lifetime

On the other hand, the code for dealing with states and adding required asserts is more verbose and un-natural, Particularly,  `Source::drop` now has to do a funny thing with early `drop` of `cell::Ref`, and `Source::take` is a bit roundabout (aside: I think we should replace take_result/add_waiter pair with a single `fn poll(&self, cx) -> Poll<io::Result>` which does both). 

On the third hand, these changes motivated #137, and this is a perfect example how low-level encoding of state machines in enums helps to uncover surprises in higher-level operations.

glommer:
Have you been around leaky nuclear power plants recently?
Most of us have only two hands!

I'll take a look very soon

matklad:
Now, just used to use Emacs way to much. That's why I've switched to VS Code :)

scipio/src/sys/mod.rs
```diff
@@ -185,54 +185,38 @@ impl From<Duration> for TimeSpec64 {
...
-    enqueued: Cell<Option<EnqueuedSource>>,
+    state: RefCell<SourceState>,
 }
 
-pub struct EnqueuedSource {
-    pub(crate) id: SourceId,
-    pub(crate) queue: ReactorQueue,
+#[derive(Debug)]
+enum SourceState {
+    /// Neutral state, the Source either wasn't submitted yet, or was fully processed.
```
glommer, line 203: if it was fully processed, wouldn't it have a result ?

glommer:
Sorry to hear about the Emacs thing. Nobody should have to go through that.

In the past people have told me I am prolific. Little do they know that my secret is: I opened a vi session 15 years ago and I wasn't able to exit it yet. So I end up just coding to see if I can exit the damn thing by some lucky strike.

As to your code, the state machine looks good to me. So in terms of direction ack for that.

I want to read this a couple of times more to convince myself this is all correct, especially around the cancellation of a non-queued request. Just because this is tricky...

scipio/src/sys/mod.rs
glommer: I see now that you move the Source back to this state when the result is taken.

I am a little bit on the fence on that one: it clearly works, the question is: are there any situations in which we want to check if a source was already processed?

One alternative way of doing this then, which would make the transitions clear is to have a 4th state:  `Consumed`
takine the result would move it to that state.

It is one more state, but at least the meaning of no state is overloaded.

matklad: Yeah, this makes sense -- if we are doing the right thing, might do it the right way as well, amended the commit.

glommer (approved):
Very good.

Cancelling consumes the request and that symmetry allows us to treat a cancelled request as consumed without an extra state.

I happen to think that when things end up lining up like this in unexpected ways, that's the hallmark of a good design!

glommer:
Heads up: I am now seeing a panic in my application that I traced down to this series.
Reverting it makes it work again.

It crashes with
```
thread 'unnamed-0' panicked at 'Attempting to get a result from Source, which wasn't queued', /home/glauber/scipio/scipio/src/sys/uring.rs:542:17
```

That likely means we got some of the transitions wrong. I am investigating.

glommer:
It seems to come from networking, for which we have very little testing since we got that from async-io and haven't refactored yet:

```
  11: std::panicking::begin_panic
             at /rustc/04488afe34512aa4c33566eb16d8c912a3ae04f9/src/libstd/panicking.rs:456
  12: scipio::sys::uring::<impl scipio::sys::Source>::take_result
             at /home/glauber/scipio/scipio/src/sys/uring.rs:542
  13: scipio::parking::<impl scipio::sys::Source>::readable::{{closure}}::{{closure}}
             at /home/glauber/scipio/scipio/src/parking.rs:478
  14: <futures_lite::future::PollFn<F> as core::future::future::Future>::poll
             at /home/glauber/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-lite-1.11.2/src/future.rs:246
  15: scipio::parking::<impl scipio::sys::Source>::readable::{{closure}}
             at /home/glauber/scipio/scipio/src/parking.rs:477
  16: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/04488afe34512aa4c33566eb16d8c912a3ae04f9/src/libcore/future/mod.rs:78
  17: scipio::pollable::Async<T>::readable::{{closure}}
             at /home/glauber/scipio/scipio/src/pollable.rs:183
  18: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/04488afe34512aa4c33566eb16d8c912a3ae04f9/src/libcore/future/mod.rs:78
  19: scipio::pollable::optimistic::{{closure}}::{{closure}}
             at /home/glauber/scipio/scipio/src/pollable.rs:469
  20: <futures_lite::future::PollFn<F> as core::future::future::Future>::poll
             at /home/glauber/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-lite-1.11.2/src/future.rs:246
  21: scipio::pollable::optimistic::{{closure}}
             at /home/glauber/scipio/scipio/src/pollable.rs:466
  22: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/04488afe34512aa4c33566eb16d8c912a3ae04f9/src/libcore/future/mod.rs:78
  23: scipio::pollable::Async<T>::read_with::{{closure}}
             at /home/glauber/scipio/scipio/src/pollable.rs:237
  24: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/04488afe34512aa4c33566eb16d8c912a3ae04f9/src/libcore/future/mod.rs:78
  25: scipio::networking::<impl scipio::pollable::Async<std::sys::unix::ext::net::UnixListener>>::accept::{{closure}}
             at /home/glauber/scipio/scipio/src/networking.rs:380
```

glommer:
indeed the implementation of `readable()` explicitly tries to take a result from a source that is not ready:

```
    /// Waits until the I/O source is readable.
    pub(crate) async fn readable(&self) -> io::Result<()> {
        future::poll_fn(|cx| {
            if self.take_result().is_some() {
                return Poll::Ready(Ok(()));
            }

            self.add_waiter(cx.waker().clone());
            Reactor::get().sys.interest(self, true, false);
            Poll::Pending
        })
        .await
    }
```

Under our new state machine that is illegal. The reason is that pollable_fds reuse the same source and just register interest in a file descriptor every time it wants to read / write something.

Long term pollable fds are not very interesting: I want to use uring's native networking here. But for now I think the simplest fix is to just allow one to query for a source that is not queued. The other alternative is to refactor and reachitect the pollable fd code not to do it. Getting rid of if in favor of uring's native networking is time better spent.

glommer:
@matklad I am reverting this for now. 

BTW, one of our examples reproduces this:
```
cargo run --example echo
```

Once I'm done with some approaching deadlines on my end I should be able to take more care of this and revamp the networking code not to use pollable fds =(

For now I'll leave it up to you whether you think there is a minimum amount of work that can be done with pollable fds to respect the state machine or if you want to relax it.

matklad:
>indeed the implementation of readable() explicitly tries to take a result from a source that is not ready:

Ah, I should have seen this coming!

Originally, I removed `take_result` and `add_waiter` functions altogether, and replaced them with

```rust
impl Source {
  fn poll(&self, cx) -> Poll<io::Result>
}
```

which protects exactly from this problem. But, when I was rebasing the work on top of master, I've noticed several new usages of `take_result` / `add_waiter` (in the buffered file sutff?) and decided to revert this change to not inflate the diff.

matklad:
I guess, the simplest hot-fix would be to change `take_result` to be more forgiving about the current state. But now I am much more convinced that `poll` is the right API here.

matklad:
I might be able to look at this tomorrow!

glommer:
the pollable fd support is a bit alien.
It's good to have, because technically one can still use poll for other stuff except networking.

glommer:
there's no need for a hotfix, the hotfix is to revert it =)
(https://github.com/DataDog/scipio/pull/146)

I'd rather do that and you take your time. No need to rush this

PR #137, opened 2020-10-23
Simplify handling of timeouts

matklad:
### What does this PR do?

Leverage `Source::drop` for cancellation of ring's timer.

### Motivation

Mainly to get rid of "single source is present twice in the SourceMap" condition, but also just to reduce the amount of code

### Checklist

### DIFF
--- a/scipio/src/sys/uring.rs
+++ b/scipio/src/sys/uring.rs
@@ -554,15 +554,11 @@ impl SleepableRing {
         let source = Source::new(
             IoRequirements::default(),
             -1,
-            SourceType::Timeout(None, TimeSpec64::default()),
+            SourceType::Timeout(TimeSpec64::from(d)),
         );
         let new_id = add_source(&source, self.submission_queue.clone());
-        source.update_source_type(SourceType::Timeout(
-            Some(to_user_data(new_id)),
-            TimeSpec64::from(d),
-        ));
         let op = match &*source.source_type() {
-            SourceType::Timeout(_, ts) => UringOpDescriptor::Timeout(&ts.raw as *const _),
+            SourceType::Timeout(ts) => UringOpDescriptor::Timeout(&ts.raw as *const _),
             _ => unreachable!(),
         };
 
@@ -634,14 +630,11 @@ impl UringCommon for SleepableRing {
                 SourceType::LinkRings(LinkStatus::Freestanding) => {
                     panic!("Impossible to have an event firing like this");
                 }
-                SourceType::Timeout(id @ Some(_), _) => {
-                    *id = None;
-                    Some(())
-                }
-                // This is actually possible: when the request is cancelled
-                // the original source does complete, and the cancellation
-                // would have marked us as false. Just ignore it.
-                SourceType::Timeout(None, _) => None,
+                SourceType::Timeout(_) => Some(()),
+                // // This is actually possible: when the request is cancelled
+                // // the original source does complete, and the cancellation
+                // // would have marked us as false. Just ignore it.
+                // SourceType::Timeout(None, _) => None,
                 _ => None,
             },
             wakers,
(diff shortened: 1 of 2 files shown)


scipio/src/sys/uring.rs
```diff
@@ -891,10 +872,10 @@ impl Reactor {
         let mut should_sleep = match timeout {
             None => true,
             Some(dur) => {
-                let mut src = self.timeout_src.borrow_mut();
-                lat_ring.cancel_preempt_timer(&mut src);
+                // Cancel the old timer.
+                self.timeout_src.take();
```
glommer, line 869: Does Rust guarantee that the timer will be dropped at exactly this point and not later?

For correctness the timer must be gone before we call `flush_cancellations` below. If we don't do that we may end up racing and cancelling the new timer we just installed (if they use the same id), or having a spurious wake up if they don't.

My biggest concern when I started reading your patch was "what about flushing the cancellations?" 
But you kept it in a good place. If we cancel here then I think your change is good.

scipio/src/sys/uring.rs
```diff
@@ -912,8 +893,8 @@ impl Reactor {
             // We are about to go to sleep. It's ok to sleep, but if there
             // is a timer set, we need to make sure we wake up to handle it.
             if let Some(dur) = timer_expiration {
-                let mut src = self.timeout_src.borrow_mut();
-                lat_ring.cancel_preempt_timer(&mut src);
```
glommer, line 916: I think the only comment that I have is that it would be nice to force flush_cancellations to be called when we drop the timer so we don't forget to pair them. I guess it is not, because you need a wakers pointer which is different all the time, plus a reference to the ring.

glommer (approved):

scipio/src/sys/uring.rs
```diff
@@ -634,14 +630,11 @@ impl UringCommon for SleepableRing {
...
                 }
-                SourceType::Timeout(id @ Some(_), _) => {
-                    *id = None;
-                    Some(())
-                }
-                // This is actually possible: when the request is cancelled
-                // the original source does complete, and the cancellation
-                // would have marked us as false. Just ignore it.
-                SourceType::Timeout(None, _) => None,
+                SourceType::Timeout(_) => Some(()),
```
glommer, line 633: As we no longer have this case, as a cleanup, we should remove the comments

scipio/src/sys/uring.rs
matklad: Yes, rust does guarantee this. We can over-emphasize this by writing `drop(self.timeout_src.take())`, but I am not a fan of spelling `drop` just for emphasis.  Couple of related things:

* Drop order in Rust is not influenced by lifetimes. The (implicit) drops are inserted before the  borrow checker is run, the checker then just verifies that it is possible to work-out reasonable lifetimes with the given drops.
* `let _ = foo();` will drop the result of `foo()` immediately, `let _foo = foo();`  will keep it around until the end of the block. 
* I am not a fan of explicit `drop`, because it may change the drop order, but only on the happy path:

```rust
let a = foo();
let b = bar()

spam()?;

drop(a)
drop(b)
```

In this case, the drop order would be `a, b`  if `spam` returns Ok and `b, a` if it Errs. I've once hit a deadlock on program termination for exactly this reason. Since then, I just use extra `{}` and order local variables such that implicit drop order is the correct one.

scipio/src/sys/uring.rs
matklad: Oups totally meant to remove this, sent #138

scipio/src/sys/uring.rs
matklad: Yeah, I considered extracting this into a `self.cancel_timeout` function, but decided against it -- we only need twice, and both usages fit into the single screen. Addition a function call abstraction here would make it more difficult to understand what actually is going on here, although, tbh, I don't understand what's going on in details here anyway, especially why we need two `Option<Duration>` arguments

glommer: @sitano should be able to explain to you, since I explained to him in his draft the other day in depth =) 
https://github.com/DataDog/scipio/pull/106

In summary one of them is the preempt timer. The other represents user timers.
We need to deal with them differently so I figured it was best to have two parameters rather than overloading one.
(They did evolve from a single parameter which is why the naming could be better, I think)

In particular, we must never sleep with the preempt timer armed. But we always sleep with the user timer (if there is any) armed.

I'll add comments and change the names a bit. This code evolved organically and could use some love by me too

sitano: yes, that was clear to me.

glommer: @matklad https://github.com/DataDog/scipio/pull/139

PR #142, opened 2020-10-24
Dma writes are broken if file is bigger than write_behind * buffer_size

### LINKED ISSUE #141: Writes failed if file is bigger than Buffer_size * write_behind
The commit that changed the write api (f494c5078ddb639d4841555df7c3946d96a1cd5a) has a bug.

The unit tests don't catch it because the buffer size may be aligned up due to Direct I/O requirements, so turns out we have no test currently that writes more than the write behind.

glommer:
This is because we add elements to the pending list, but never take them
out. This code worked before because we kept a fixed size list of buffers
and gave buffers back. Pending had no role in determining whether or not
we could write.

The worst problem, though, was that the unit tests that should cover this
didn't: because the buffer size is aligned to the Direct I/O alignment
we don't really write with the buffer size we requested so all tests
ended up being too small to trigger this.

Fixes #141

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/io/dma_file_stream.rs
+++ b/scipio/src/io/dma_file_stream.rs
@@ -675,9 +675,10 @@ struct DmaStreamWriterState {
     waker: Option<Waker>,
     file_status: FileStatus,
     error: Option<io::Error>,
-    pending: Vec<task::JoinHandle<(), ()>>,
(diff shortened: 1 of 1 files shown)


glommer:
I am fast tracking this because large file writing are essentially broken

PR #143, opened 2020-10-25
Remove no-op cancellations

matklad:
### What does this PR do?

`JoinHandle`s implement structured concurrency and already do
cancellation in drop, so there's no need for explicit cancellations.

### Additional Notes

One interesting thing I've noticed is that both `.cancel` and `.drop` are sync. Ideally, you'd `.await` for task after its cancellation. I am not sure how this should interact with scipio:

* on the one hand, because executor runs on a single thread, I *think* we can implement fully synchronous task cancellation, such that `cancel` immediately frees all task's resources. 
* on the other hand, the cancellation *should* be fully async. Let's say we have a task, which is waiting on an uring operation. Currently, when we cancel the task, we just enqueue cancellation operation into uring and, IIUC, we never actually await for the cancellation to happen. So, if you allocate 10GB buffer to do a read, and then cancel the read, that task will return from `.await` reporting that the cancellation concluded, but there still be 10GB buffer in the SourceMap for some time. 

### Checklist

### DIFF
--- a/scipio/src/io/dma_file_stream.rs
+++ b/scipio/src/io/dma_file_stream.rs
@@ -138,9 +138,7 @@ struct DmaStreamReaderState {
 impl DmaStreamReaderState {
     fn discard_buffer(&mut self, id: u64) {
         self.buffermap.remove(&id);
-        if let Some(handle) = self.pending.remove(&id) {
-            handle.cancel();
-        }
+        self.pending.remove(&id);
     }
 
     fn replenish_read_ahead(&mut self, state: Rc<RefCell<Self>>, file: Rc<DmaFile>) {
@@ -236,14 +234,6 @@ impl DmaStreamReaderState {
             handles.push(v);
         }
         handles
-    }
-}
-
-impl Drop for DmaStreamReaderState {
-    fn drop(&mut self) {
-        for (_k, v) in self.pending.drain() {
-            v.cancel();
-        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


matklad:
Wait, we detach the tasks, so this is wrong...

matklad:
Yeah, so this PR is clearly wrong, because we are detaching the tasks. But I wonder *why* do we need to detach tasks here? Without having read the code to closely, it seems like tasks are scoped to `Stream`s, so it seems like a perfect use-case for drop-based emergency resource cleanup?

glommer:
The whole purpose of the write-behind / read-ahead mechanism is that you will keep many such requests in flight.
They come and go as they please and you never wait for all of them together.

My understanding is that if I don't `detach` those tasks, I will have to `.await` them right away because rust futures are lazy and then it becomes quite hard to achieve this level of parallelism.

I could be wrong in that understanding, but in fact removing the `.detach` the compiler tells me this:

```
   Compiling scipio v0.1.0 (/home/glauber/scipio/scipio)
warning: unused `executor::Task` that must be used
   --> scipio/src/io/dma_file_stream.rs:714:9
    |
714 | /         Local::local(async move {
715 | |             defer! {
716 | |                 waker.wake();
717 | |             }
...   |
745 | |             drop(state);
746 | |         });
    | |___________^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: tasks get canceled when dropped, use `.detach()` to run them in the background
```

glommer:
@matklad are you still pursuing this after my explanation ?

matklad:
Not at the moment: life happened :) I still think that there's a fair chance that non-detaching would work, but let's close this PR for now, as I don't have an estimate for when I can get back to it!

PR #147, opened 2020-10-29
fix semaphore permit

glommer:
Reported by Andrey Lomakin on our zulip channel today:

The permits are broken and what is even more broken is the fact that
the unit tests are not catching it because they release immediately
(let _)

A couple of fixes for the semaphore are tied together here:
* we really should, for safety, drop the borrow before wake() in case
  wake runs immediately. We are not doing that.
* we add a must_use to the permit to prevent this
* While fixing the above I notice that this implementation can generate
  wake storms, so we wake only as many wakers as would wake
  (virtual consumed is used for that). Since this is already a queue
  we can (and should) remove the is_empty() test on the list. This test
  was there for fairness so it is moved to the acquire() implementation

### Checklist

### DIFF
--- a/scipio/src/semaphore.rs
+++ b/scipio/src/semaphore.rs
@@ -42,6 +42,7 @@ impl Future for Waiter {
 struct State {
     idgen: u64,
     avail: u64,
+    virtual_consumed: u64,
     waiterset: HashMap<WaiterId, (u64, Waker)>,
     list: VecDeque<WaiterId>,
     closed: bool,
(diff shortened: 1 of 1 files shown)


laa:
Hi @glommer this is Andrey Lomakin, thank you for the fix, looks really good now. I am pleased that you have taken all my observations into account. :-)

glommer:
it's my pleasure! I was just waiting for you to validate the fix and I will publish a new version on crates with it.

Thanks

PR #149, opened 2020-10-29
shared channels: spsc connection between executors

glommer:
Sometimes we need to pass data between executors. The shared channel
will be used for that.

To implement the Shared Channel I am using the spsc queue crate
from https://github.com/polyfractal/bounded-spsc-queue. I like the implementation
and it is plenty fast. There are some things that we need to change there
(like the fact that it depends on the unstable compiler) and I sent patches.

We'll see how responsive they are but worst case the crate is simple
enough that I can copy.

The idea behind it is that the types going through the channel need to
be Send. But in order to probe for whether or not there is space in
the channel and activate the wakers we need to drill down the reactor
and check on every poll. Because of that, we require a sender or
receiver to be connected with a connect() function.

The connect() function consumes self, the inner state is not accessible
to the public, and before a connection is made there are no methods
available in either the sender or receiver: this makes it safe to force
the Send trait onto the Shared Receiver/Producer even though their
state is implemented through an Rc.

The Rc is important for two reasons: first it allow us to pass the state
deep down to the reactor so we can check if the wakers can be woken up.

But secondly, it allows for many asynchronous, detached tasks to push
data into this channel inside a single executor.

Like the local channel, I am implementing a stream interface for the
receiver. However currently there is no good way to signal a disconnect
to the remote end so it is not very useful. Performance-wise, the best
way to implement this is in the actual spsc buffer so I am waiting to see
what will happen with that code before I add more into these interfaces.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -20,3 +20,4 @@ lazy-static,https://crates.io/crates/lazy_static,MIT/Apache-2.0,Marvin Löbel
 scopeguard,https://crates.io/crates/scopeguard,MIT/Apache-2.0,bluss
 pin-project-lite:https://crates.io/crates/pin-project-lite,MIT/Apache-2.0,Taiki Endo
 log,https://crates.io/crates/log,MIT/Apache-2.0/The Rust Project Developers
+bounded-spsc-queue:https://github.com/polyfractal/bounded-spsc-queue.git,Apache-2.0,Zachary Tong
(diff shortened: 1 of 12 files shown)


glommer:
@Daniel-B-Smith this is a WIP. But I believe the interface should do the job for you.
This is a single-producer-single-consumer channel that can be used between any two executors.

The interface should be safe enough as to prevent anything that tries to use multiple producers or consumers from compiling.

PR #150, opened 2020-11-02
freelist: add values() iterator

glommer:
There are some use cases for a freelist where we need to iterate over all
values. Same as for the HashMap, add a values() iterator.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/scipio/src/free_list.rs
+++ b/scipio/src/free_list.rs
@@ -54,6 +54,10 @@ impl<T> Default for FreeList<T> {
     }
 }
 
+pub(crate) struct Values<'a, T> {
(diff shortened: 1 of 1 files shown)


glommer:
I am not sure I have a use for this, though, so we may defer until we find one.

I am now scanning the users in the `dma_file_stream` file and they are also a bit more complex: the freelist would have to host a task that itself needs to know its id -- so it can deallocate itself when it is done.

glommer:
This is a bit less important now that we are using `aHash` (or whichever other fast implementation we end up using)

PR #159, opened 2020-11-06
Update README with link to glommio docs.rs

pmalek-sumo:
### What does this PR do?

Updates link in README to point to glommio instead of scipio at docs.rs


glommer:
Hi! Thanks!

I was waiting for the zulip situation to settle before merging them both together, but I see no reason not to merge this as is now.

glommer (approved):

PR #166, opened 2020-11-12
Implementation of ReadWriteLock.

laa:
### What does this PR do?

Implementation of read-write lock similar to std::sync::RwLock. 

### Motivation

Sometimes users need to wait till the execution of async IO will be finished keeping invariants of the shared state.

In my personal case, I can not close component till data will not be flushed out from the case and related metadata which reflects the position of the data on the disk will not be updated correctly.

### Related issues

We discussed this feature in Zulip chat.

### Additional Notes

1. Lock is implemented taking into account that wakers will produce delayed execution of "fibers". It looks like quite a healthy assumption for me, but if you against such assumption I will fix code.
2. Lock is not re-entrant but I plan to provide subsequent changes to make it re-entrant at least for the `Task`s.
3. There are several optimizations to wake up only "fibers" which can proceed were implemented.
4. Additional `rand` dependency was added in dev-dependencies for testing purpose only.

### Checklist


glommer:
Reviewing.

glommio/src/lib.rs
```diff
@@ -364,6 +364,7 @@ pub mod io;
 mod multitask;
 mod networking;
 mod pollable;
+pub mod read_write_lock;
```
glommer, line 367: That means users will import `glommio::read_write_lock::ReadWriteLock` right?

That's suboptimal, I think. It is best if modules are a collected of related things.
There are two options:
1. Don't use a module, and just export `ReadWriteLock` as we do with the `Semaphore`.
2. Put this, together with the `Semaphore` in their own module. People in the past suggested `sync` for this. Now that we have a `Semaphore` and a rwlock it makes sense. This breaks API but I won't be too concerned with that until we reach 1.0. Best to get things right!

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
+// MIT/Apache-2.0 License, at your convenience
+
+//! Read-write locks.
+//!
+//! Provides functionality similar to the ['std::sync::RwLock'] except that lock can not be poisoned.
```
glommer, line 6: Should we call it `RwLock` for consistency ?

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+        });
+    }
+
+    #[test]
+    fn test_close_wr() {
+        test_executor!(async move {
+            let rc = Rc::new(ReadWriteLock::new(1));
+            let rc2 = rc.clone();
+
+            Task::local(async move {
```
glommer, line 888: If I read this correctly this test acquires the lock in write mode a task queue and tries to read it from another (the default one).

While this works, it happens to work due to an implementation detail that can change. Nothing guarantees which task queue runs first. And as a matter of fact I am surprised this works. I would expect the default tq to run first.

Where I need to test things like this I have been relying on explicit serialization: Essentially add a control variable inside a `Cell`, and a yielding loop `while status.get() == x { Local::later().await; }`

For an example of that you can look at the `permit_raii_works` in the semaphore tests.

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+        });
+    }
+
+    #[test]
+    fn test_close_ww() {
+        test_executor!(async move {
+            let rc = Rc::new(ReadWriteLock::new(1));
+            let rc2 = rc.clone();
+
+            Task::local(async move {
```
glommer, line 904: same issue as above, although this one could be solved in a different manner.
(one of them should fail, doesn't matter which)

However it is good to have the same logic everywhere. We can come up with a general test macro for this if it helps.

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+            let ball2 = ball.clone();
+            let ball3 = ball.clone();
+
+            let mut fibers = Vec::new();
+
+            let pinger = Task::local(async move {
+                let mut prev = -1;
+                loop {
+                    //give a room for other fibers to participate
+                    Task::<()>::later().await;
```
glommer, line 1045: `Local` is a bit more idiomatic than `Task::<()>`

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+    pub fn get_mut(&mut self) -> ReadWriteLockResult<&mut T> {
+        let state = (*self.rw).borrow();
+        if state.closed {
+            return Err(LockClosedError);
+        }
+
+        Ok(unsafe { &mut *(state.value.get()) })
+    }
+
+    /// Locks this ReadWriteLock with shared read access, blocking the current fiber
```
glommer, line 439: suggested saying "suspending" instead of blocking, as "blocking" has another meaning which is very loaded for us.

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+    /// until it can be acquired.
+    ///
+    /// The calling fiber will be blocked until there are no more writers which
```
glommer, line 442: suspended here too

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+
+impl Debug for LockClosedError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.write_str("Closed")
+    }
+}
+
+///Error which indicates reason of the error returned by [`try_read`] and ['try_write'] methods
+///on ['ReadWriteLock']
+pub enum TryReadWriteLockError {
```
glommer, line 82: what's the advantage of using our own error class?
For the channels I have used standard `io::Error`. A closed lock can return `BrokenPipe` and also `WouldBlock`.
The advantage of using `io::Error` is that we can just convert easily between errors coming from different places of the system.

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+impl<'a, T> Deref for ReadWriteLockReadGuard<'a, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        let state = (*self.rw).borrow();
+        if state.closed {
+            panic!("Related ReadWriteLock is already closed");
+        }
+
+        let ptr = state.value.get();
```
glommer, line 314: do we really need to use an `UnsafeCell` here? I am wondering if we couldn't use a `RefCell` and move the `RefMut`inside the `ReadWriteLockxxGuard`.

I don't feel too strongly, but @matklad has been cracking down hard on our `UnsafeCell`!

glommer (changes requested):
Very good overall!!

A couple of comments especially in testing, where I think (if I understood your tests correctly) that we need to do a slightly better job in future proofing their ordering.

glommio/Cargo.toml
```diff
@@ -34,6 +34,7 @@ pin-project-lite = "0.1.10"
 
 [dev-dependencies]
 futures = "0.3.5"
+rand = "0.7.3"
```
matklad, line 37: FWIW, I prefer `fast_rand` or `oorandom` crates for tests -- fewer deps, faster compile times. OTOH, we probably already have `rand` somewhere in the dep graph?

glommio/src/lib.rs
matklad: I suggest adding `pub mod sync;` for all synchronization primitives, following std. 

Though, `std::sync` is a bit of a kitchen sink (pun not intended) and there are plans to, eg, uplift `atomic` to top level, move `Arc` to `rc`, etc.

glommio/src/read_write_lock.rs
matklad: :heavy_plus_sign: :100:

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+
+///Error which indicates that ReadWriteLock is closed and can not be use
+///to request lock access
+pub struct LockClosedError;
+
+impl Error for LockClosedError {}
+
+impl Display for LockClosedError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.write_str("Lock already closed")
```
matklad, line 70: Here and elsewhere

```suggestion
        f.write_str("lock already closed")
```

https://github.com/rust-lang/api-guidelines/commit/201d9036dbe8640eef6054d2d9632ea0a0532ab6

If you are wondering why `io::Error` looks like `"Permission denied"`, that's because they use C library function to covert errno to message, and that is uppercase for Linux (but lowercase for windwos).

glommio/src/lib.rs
laa: I will move to the sync module then.

glommio/src/read_write_lock.rs
matklad: My gut feeling is that we should use custom error for channels as well, but I can't really articulate this without reflecting on this for like half an hour :)

glommio/src/lib.rs
glommer: Thanks! If you are familiar with git black magic, it'd be slightly better to first have a patch that moves the semaphore into sync, and then add this on top.

glommio/src/read_write_lock.rs
laa: In our previous conversation, about methods which fully read and write from the disk you suggested to use a bit different names to do not have confusion with futures API. I supposed that it is good to keep this strategy and call this lock as  ReadWriteLock instead of RwLock. But I am pro for any choices. That depends on the overall strategy of the creation of API. So at your choice. What do you think is better to distinguish names which are used in std library or use similar ones?

glommio/src/read_write_lock.rs
```diff
@@ -0,0 +1,1214 @@
...
+///
+/// [`Mutex`]: super::Mutex
+#[derive(Debug)]
+pub struct ReadWriteLock<T> {
+    rw: Rc<RefCell<State<T>>>,
+}
+
+#[derive(Debug)]
+struct State<T> {
+    value: UnsafeCell<T>,
```
matklad, line 177: Yeah, given that this *already* is behind `RefCell`, I don't think an `UnsafeCell` is needed? I think we'll need a smudge of unsafe code here to extend lifetimes past `Ref` for guards, that can be done without UnsafeCell probably. 

Cursory look, might be wrong here :-)

glommio/src/read_write_lock.rs
laa: I noticed that you use io::Error, but errors in "sync" API, let's start to call it like this, have nothing to do with io. It is not so hard to implement From trait for them. But when I look at "Broken pipe" message in io::Error I really do not understand how that is related to the problem and it makes me fell that developers decided to do not put too much of effort to create "clear" API. That is why I used a different error class. It is more about the perception of API than functionality. Do you have a strong feeling that io::Error should be used instead ?

glommio/src/read_write_lock.rs
laa: Not taking into account that is a small but unnecessary slowdown, unsafe API is very localized, I had a feeling that I will not be able to rid off Ref/RefMut and they will be exposed to the user API, which contradicts with the meaning of "clear"  API. That was the feeling which I have got from my previous experience. Those things quity sticky once you start to work with them as in this example https://rust-unofficial.github.io/too-many-lists/fourth-peek.html .  But I will try to fix that. Maybe that is very easy and my feelings are wrong.

glommio/src/read_write_lock.rs
matklad: >It is more about the perception of API than functionality.

It's a bit less nebulous than that. `io::Error` has support for dynamically dispatched errors, it is a bit like `RuntimeException` from other languages, and is the perfect choice when:

* you want to expose underlying OS-errors as is
* you want some generic error with type erasure 

(I've wrote a bit about its design here: https://matklad.github.io/2020/10/15/study-of-std-io-error.html)

For sync primitives, the set of error conditions is fixed (closed), so an `enum` works better. And, as an argument from authority, crossbeam-channel does that.

glommio/Cargo.toml
laa: Sure, I will check or change the crate. Not a problem at all :-)

glommio/src/read_write_lock.rs
glommer: I have been welcoming @matklad 's perspective on this as a more experienced rustacean and he seems to slightly favor what you did.

For me personally I always found "io" to be a bit of a misnomer: they are errors returned in system calls, and not all system calls do I/O anyway...

The appeal of using `io::Result` for me is that you can call that together in a function that does other things as well and because they all return the same error you can `?` without conversions.

You can always implement conversions, that's true, but as far as I know you cannot implement out-of-crate conversions. So the poor user of this would not be able to easily convert to `io::Error`: we'd have to provide the conversion class.

And if we are the ones providing the conversion class, why not go all the way ?

Unless I am missing something with the out-of-crate conversion, then if you keep this as a standard error I think we should at least provide the conversion to `io::Error`. `io::Error`s are so ubiquitous in an I/O-intensive crate, which is what most functions end up returning anyway. (Think of using a `rwlock` to intermediate access to a file... you will want your function that does that to return `io::Error`!)

glommio/src/read_write_lock.rs
laa: OK

glommio/src/read_write_lock.rs
laa: Sure

glommio/src/read_write_lock.rs
laa: OK

glommio/src/read_write_lock.rs
glommer: I am fine with custom errors but in that case I think we should implement our own conversions to `io::Error`. That is, unless I am mistaken about out-of-crate implementation of traits.

Think of the example I gave above, where you may have a `read_file` function that protects the file with a `rwlock`.
It's very clear to me that the top level function should return `io::Result` so we should make as easy as possible to convert back to that.

laa:
Hi guys, thank you for your review! I answered on a couple of observations. About the rest, I need to have spare time to address them. Will do once I have the time, probably this evening. I feel that we have quite different time zones so you will see my changes only tomorrow :-).

glommio/src/read_write_lock.rs
laa: Hi @glommer @matklad , I tried to use RefCell both as a wrapper on the value and using top-level RefCell in both cases I get  `returns a value referencing data owned by the current function` the same as in this example https://rust-unofficial.github.io/too-many-lists/fourth-peek.html also I tend to agree with the author of example that in such case Ref/RefMut can not be hidden from the user. I prefer to do not expose internals to the user and add one more level of indirection. UnsafeCell in this context is used in absolutely the same way as it is done in RwLock, unsafe code is quite localized and as result not dangerous. If you do not mind I prefer to keep UnsafeCell. Looking for your feedback :-)

laa: Hi, @glommer @matklad all problems with UnsafeCell are boiled down to the single one - `get_mut` can not be implemented without exposing of Ref/RefMut to the client, the reason is quite simple in such case reference which is owned by Ref/RefMut is exposed to the client while Ref is dropped in the function. If you think that is OK to expose Ref or better IMHO WriteGuard to the client I will do that otherwise I think UnsafeCell should be kept in the code.

glommio/src/read_write_lock.rs
laa: Hi, @glommer in this test at first we await that task will be finished first and once it is finished we check that error is returned as a result of the read. So as I understand `.await` at the end of the task creation guarantees this order, because of for me it looks like analogous of .join() for a normal thread. Am I mistaken?

glommio/src/read_write_lock.rs
laa: Hi, the same as above I suppose that order of tasks is predefined because we are awaiting its execution to be finished. As I understand if we call `.dispatch()` at the end of task creation, the task will be executed at some moment in background, but if we call `.await` it is the same as we would call `.join()` once we spawn a new thread. Am I mistaken?

glommio/src/read_write_lock.rs
laa: Hi, @glommer @matklad let me summarize. So we keep custom errors but will provide conversion to the io::Error is that correct? @matklad thank you for the link I will read it today. BTW if we use Lock errors for the RwLock it would be also useful to use the same error for the Semaphore for the consistency fo the view. Do not you think so?

glommio/src/read_write_lock.rs
matklad: >get_mut can not be implemented without exposing of Ref/RefMut to the client,

I mean, we can do this with `unsafe`, but without `UnsafeCell`: `unsafe { &*(&state.value as *const _) }`. We fundamentally need unsafe *somewhere*, as we are implementing interior mutability primitive. The question is about minimizing it, not eliminating (in this case). 

Although I am not entirely clear on safety implications of either approach, I'll take a thorough look once I have resources for a thorough look.

One thing that gives me a pause though, is that std's MutexGuard looks like this:

```rust
pub struct MutexGuard<'a, T: ?Sized + 'a> {
    lock: &'a Mutex<T>,
    poison: poison::Guard,
}
```

Naively, I would expect to see a similar lifetime setup here: a reference instead of a phantom data. Intuitively, I don't think there should be differences in how ownership is setup between async and sync cases. But this is intuition, I might be wrong.

laa: Hi @matklad . I have noticed this approach too. I thought that in our case a reference to the RwLock is at the fact the reference to its state but without additional indirection + PhantomData would cost nothing in that case. That is why I used the state field directly.  As about the get_mut I understood your idea. As for guards, I plan to incorporate Ref/RefMut as a field which will remove the need to have PhantomData and unsafe code at all. But in such case, I suppose there is no need to use a reference to the lock itself as the lifetime will be dictated by the Ref lifetime. Do you have any objections?

laa: I mean if I would use Ref to the state, actually I suppose I need even RefMut, acquisitions of reading accesses to the lock would panic during switching of execution inside of fibres.

glommio/src/read_write_lock.rs
glommer: I think those are different scenarios.
One thing is to have methods that exist in general traits having the same name but doing different things: that should definitely be different.

The other is to have different classes which serves the same purpose but are different in implementation.

glommio/src/read_write_lock.rs
glommer: I am ok with using custom errors with a provided conversion.

Indeed if you are willing to go the extra mile, converting the existing errors for Semaphore and even Channels for consistency would be super.

glommio/src/read_write_lock.rs
glommer: Ok. It's my bad. You are testing a different thing than what I thought so my brain mentally attached a `.detach()` there =)
(You knw wn you rd jst half a sentc but your brain fills the rest?)

You are testing that the read or write will fail because the lock was closed, not that the lock was held.

You are correct, the task will spawn but the `.await` statement is a synchronization point.

laa: > You knw wn you rd jst half a sentc but your brain fills the rest? 
I always skim the news articles and sometimes the next day I become surprised that article, when I read it again, says absolutely different thing :-D .

glommer: Well in that case maybe they changed it as they declared war on Eastasia

glommio/src/read_write_lock.rs
laa: Sure will do. Will provide a separate PRs.

laa:
Hi, @glommer @matklad I have fixed all observations with exception of the removal of UnsafeCell. Because I am waiting for feedback from @matklad in order to do not redevelop the same things twice.

laa:
Hm on my side all tests are passing let me fix the build and I will push again.

matklad:
>Because I am waiting for feedback from @matklad in order to do not redevelop the same things twice.

I suggest switching to an obstruction-free version of the algorithm here, glommio, while exciting, is not on the top of my priority list right now, and you'll get priority inversion if you block on me ;)

More generally, I believe it is loads easier to modify existing impl to be better, than to write the best impl from scratch, so, even if the end design will turn slightly different, i don't think there would be any wasted effort.

laa:
> I suggest switching to an obstruction-free version of the algorithm here
I have replaced `UnsafeCell` by value wrapped by `Rc<RefCell>`.  I wanted to replace only by `RefCell` but in such case, I would not implement into_inner using safe code at least.

glommio/src/read_write_lock.rs
```diff
@@ -43,14 +43,16 @@ use std::future::Future;
 use std::ops::{Deref, DerefMut};
 use std::pin::Pin;
 use std::task::{Context, Poll, Waker};
+use std::io;
+use futures_lite::io::ErrorKind;
```
glommer: any particular reason we're using `ErrorKind` from `futures_lite` instead of `std::io::ErrorKind` ?

glommio/src/read_write_lock.rs
```diff
@@ -112,6 +114,18 @@ impl From<LockClosedError> for TryReadWriteLockError {
     }
 }
 
+impl From<LockClosedError> for io::Error {
+    fn from(er: LockClosedError) -> Self {
```
glommer: I think it is useful to map - whenever possible - particular error kinds to specific `io::Error` error kinds.
This allows the user to test for what happened a lot easier in a call chain that is removed many layers from the first caller.

When we convert, I'd lean towards converting `Closed` to `BrokenPipe` and `WouldSuspend` (I like the name!) to `WouldBlock`. This will also make the conversion of channels easier.

Bringing @matklad into this since he had opinions about the errors as well.

glommio/src/read_write_lock.rs
```diff
@@ -436,10 +450,10 @@ impl<T> ReadWriteLock<T> {
         Ok(unsafe { &mut *(state.value.get()) })
     }
 
-    /// Locks this ReadWriteLock with shared read access, blocking the current fiber
-    /// until it can be acquired.
+    /// Locks this ReadWriteLock with shared read access, suspending the current fiber
+    /// until lock can be acquired.
```
glommer: good! I didn't comment on that, but a pet peeve of mine just so you know:
I think the words "it" and "this" should be avoided whenever possible. Most of the time they are ambiguous and it is not clear to which noun they refer to.
(your usage of "this" above is clear enough because it is attached to `ReadWriteLock`)

glommio/src/sync/rwlock.rs
```diff
@@ -126,10 +125,10 @@ impl From<TryLockError> for io::Error {
     }
 }
 
-struct Waiter<T> {
+struct Waiter {
```
glommer, line 128: Looks better to my eyes.

@matklad I'll give you a chance to comment before settling, though.

If any of you know anybody who works on github, please beg them to implement comments in the commit message. This comment is about the entire diff, not about the line in which I had to attach it!!

glommer (commented):
Almost there! The main two issues are:

- I don't understand why we are using futures_lite version of ErrorKind.
- I have a slight preference towards using definite Error kinds in the conversion to `io::Error`. I will start a chat on Zulip about this so we can coordinate. If you guys have a strong case for using `Other` then the code as is is fine

glommio/src/read_write_lock.rs
laa: Hi @glommer it is already fixed in the latest commit. I suppose you missed it. IntelliJ automatically inserted the use of futures_lite, I have changed that already.

glommio/src/read_write_lock.rs
laa: Completely agree, seem I missed that. I do not like it too. Sometimes use automatically though. It is not clear what "it" meanst. I will be a bit of busy at the bigging of next week will provide the next PR in the middle of next week.

glommio/src/read_write_lock.rs
laa: Well, I suppose in general error processing system is quite raw in Rust, yet. I suppose they already understood that and created error handling working group. If that would be an application I would suggest to use one of error processing creates which enhance embedded rust functionality. I like `anyhow` BTW. That is a minor fix, will do that ASAP.

laa:
Hi @glommer . Fixed all observations.

glommer (approved):

PR #170, opened 2020-11-13
improvements for io tests

glommer:
### What does this PR do?

Fixes so tests work on docker, and also when /tmp is not backed by tmpfs.


laa:
Thank you for the fix. I wanted to report an issue but you were ahead of me :-)

PR #171, opened 2020-11-14
replace std hashmap with ahashmap

glommer:
This is a non-cryptographic hash, therefore faster than rust's standard
hash.

Although we have a FreeList type, which is useful for simple integer
cases, the FreeList API is a bit more restrictive than the Hash and
converting some uses would require work.

Contrast this with AHash, which was essentially a bunch of sed invocations.

This gave me a 5% boost in the storage benchmark I am about to post.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -32,6 +32,7 @@ enclose = "1.1.8"
 scopeguard = "1.1.0"
 pin-project-lite = "0.1.10"
 smallvec = "1.4.2"
+ahash = "0.5.8"
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


Daniel-B-Smith:
Any reason you chose AHashMap instead of hashbrown? The latter uses AHash by default as well but gives you access to niceties like the experimental RawEntry API.

glommer:
No particular reason. I found it first and seemed easy enough to use.

Happy to take a PR to hashbrown if you find it nicer.

PR #175, opened 2020-11-15
Doc Touchups

wackywendell:
### What does this PR do?

These are mostly "touchups" - some documentation clarifications and a bit of copy-editing. There are a couple small changes that are code-related, but they are either no-ops or close to it.

### Motivation

I was reading through the code and the docs to try and understand what this library does and how it works; I thought I'd try and make some clarifications along the way. I used column-100 wrapping with most of my changes; that sort of thing is always tricky to standardize, but that seemed like what this crate was mostly doing. Unfortunately, it means that some of these changes look bigger than they are.

I'm pretty new to this codebase, so my clarifications could use double-checking - and I'd love to know if they are wrong!

I'll add a couple comments to this pull request itself to point out the actual code changes.

### Additional Notes

I still have a lot to learn here, and plan on reading my way through some more code and perhaps making some more contributions along these lines. This PR is already of moderate size, so it felt like a good pausing place to gather some feedback about what is useful here before I get too much farther.

### Checklist

- N/A I have added unit tests to the code I am submitting
- N/A My unit tests cover both failure and success scenarios
- [x] If applicable, I have discussed my architecture
- [x] The new code I am adding is formatted using `rustfmt`


glommio/src/shares.rs
```diff
@@ -7,6 +7,8 @@ use core::fmt::Debug;
 use std::rc::Rc;
 use std::time::Duration;
 
+pub(crate) const MAX_SHARES: u64 = 1u64 << 22;
```
wackywendell, line 10: This is the only actual code change in this PR; I moved a constant used in two places out to a named one.

This could probably use a better name and a comment, but I'm not sure I understand it well enough to add myself.

glommio/src/lib.rs
```diff
@@ -389,26 +406,28 @@ pub mod prelude {
...
-/// local_into(), the type is always () and Rust is not able to elide.
+/// However for associated functions without a self parameter, like `local()` and
+/// `local_into()`, the type is always `()` and Rust is not able to elide.
 ///
-/// Writing Task::<()>::function() works, but it is not very ergonomic.
+/// Writing `Task::<()>::function()` works, but it is not very ergonomic.
 pub type Local = Task<()>;
 
-/// An attribute of a TaskQueue, passed during its creation.
+/// An attribute of a [`TaskQueue`], passed during its creation.
```
wackywendell, line 415: Note that `TaskQueue` is not an exported type of this crate; only `TaskQueueHandle` is exported.

I wonder if we could perhaps change the interface a bit to have `spawn_into` be a method of `TaskQueueHandle`, and then rename the private `TaskQueue` to... I don't know, something else, and rename `TaskQueueHandle` to `TaskQueue`, making the interface a bit clearer by hiding the difference between a `Queue` and a `Handle` for users.

Questions there:
- Can `spawn_into` be moved onto `TaskQueue`?
- Related: what happens if you try to `spawn_into` with a `LocalExecutor` and a `TaskQueueHandle` that don't match?

glommio/src/lib.rs
```diff
@@ -29,12 +34,14 @@
 //!     }).unwrap();
 //! ```
 //!
-//! Although this is not forced upon the user, by creating N executors and binding
-//! each to a specific CPU one can use this crate to implement a thread-per-core
-//! system where context switches essentially never happen, achieving maximum efficiency.
+//! ### Pinned threads
```
wackywendell, line 37: I added in some section headers and changed some wording through here. Unfortunately, I wasn't really sure on the line-lengths being used, and did some line-wrapping at 100 characters; this means that its hard to tell from the diff what did and did not change.

glommer: My philosophy on style is that if a machine does not complain a human shouldn't either.
So whatever passes `cargo clippy` and `cargo fmt` is acceptable to me.

Yes, I fully acknowledge that a consistent code base is desirable (and `cargo clippy` and `cargo fmt` do a great job at that) but as with most things in life it comes with hidden costs: in this case, all the time that is spent with people obsessing and fighting over personal preferences.

tl;dr: do as you please!

glommio/src/io/buffered_file.rs
```diff
@@ -122,7 +128,7 @@ impl BufferedFile {
         enhanced_try!(source.collect_rw().await, "Writing", self.file)
     }
 
-    /// Reads data at the specified position into the user-provided buffer `buf`.
+    /// Reads data at the specified position into a buffer allocated by the OS.
```
glommer, line 130: technically it is not the OS that allocates the buffer.

You are right that the previous comment was wrong (bad copy pasta from write, probably), but the comment refers to the `Vec<u8>` buffer that this returns. That one is allocated by us, not the OS.

glommio/src/io/buffered_file.rs
```diff
@@ -137,42 +143,44 @@ impl BufferedFile {
         Ok(buffer)
     }
 
-    /// Issues fdatasync into the underlying file.
+    /// Issues `fdatasync` for the underlying file, instructing the OS to flush
+    /// all reads/writes to the device, providing durability even if the system
```
glommer: there is no need to flush any reads.

glommio/src/io/dma_file.rs
```diff
@@ -221,51 +229,61 @@ impl DmaFile {
         Ok(ReadResult::from_whole_buffer(buffer))
     }
 
-    /// Issues fdatasync into the underlying file.
+    /// Issues `fdatasync` for the underlying file, instructing the OS to flush
+    /// all reads/writes to the device, providing durability even if the system
```
glommer: same thing here, poor reads are innocent.

glommio/src/io/dma_file.rs
```diff
@@ -221,51 +229,61 @@ impl DmaFile {
...
+    /// crashes or is rebooted.
+    ///
+    /// As this is a DMA file, the OS should not be caching this file; however,
```
glommer: it is guaranteed that the OS should not be caching the file so there is no concern there.

The reason `fdatasync` is needed is, as you well put later, that the device also has an internal cache.

glommio/src/lib.rs
glommer: I liked the section headers, it is a nice touch.

Thanks a lot for this, Wendell.

glommio/src/lib.rs
```diff
@@ -106,89 +127,82 @@
...
-//!    high IOPS scenarios
+//!    poll ring. The poll ring does not rely on interrupts so the system has to
+//!    keep constantly polling if there is any pending work. By not relying on
+//!    interrupts we can be even more efficient with I/O in high IOPS scenarios
 //!
 //! ## Before using Glommio
 //!
-//! Please note Glommio requires at least 256 KiB of locked memory for `io_uring`
-//! to work. You can increase the `memlock` resource limit (rlimit) as follows:
+//! Please note Glommio requires at least 256 KiB of locked memory for
```
glommer: This is a pre-existing problem, but this is a mistake. The minimum requirement currently sits at 512kB.

glommer: See that in the instructions below we actually say 512...

glommio/src/lib.rs
```diff
@@ -197,51 +211,51 @@
...
-//! Due to our immediate needs which are a lot narrower, we make
-//! the following design assumptions:
+//! Due to our immediate needs which are a lot narrower, we make the following
+//! design assumptions:
 //!
-//!  - NVMe. Supports for any other storage type is not even considered.
-//!    This allow us to use io uring's poll ring for reads and writes which
-//!    are interrupt free. This also assumes that one is running either `XFS`
-//!    or `Ext4` (an assumption that Seastar also makes)
+//!  - NVMe. Supports for any other storage type is not even considered. This
```
glommer: The need for NVMe is lifted. You can now use this with other types of media and it will work.
The design decisions are still assuming NVMe: it just won't refuse to work with other types of media.

So we could massage this phrase a bit.

glommio/src/lib.rs
```diff
@@ -197,51 +211,51 @@
...
+//!    allow us to use io uring's poll ring for reads and writes which are
+//!    interrupt free. This also assumes that one is running either `XFS` or
+//!    `Ext4` (an assumption that Seastar also makes)
 //!
 //!  - A corollary to the above is that the CPUs are likely to be the
-//!    bottleneck, so this crate has a CPU scheduler but lacks an I/O
-//!    scheduler. That, however, would be a welcome addition.
+//!    bottleneck, so this crate has a CPU scheduler but lacks an I/O scheduler.
```
glommer: This is an example of where the design decisions are still NVMe-centric.

glommio/src/lib.rs
```diff
@@ -197,51 +211,51 @@
...
-//!    present. In fact, we require a kernel so recent that it doesn't event
-//!    exist: operations like `mkdir, ftruncate`, etc which are not present
-//!    in today's (5.8) `io_uring` are simply synchronous and we'll live
-//!    with the pain in the hopes that Linux will eventually add support for
-//!    them.
+//!  - A recent kernel is no impediment, as long as a fully functional I/O uring
+//!    is present. In fact, we require a kernel so recent that it doesn't event
+//!    exist: operations like `mkdir, ftruncate`, etc which are not present in
+//!    today's (5.8) `io_uring` are simply synchronous and we'll live with the
+//!    pain in the hopes that Linux will eventually add support for them.
```
glommer: and that is coming in 5.12!!! So excited.

glommio/src/lib.rs
```diff
@@ -197,51 +211,51 @@
...
 //!
 //! ## Missing features
 //!
 //! There are many. In particular:
 //!
-//! * There is no yet cross-shard communication nor ergonomic primitives to
+//! * There is not yet any cross-shard communication nor ergonomic primitives to
```
glommer: This is actually false now. I have merged support for cross-shard communication.
We should remove this fully.

glommio/src/shares.rs
glommer: This isn't `MAX_SHARES`.

I don't mind a constant here, but this is really just a math shift. The role of this is to turn a small quantity big so we don't have to resort to float math nor lose precision. When the reciprocal shares are turned to shares there is another shift that takes some of those bits away.

In any case I would prefer you would rebase the tree and remove this patch: it is a documentation PR so it's best and less confusing if it only touch documentation. Happy to tackle this later separately.

glommio/src/lib.rs
glommer: humm...
(I am thinking ...)

There is some merit but I am not sure I like it.

The first disadvantage  is that now you have many different spawn functions in different entities: if the reader is not careful, one can easily mistake `tq::spawn` with `ex::spawn` and end up doing the wrong thing.

Another consideration is what to do if the `TaskQueue` goes out of scope. It becomes impossible to spawn new tasks into it. Should we then just destroy it?

But last and most importantly, the `TaskQueueHandle` is an opaque that can be passed around (because it is `Copy` to many executors).

Maybe there is a way to make it work but I'd like to see it coupled with a good ergonomic way to use it across executors at least. In any case it would be the subject of a different PR

glommer (changes requested):
Very minor changes requested.

Thanks!

glommer:
Quick heads up that I have just merged https://github.com/DataDog/glommio/pull/176 to make sure that the links you are adding are sane.

Thanks again!

glommio/src/io/buffered_file.rs
wackywendell: Ah, thanks. In that case, "allocated by this library" would perhaps be more appropriate?

glommio/src/io/buffered_file.rs
wackywendell: 🤦  Of course

glommio/src/io/buffered_file.rs
glommer: yes

glommio/src/lib.rs
wackywendell: Fair enough; it works well enough as is, and unless a new interface here is clearly better, its not worth it.

> Maybe there is a way to make it work but I'd like to see it coupled with a good ergonomic way to use it across executors at least.

OH, there is something I missed. Is it safe to use the same `TaskQueueHandle` in different executors?

glommer: Not like this, no.
a `TaskQueue` is registered with a specific executor so if you try to use it with another one it may not exist.

At the end of the day it is just an integer, so it would be useful to have an API to register it in all executors (seastar has such an API, for example)

glommer (approved):

wackywendell:
:thinking: The re-exporting through `prelude` breaks the `LocalExecutor` -> `Latency` link inside `prelude`, because the `LocalExecutor` docs are created at both `glommio/struct.LocalExecutor.html` and at `glommio/prelude/prelude/struct.LocalExecutor.html`, but `Latency` is only created at `glommio/enum.Latency.html`. Hm.

glommer:
I am considering ditching the prelude.

@matklad is of the opinion that this should only be used for traits, not to simplify exports.
That does seem to be the way the rust community is leaning towards and I am growing sympathetic to this opinion as I code around it.

wackywendell:
@glommer that makes sense to me; Rust tends toward explicitness. Although that seems like a big change for one pesky dead link :thinking:

glommer:
well, it's all semantics. If we were about to do it anyway, then it is not because of the deadlink, but motivated by it =)

wackywendell:
Haha, fair enough :laughing: 

Well, to unblock the whole thing, I added `#[doc(no_inline)]` which means that in the docs for `prelude`, all the structs / enums / etc. are shown as re-exports, and you have to click through for documentation. This clears the dead links and passes `cargo deadlinks` locally; I assume CI will pass as well. Sound good to you?

glommer:
yes

PR #182, opened 2020-11-19
Double access to the map was replaced by entry access in LocalExecutor.

laa:
### What does this PR do?

During removal of the queue from local executor double access to the available_executors map is performed.
This access was replaced by access to the entry acquired from the same map.

### Motivation

It is a small change from a non-critical path. But such double access to the map always catches the eye so I decided to replace it.

### Additional Notes

There is also a small change in the comment of the run method of LocalExecutor

### Checklist

### DIFF
--- a/glommio/src/executor.rs
+++ b/glommio/src/executor.rs
@@ -31,6 +31,7 @@
 #![warn(missing_docs, missing_debug_implementations)]
 
 use std::cell::RefCell;
+use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, BinaryHeap};
 use std::fmt;
 use std::future::Future;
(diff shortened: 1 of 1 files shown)


glommer:
Looks good but it's strange that cargo doc is not passing.
Probably some other patch broke it, which is still strange since we test for all.

Likely this is a result of some PR that was sent before `cargo deadlinks` were added and got merged.

I'll fix that then merge this (But to set expectations, I am a bit busy today)

laa:
Well, not my fault that links are broken but I will provide updated PR in a while :-)

glommer:
Agree, I looked through the logs and it is some file stuff. So it was most likely me =)

Trying to wonder how it happened but... it doesn't matter.

You don't need to do anything here. Once I fix the master branch this will automatically pass.

glommer:
My best guess is that CI is broken due to some versioning issue with cargo deadlinks. It passes for me locally.
I am merging this because this is clearly innocent then I'll figure out what to do.

The rust tooling in the CI is pretty old too, so it may generate new warnings once I update. we'll see.

PR #184, opened 2020-11-20
Preventing Semaphore keeping of duplicates of wakers

laa:
### What does this PR do?
Because LocalExecutor checks readiness of root future by using loop-pooling without the help of wakers there are could be situations when Semaphore future can be pooled many times generating an excessive amount of wakers stored in the waiting queue. This simple code 

```
        let ex = LocalExecutor::make_default();
        let semaphore = Rc::new(Semaphore::new(0));
        let semaphore_c = semaphore.clone();

        ex.spawn(async move {
            for _ in 0..100 {
                Timer::new(Duration::from_micros(100)).await;
            }
            assert_eq!(1, semaphore_c.state.borrow().list.len());

            semaphore_c.signal(1);
        })
        .detach();

        ex.run(async move {
            let _ = semaphore.acquire(1).await.unwrap();
        });
```

ends up with :
1. 10K of waiters registered inside of Semaphore.
2. Panic on the call of the release method.
  
RwLock does not have such problems but it keeps the earliest weaker to wake up the fiber but the good practice is to keep the latest one. This part also was implemented.

### Motivation

What inspired you to submit this pull request? - My desire to increase the stability of glommio 

### Checklist

### DIFF
--- a/glommio/src/sync/rwlock.rs
+++ b/glommio/src/sync/rwlock.rs
@@ -289,12 +289,22 @@ impl State {
         debug_assert!(!(self.readers > 0 && self.writers > 0));
 
         let entry = self.waiters_map.entry(id);
-        if let Vacant(entry) = entry {
-            entry.insert((kind, waker));
-            self.waiters.push_back(id);
+        match entry {
+            Vacant(entry) => {
+                entry.insert((kind, waker));
+                self.waiters.push_back(id);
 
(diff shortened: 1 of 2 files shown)


laa:
Invalid links - again 👎

glommer:
I disabled `cargo deadlinks` for now

PR #190, opened 2020-11-26
Remove redundant methods and tag value from Task implmentation

laa:
### What does this PR do?

Redundant methods and tag value were removed from Task implementation

### Motivation

Those methods and value were not used. Removing them will improve code readability and supportability as result.

### Checklist

[s] The new code I am adding is formatted using `rustfmt`

### DIFF
--- a/glommio/src/multitask.rs
+++ b/glommio/src/multitask.rs
@@ -31,7 +31,7 @@ use std::task::{Context, Poll};
 /// Once a `Runnable` is run, it "vanishes" and only reappears when its future is woken. When it's
 /// woken up, its schedule function is called, which means the `Runnable` gets pushed into a task
 /// queue in an executor.
-pub type Runnable = task_impl::Task<()>;
+pub type Runnable = task_impl::Task;
 
 /// A spawned future.
 ///
@@ -47,7 +47,7 @@ pub type Runnable = task_impl::Task<()>;
 /// ```
 #[must_use = "tasks get canceled when dropped, use `.detach()` to run them in the background"]
 #[derive(Debug)]
-pub struct Task<T>(Option<JoinHandle<T, ()>>);
+pub struct Task<T>(Option<JoinHandle<T>>);
 
 impl<T> Task<T> {
     /// Detaches the task to let it keep running in the background.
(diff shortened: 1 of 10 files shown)


glommio/src/multitask.rs
```diff
@@ -31,7 +31,7 @@ use std::task::{Context, Poll};
 /// Once a `Runnable` is run, it "vanishes" and only reappears when its future is woken. When it's
 /// woken up, its schedule function is called, which means the `Runnable` gets pushed into a task
 /// queue in an executor.
-pub type Runnable = task_impl::Task<()>;
+pub type Runnable = task_impl::Task;
```
glommer, line 34: No need to change anything because this is not related. But as I read this, I wonder if we shouldn't just get rid of Runnable now ? Or at least make it non-public.

It's just a task.

glommio/src/task/task_impl.rs
```diff
@@ -125,74 +118,9 @@ impl<T> Task<T> {
 
         unsafe { ((*header).vtable.run)(ptr) }
     }
-
```
glommer, line 125: wait, this feels like too much. or at least needs to be justified separately.

We can't cancel tasks now?

glommer (commented):
They are mostly okay but I don't understand how it is safe to remove the cancel method from the task. We certainly can ?

Maybe they feel like unused because there are no tests in the crate using them (in which case we should probably add them), but that doesn't mean we should remove the ability to cancel a task.

glommio/src/task/task_impl.rs
laa: Hi @glommer , the task can be cancelled using JoinHandle. Here is the code from the Task API.

```
pub async fn cancel(self) -> Option<T> {
        self.0.cancel().await
    }
```

And down to the stack of calls to the JoinHandle.

glommio/src/multitask.rs
laa: I wondered the same. After some reflection, I came to the opinion that Runnable as for now is part of the public API and  task_impl::Task is more like the internal type. So I personally still like this type of aliasing.

glommio/src/task/task_impl.rs
glommer: Thanks.

That makes sense. In the future, I expect these kind of things to feature in the commit message, to help
both the reviewer and future readers understand the change.

I consider you a part of the family now,  so I'll be a bit more strict on this =)

for now I am merging this

glommer (approved):

PR #192, opened 2020-11-27
state field of Task's state machine was changed from AtomicUsize to usize

laa:
### What does this PR do?

state field of Task's state machine was changed from AtomicUsize to usize.
States which were used only to ensure thread safety (REGISTRING and NOTIFYING) were removed as not needed.

### Motivation

Removing of usage of membars during changing of a state of the state machine.

### Checklist

### Notes from maintainer:

The effects of this are visible on the executor benchmark. Quite consistently I get a 70s speedup in task execution

Before:
```
cost to run task no task queue: 216ns
cost to run task in task queue: 377ns
```
After

```
cost to run task no task queue: 129ns
cost to run task in task queue: 301ns
```

There are similar (albeit a bit less) speed ups in the contended cases for semaphore and local channel.

Very good job, Andrey!

### DIFF
--- a/glommio/src/task/header.rs
+++ b/glommio/src/task/header.rs
@@ -4,9 +4,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
 use core::alloc::Layout;
-use core::cell::UnsafeCell;
 use core::fmt;
-use core::sync::atomic::{AtomicUsize, Ordering};
 use core::task::Waker;
 
 use crate::task::raw::TaskVTable;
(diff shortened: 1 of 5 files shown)


glommer (commented):
Hi Andrey

It would really help with review if you could craft a more informative commit message. That helps us in the future as well to navigate past changes.

I have been a bit relaxed with this requirement but this is such a crucial change that I'd like to understand it well.

Things I'd like to see in the message:
- you mentioned that it was clear for you why this failed before. Why was it?
- Where should a reviewer focus his attention when approaching this?
- Aside from the removal of the two states, is the rest of the work just mechanical (fetch_and == and, etc) or is there any part that had to change to adapt?

Feel free to just do a `git commit --amend` with a new message and then a `git force push` with the result

laa:
Hi, @glommer I have updated comment.  As for mistake which I noticed yesterday that was my mistake actually :-) that piece of code appeared to be correct :-) Probably it was a mechanical mistake during conversion or something very different hard to say as for now.

glommer (approved):
And the best thing is that the effect of this is visible in the executor benchmark.

Before:
```
cost to run task no task queue: 216ns
cost to run task in task queue: 377ns
```

After
```
cost to run task no task queue: 129ns
cost to run task in task queue: 301ns
```

Around 70ns, quite respectable!

glommer:
It kind of sucks that github doesn't include the PR description in the merge commit.
That's good information to have in the tree, I wanted the benchmark results to be there =(

Maybe there's a setting for that?

PR #200, opened 2020-12-01
improve efficiency of the executor

glommer:
Improves efficiency of spawning tasks in the executor, by pre-computing the task queue index

Before:
```
cost to run task no task queue: 126ns
cost to run task in task queue: 300ns
```

After
```
cost to run task no task queue: 93ns
cost to run task in task queue: 257ns
```

No regressions found in any of the existing benchmarks. Semaphore and local channel show similar improvements for the contended case.

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "glommio"
-version = "0.2.0"
+version = "0.3.0"
 authors = [ "Glauber Costa <glauber@datadoghq.com>",
             "Hippolyte Barraud <hippolyte.barraud@datadoghq.com>",
             "DataDog"]
--- a/glommio/src/executor.rs
+++ b/glommio/src/executor.rs
@@ -50,7 +50,7 @@ use crate::multitask;
 use crate::parking;
 use crate::task::{self, waker_fn::waker_fn};
 use crate::{IoRequirements, Latency};
-use crate::{Local, Reactor, Shares};
+use crate::{Reactor, Shares};
 use ahash::AHashMap;
 
 static EXECUTOR_ID: AtomicUsize = AtomicUsize::new(0);
@@ -114,8 +114,8 @@ impl Default for TaskQueueHandle {
 }
 
 #[derive(Debug)]
-struct TaskQueue {
-    ex: Rc<multitask::LocalExecutor>,
+pub(crate) struct TaskQueue {
+    pub(crate) ex: Rc<multitask::LocalExecutor>,
     active: bool,
     shares: Shares,
     vruntime: u64,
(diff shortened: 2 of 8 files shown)


glommio/src/executor.rs
```diff
@@ -781,15 +753,21 @@ impl LocalExecutor {
...
             .borrow()
             .active_executing
             .as_ref()
-            .map(|x| x.borrow().ex.clone())
-            .or_else(|| self.get_executor(&TaskQueueHandle { index: 0 }))
-            .unwrap();
-        Task(ex.spawn(future))
+            .map_or_else(
+                || self.get_queue(&TaskQueueHandle { index: 0 }),
+                |x| Some(x).cloned(),
```
zserik, line 763:
```suggestion
                |x| Some(x.clone()),
```

glommer:
Thanks @zserik . I will leave this patch cooking for a bit longer. Due to a change that I consider too drastic (the fact that you can no longer spawn before run), I will bump the version to 0.3.0 once I do it.

glommio/src/executor.rs
zserik: Ah, I didn't realise that `x` is a `Rc`... so it should be `|x| Some((*x).clone())` instead. Sorry.

glommio/src/executor.rs
```diff
@@ -781,15 +753,21 @@ impl LocalExecutor {
...
             .borrow()
             .active_executing
             .as_ref()
-            .map(|x| x.borrow().ex.clone())
-            .or_else(|| self.get_executor(&TaskQueueHandle { index: 0 }))
-            .unwrap();
-        Task(ex.spawn(future))
+            .map_or_else(
+                || self.get_queue(&TaskQueueHandle { index: 0 }),
+                |&x| Some(x.clone()),
```
zserik, line 763: Why do we even need a `clone()` here?

zserik:
I created a PR to `glommer:executor` which contains a better suggestion, as I've made an error in the last one...

glommio/src/executor.rs
```diff
@@ -915,6 +896,24 @@ impl LocalExecutor {
         }
     }
 
+    // this happens if we spawn() into the executor before we run().
+    // I am considering making this illegal, but in the meantime...
+    fn activate_pre_spawned_queues(&self) {
```
zserik, line 901: If we can no longer `spawn` before `run`, why is this `fn` here?

glommer: should have been deleted. It was added to handle the case where we spawn before run and I forgot to delete. Thanks!

glommio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "glommio"
-version = "0.2.0"
+version = "0.3.0"
```
zserik, line 3: `examples/Cargo.toml` `[dependencies] glommio = ...` should be updated to match the new version number.

glommer: you are faster than the CI =p

glommio/src/executor.rs
glommer: I am now wondering if I haven't simply uploaded the wrong version entirely? The spawn functions are still public =(

glommer:
ETA to publish a new crate is ~ 2 weeks, hopefully when the network API stabilizes

PR #201, opened 2020-12-02
refactoring / maybe some improvements

zserik:
### What does this PR do?

This PR gets rid of some `unwrap()`s, mainly in destructors, which should make it easier to guarantee that these `Drop` impls can't crash the program (otherwise, panic'ing in `Drop` impls could cause an abort if the `Drop` was invoced while panic-unwinding).

### Motivation

I wanted to get familiar with the code and inspected the code for somewhat common maybe-code-smells I was familiar with.

### Related issues

None.

### Additional Notes

I was able to run the test and bench suite. I dropped changes which hadn't the intended effect.

### Checklist

No new unit tests should be needed, as no new functionality was added, only existing functionality was refactored without "completely" breaking fn signatures
* [X] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/src/channels/local_channel.rs
+++ b/glommio/src/channels/local_channel.rs
@@ -6,7 +6,6 @@
 use crate::channels::{ChannelCapacity, ChannelError};
 use futures_lite::future;
 use futures_lite::stream::Stream;
-use futures_lite::FutureExt;
 use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::io;
@@ -224,13 +223,11 @@ impl<T> LocalSender<T> {
     ///
     /// [`try_send`]: struct.LocalSender.html#method.try_send
     pub async fn send(&self, item: T) -> Result<(), ChannelError<T>> {
-        if !self.is_full() {
-            self.try_send(item)
-        } else {
+        if self.is_full() {
             let waiter = future::poll_fn(|cx| self.wait_for_room(cx));
             waiter.await;
-            self.try_send(item)
         }
+        self.try_send(item)
     }
 
     /// Checks if there is room to send data in this channel
@@ -299,14 +296,13 @@ impl<T> Drop for LocalSender<T> {
         // wants to wake up anyway.
         state.send_waiters.take();
 
-        if state.recv_waiters.is_none() {
-            return;
-        }
-        let waiters = state.recv_waiters.replace(VecDeque::new()).unwrap();
-        drop(state);
+        if let Some(ref mut waiters) = &mut state.recv_waiters {
+            let waiters = core::mem::replace(waiters, VecDeque::new());
+            drop(state);
 
-        for w in waiters {
-            w.wake();
+            for w in waiters {
+                w.wake();
+            }
         }
     }
 }
@@ -316,15 +312,13 @@ impl<T> Drop for LocalReceiver<T> {
         let mut state = self.channel.state.borrow_mut();
         state.recv_waiters.take();
 
-        if state.send_waiters.is_none() {
-            return;
-        }
+        if let Some(ref mut waiters) = &mut state.send_waiters {
+            let waiters = core::mem::replace(waiters, VecDeque::new());
+            drop(state);
 
-        let waiters = state.send_waiters.replace(VecDeque::new()).unwrap();
-        drop(state);
-
-        for w in waiters {
-            w.wake();
+            for w in waiters {
+                w.wake();
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


zserik:
Don't merge. I found other locations of the second pattern.

glommer:
So if I understand you correctly, the Rust compiler is not able to inline things like "async fn small_fn() -> x {}" ?
Isn't that a compiler issue?

Couple of problems from the get go:
cargo bench fails with:
```
thread 'unnamed-0' panicked at 'cannot access a scoped thread local variable without calling `set` first', /home/glauber/.cargo/registry/src/github.com-1ecc6299db9ec823/scoped-tls-1.0.0/src/lib.rs:168:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any', glommio/benches/executor.rs:19:25
error: bench failed
```

Many similar failures with cargo test as well.

We are currentlu not running `cargo test` on CI because CircleCI's kernel is unfortunately not recent enough to run io_uring.

zserik:
ok, I was able to upgradde my kernel, I'll try to fix the failing tests.

zserik:
I ran each file separately through `cargo test`, at least the tests should, except for maybe some flaky tests, now run without failures. My patch of `glommio/src/executor.rs` was probably the faulty one.

zserik:
Ok, I can't get the patch for `executor.rs` to work. It seems to fail at `executor::test::test_ping_pong_yield`, but that test seems pretty flaky to me, it fails at:
```rust
    // inside 'async fn do_ping_pong_work(...)'
    assert!(counter.get() < 10);
```
but only sometimes...
It seems like the semaphore stops working, I dunno why. If I add a `eprintln!`, the execution is serialized and the error doesn't occur, which makes this a bit difficult to debug...

zserik:
perf differences on `nightly rustc` are negligible.
```
master branch:
     Running target/release/deps/executor-812c847ee363f1ac
cost to spawn an executor : 23.922048ms
cost to spawn and join an executor : 24.154901ms
cost to run task no task queue: 199ns
cost to run task in task queue: 469ns
     Running target/release/deps/local_channel-773bfbfcfadca837
cost of sending contended local channel: 182ns
cost of sending uncontended local channel: 15ns
     Running target/release/deps/preempt-3b94b415f25b4a5c
cost of checking for need_preempt: 7ns
     Running target/release/deps/semaphore-8e8437bf33768304
cost of acquiring uncontended semaphore: 21ns
cost of acquiring contended semaphore: 374ns
     Running target/release/deps/shared_channel-989065d162b7d576
cost of sending rust standard channel 283ns, capacity: 1024
cost of receiving rust standard channel: 283ns, capacity 1024
cost of sending shared channel 28ns, capacity: 1024
cost of receiving shared channel: 29ns, capacity 1024


misc branch:
     Running target/release/deps/executor-812c847ee363f1ac
cost to spawn an executor : 23.602599ms
cost to spawn and join an executor : 24.043879ms
cost to run task no task queue: 196ns
cost to run task in task queue: 472ns
     Running target/release/deps/local_channel-773bfbfcfadca837
cost of sending contended local channel: 182ns
cost of sending uncontended local channel: 16ns
     Running target/release/deps/preempt-3b94b415f25b4a5c
cost of checking for need_preempt: 7ns
     Running target/release/deps/semaphore-8e8437bf33768304
cost of acquiring uncontended semaphore: 21ns
cost of acquiring contended semaphore: 379ns
     Running target/release/deps/shared_channel-989065d162b7d576
cost of sending rust standard channel 318ns, capacity: 1024
cost of receiving rust standard channel: 317ns, capacity 1024
cost of sending shared channel 27ns, capacity: 1024
cost of receiving shared channel: 27ns, capacity 1024
```

zserik:
I'll probably continue to split this PR into two parts (and probably throw away the second part), and reduce this to only the changes which don't transfrom `async` fns to `return` `impl Future` fns.

zserik:
I've reverted the change to `optimistic` as I wasn't sure if this crate depends on the call time of `Drop` of the `Future` argument.

glommer:
In general I think it is better to go for small localized changes anyway. So feel free to push just a piece of this that you feel is ready to go and we'll look at that separately

zserik:
I condensed the PR pretty much, the parts are pretty small, I don't really think it is worth it opening 1 PR per change...

Changes:
 * `channels::..::LocalSender::send`: move the `try_send` outside of the if block, it doesn't depend on the condition
 * `LocalSender/Receiver::drop`: use `if let` bindings instead of `.is_none()` + `.unwrap()`
 * `LocalReceiver::(Stream)poll_next`, `shared_channel::ConnectedReceiver::(Stream)poll_next`: get rid of unnecessary usage of `futures_lite::future::poll_fn`
 * `multitask.rs/::Task::poll`: use `.map` instead of `match`

glommer:
Thanks for the cleanups!

glommer (approved):

Issue #31, opened 2020-09-07
Provide an ergonomic way to open a file given a directory


glommer:
Linux has an `open_at` syscall that opens a file given a directory.
The reason we don't have the same yet is that `open_at` overloads its meaning: if you pass an absolute path the dir argument is ignored.

That doesn't sit well with our level of abstraction: if we have an API that opens (or create) a file inside a directory we should make sure that the pathbuf does not contain directories and is just a file name.

The advantages of doing that include:
- We will be able to enhance error messages, as we'll always be able to tell the user which directory was involved in an open/create
- We will be able to store information easily about properties of a file (for instance block alignment, preferred I/O ring), and in the future device for scheduling, without the need to stat
- We will be able to use io_uring's registered fd support by pre-opening a bunch of file descriptors in that directory

glommer:
I think just having the API is a good first issue. Taking advantage of that to use more advanced features is a good follow up and will take more time. So adding both labels

vasantteja:
I would like to take look. Can I assign it to myself?

glommer:
if you need anything let me know!

glommer:
hi @vasantteja , how is this going for you ?

vasantteja:
@glommer  sorry, I was traveling and I couldn't take a look at this. I am currently working on this. I will reach out to you if I have any doubts. Thanks for the patience.

jefftt:
hello, since this issue seemed to have staled, i'd be willing to take a stab at it. 

@glommer did you have any ideas on how the API should look? i came up with something pretty naive (taking RawFd as input for dir) but unsure if this is sufficient:

```
buffered_file.rs

    /// Opens a file at a given directory
    ///
    /// Note that `name` must just be the file name and not contain any directories
    pub async fn open_at<P: AsRef<Path>>(dir: RawFd, name: P) -> io::Result<BufferedFile> {
        unimplemented!();
    }
```

glommer:
hi @jefftt you would be most welcome.

The API has to be methods of `Directory`. The goal of this API would be to be ergonomic: you already have a directory object, then you create a file inside that directory.

So it would be something like:

```rust
directory.rs:

 pub async fn create_file<P: AsRef<Path>>(name: P) -> io::Result<DmaFile> {}
```

I don't think we should do this for buffered file. All of the advantages of this API that I listed in the description would not apply to buffered files (they have no alignment needs, no ring selection, etc)

[...]

PR #202, opened 2020-12-03
Ergonomic directory file APIs

jefftt:
### What does this PR do?

Provides APIs to create/open files under a given directory

### Motivation

Address #31 

### Related issues

See above

### Checklist

### DIFF
--- a/glommio/src/io/directory.rs
+++ b/glommio/src/io/directory.rs
@@ -68,6 +68,21 @@ impl Directory {
         Ok(Directory { file })
     }
 
+    /// Opens a file under this directory, returns a DMA file
+    ///
+    /// NOTE: Path must not contain directories and just be a file name
+    pub async fn open_file<P: AsRef<Path>>(&self, path: P) -> io::Result<DmaFile> {
+        if path.as_ref().components().count() > 1 {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "Path cannot contain directories",
+            ));
+        }
+
+        let path = self.file.path_required("open file")?.join(path.as_ref());
+        DmaFile::open(path).await
+    }
+
     /// Similar to create() in the standard library, but returns a DMA file
     pub fn sync_create<P: AsRef<Path>>(path: P) -> io::Result<Directory> {
         let path = path.as_ref().to_owned();
@@ -92,11 +107,7 @@ impl Directory {
     ///
     /// NOTE: Path must not contain directories and just be a file name
     pub async fn create_file<P: AsRef<Path>>(&self, path: P) -> io::Result<DmaFile> {
-        if path
-            .as_ref()
-            .to_string_lossy()
-            .contains(std::path::MAIN_SEPARATOR)
-        {
+        if path.as_ref().components().count() > 1 {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
                 "Path cannot contain directories",
--- a/glommio/src/io/directory.rs
+++ b/glommio/src/io/directory.rs
@@ -72,7 +72,7 @@ impl Directory {
     ///
     /// NOTE: Path must not contain directories and just be a file name
     pub async fn open_file<P: AsRef<Path>>(&self, path: P) -> io::Result<DmaFile> {
-        if contains_dir(&path) {
+        if contains_dir(path.as_ref()) {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
                 "Path cannot contain directories",
@@ -107,7 +107,7 @@ impl Directory {
     ///
     /// NOTE: Path must not contain directories and just be a file name
     pub async fn create_file<P: AsRef<Path>>(&self, path: P) -> io::Result<DmaFile> {
-        if contains_dir(&path) {
+        if contains_dir(path.as_ref()) {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
                 "Path cannot contain directories",
@@ -137,8 +137,8 @@ impl Directory {
     }
 }
 
-fn contains_dir<P: AsRef<Path>>(path: P) -> bool {
-    let mut iter = path.as_ref().components();
+fn contains_dir(path: &Path) -> bool {
+    let mut iter = path.components();
     match iter.next() {
         Some(std::path::Component::Normal(_)) => iter.next().is_some(),
         _ => true,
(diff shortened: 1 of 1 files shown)


glommio/src/io/directory.rs
```diff
@@ -87,6 +88,25 @@ impl Directory {
         Self::sync_open(&path)
     }
 
+    /// Creates a file under this directory, returns a DMA file
+    ///
+    /// NOTE: Path must not contain directories and just be a file name
+    pub async fn create_file<P: AsRef<Path>>(&self, path: P) -> io::Result<DmaFile> {
```
zserik, line 94: This should probably accept a `AsRef<OsStr>` instead of `AsRef<Path>`, as it only expects a single path component.

glommio/src/io/directory.rs
```diff
@@ -87,6 +88,25 @@ impl Directory {
...
+        if path
+            .as_ref()
+            .to_string_lossy()
```
zserik, line 97: It is not a good idea to create a `String` every time this is called.
If the argument is replaced with `AsRef<OsStr>`, `Path:new` would be called here, and I think it would be a good idea to call `.components().count() > 1`.

zserik (changes requested):

glommio/src/io/directory.rs
glommer: But doesn't this make it harder to validate if this is indeed a relative path ?

glommio/src/io/directory.rs
glommer: We could however create a `PathBuf`. We store the `PathBuf` in the `GlommioFile`, so we would be just creating it earlier. `open_at` would have to change so that it could take a `PathBuf` as well, but that's doable.

glommer: Ok, I looked at the actual patch now and I agree with @zserik .

The best way to validate this is with `.components()` as path without transforming it into anything: nor a `PathBuf` nor a `String`

I am kind of in the fence with `OsStr`. `Path` seems more inline with the rest of the File.

glommer (changes requested):
Great work, thanks!

2 changes I would like to see:
* There is no reason not to go the extra mile and do this for Open as well. It's essentially the same code.
* avoid the transformation to string as @zserik mentioned. We can use `path.components()` for this

glommio/src/io/directory.rs
jefftt: yeah makes sense, i wasn't sure if path `/` should be valid or not.

Thanks looking!

glommio/src/io/directory.rs
```diff
@@ -68,6 +68,21 @@ impl Directory {
         Ok(Directory { file })
     }
 
+    /// Opens a file under this directory, returns a DMA file
+    ///
+    /// NOTE: Path must not contain directories and just be a file name
+    pub async fn open_file<P: AsRef<Path>>(&self, path: P) -> io::Result<DmaFile> {
+        if path.as_ref().components().count() > 1 {
```
zserik, line 75: This error handling code should be split out into a non-async separate helper function, as this code is now present in at least 2 functions. Additionally, the code should check if the path contains *only* a single directory separator.

jefftt: 👍 i added a function that returns a `bool` but wasn't sure if the convention was to return `Result<()>` instead so the caller can use `?` operator on it, lmk if i should change it

glommio/src/io/directory.rs
```diff
@@ -105,3 +136,12 @@ impl Directory {
...
     }
 }
+
+fn contains_dir(path: &Path) -> bool {
+    let mut iter = path.components().peekable();
+    let is_file = match iter.peek() {
+        Some(std::path::Component::Normal(_)) => true,
+        _ => false,
+    };
+    iter.count() > 1 || !is_file
```
zserik, line 146: The `iter.count() > 1` check should be moved inside the appropriate branch of the `match` statement (`Some(std::path::Component::Normal(_))`). Then, the iterator doesn't need to be `peekable`, because if it wouldn't be peekable, only the `iter.count() > 1` check would change to `iter.count() > 0`, which is equivalent to `iter.next().is_some()`.

Code suggestion:
```rust
fn contains_dir(path: &Path) -> bool {
    let mut iter = path.components();
    match iter.next() {
        Some(std::path::Component::Normal(_)) => iter.next().is_some(),
        _ => true,
    }
}
```

I'm not completely sure if that suggestion is correct, this should be tested.

jefftt: yes that's much better, thanks for the suggestion! 

I'm not sure what the correct behavior should be in the case where path is `foo/`. the `components()` method would [normalizedthe trailing slash away](https://doc.rust-lang.org/std/path/struct.Path.html#method.components) and so this function would return `true`. is this what we want?

zserik: I would expect that the "openat" equivalent should act like:
1. `chdir()` to the directory is called.
2. `File::create` / `File::open` with the given Path is called. If this call would also normalize the trailing `/` away, then this behavoir is probably what we want. If not, then we must check for that, too.

glommer: We should not `chdir`
If we can open a file in the current directory that is represented by the object, fine. Otherwise this needs to fail.

The main thing that needs to be handle correctly is files with `/` in their name. The example you gave, `foo/` is not a valid file name.

To the best of my knowledge `/` is invalid in path names.

I fully expect `Path` to do the right thing here across OS btw: if something is a valid filename with quotes and all, and `Path::components` break it apart, I would consider it a bug.

zserik: I know. I wrote "should *act like*". It should handle that case the same way `std::fs::OpenOptions::open` acts when given a file name with trailing slash.

glommio/src/io/directory.rs
```diff
@@ -137,11 +137,10 @@ impl Directory {
...
 }
 
-fn contains_dir(path: &Path) -> bool {
-    let mut iter = path.components().peekable();
-    let is_file = match iter.peek() {
-        Some(std::path::Component::Normal(_)) => true,
-        _ => false,
-    };
-    iter.count() > 1 || !is_file
+fn contains_dir<P: AsRef<Path>>(path: P) -> bool {
```
zserik, line 140: I don't think that this internal API should be generic, as it has no real benefit, but the disadvantage of possible making compile times longer.

glommer:
A general comment:

In general I prefer good over perfect, because even if there are things to make better, if one truly cares they can be fixed later.

I am happy enough with the state of this and am therefore merging. 

@jefftt thanks a lot for your first contribution and welcome!

PR #205, opened 2020-12-04
channels: reduce code repetition

zserik:
### What does this PR do?

This PR tries to reduce some repetition in the implementation of `channels` APIs.

### Motivation

Excess code repetition reduces the maintainablity of code.

### Checklist

No APIs were added, so no new unit tests are needed.
* [ ] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/src/channels/local_channel.rs
+++ b/glommio/src/channels/local_channel.rs
@@ -4,14 +4,9 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
 use crate::channels::{ChannelCapacity, ChannelError};
-use futures_lite::future;
-use futures_lite::stream::Stream;
-use std::cell::RefCell;
-use std::collections::VecDeque;
-use std::io;
-use std::pin::Pin;
-use std::rc::Rc;
+use futures_lite::{future, stream::Stream};
 use std::task::{Context, Poll, Waker};
+use std::{cell::RefCell, collections::VecDeque, io, pin::Pin, rc::Rc};
 
 #[derive(Debug)]
 /// Send endpoint to the `local_channel`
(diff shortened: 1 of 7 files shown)


glommio/src/channels/local_channel.rs
```diff
@@ -269,22 +294,26 @@ impl<T> LocalSender<T> {
...
+    }
+}
+
+fn wake_up_all(ws: &mut Option<VecDeque<Waker>>) {
+    if let Some(ref mut waiters) = ws {
+        // replace core::mem::replace with core::mem::take,
+        // as soon as we drop support for rustc < 1.40.0
+        // we assume here that wakes don't try to acquire a borrow on
+        // the channel.state
+        for w in core::mem::replace(waiters, VecDeque::new()) {
```
zserik, line 315: Do we still support `rustc < 1.40.0`? If not, I'll change this part of the code to use `core::mem::take`.

glommer: we do not.

glommer: For the future, I don't consider glommio API stable yet, and I believe I put a list of things I want to see happen before that in the docs website.

Until we declare the project API stable, I see no reason to ever support any version other than the current.

glommio/src/channels/local_channel.rs
```diff
@@ -295,37 +324,22 @@ impl<T> Drop for LocalSender<T> {
...
-            }
-        }
+        wake_up_all(&mut state.send_waiters);
     }
 }
 
 impl<T> Stream for LocalReceiver<T> {
     type Item = T;
 
+    #[inline]
```
glommer, line 340: Is this really necessary?

Usually I find it better to leave these decisions to the compiler, and have exceptions be justified.

I am not that versed in the behaviors of the rust compiler here. So if it is notoriously bad at inlining a function like this I am happy to go for it.

glommio/src/channels/shared_channel.rs
```diff
@@ -4,7 +4,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
 //
-use crate::channels::spsc_queue::{make, Consumer, Producer};
+use crate::channels::spsc_queue::{make, BufferHalf, Consumer, Producer};
```
glommer, line 7: I find it a bit unfortunate that we have to expose the `BufferHalf` thing to users of this, although this is all crate-local anyway so not that big of an issue.

I'll probably merge it as is

glommer:
Good enough for merge, taking it in.

Possible improvements are around not exposing the `BufferHalf` if you can think of a good way (maybe free floating functions)

glommio/src/channels/shared_channel.rs
zserik: I thought the same, but as it is only crate-local, it seems like a relatively clean solution, at it occured to me that this was the solution with the least friction.

glommio/src/channels/local_channel.rs
zserik: I try to only place `#[inline]` at functions/methods where I find it justified, which means that the function is:
1. really short
2. probably often called in loops or other "hot" sections
3. visible/exported outside of the crate (inside of the crate, it doesn't matter, I think the compiler maybe even ignores `#[inline]` if both caller and callee belong to the same crate, inside of the crate, only `#[inline(always)]` and `#[inline(never)]` *should* matter, I suppose).

glommer: All those criteria are good criteria on when to inline the function. The biggest question to me is whether rust wouldn't do it itself.

All in all, I don't really mind and I merged it as is. Thanks!

Issue #207, opened 2020-12-05
The `echo` example does not work


thirstycrow:
OS: Ubuntu 20.04
Kernel: 5.6.17-050617-generic

The `echo` example hangs after printing the first line. After retrying a dozen of times, it starts to panic at initialization. And every glommio application failed to start because of the same reason. Have to reboot to make them work again.

```
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
Server Listening on 127.0.0.1:10000 on 1 connections
^C
$ target/debug/examples/echo
thread 'server-0' panicked at 'cannot initialize I/O event notification: Os { code: 12, kind: Other, message: "Cannot allocate memory" }', glommio/src/parking.rs:249:48
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any', examples/echo.rs:97:26
$ target/debug/examples/hello_world
thread 'main' panicked at 'cannot initialize I/O event notification: Os { code: 12, kind: Other, message: "Cannot allocate memory" }', glommio/src/parking.rs:249:48
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
$ RUST_BACKTRACE=1 target/debug/examples/hello_world
thread 'main' panicked at 'cannot initialize I/O event notification: Os { code: 12, kind: Other, message: "Cannot allocate memory" }', glommio/src/parking.rs:249:48
stack backtrace:
   0: rust_begin_unwind
[...]

PR #209, opened 2020-12-07
Fix the `echo` example

zserik:
### What does this PR do?
This PR should fix an issue with the `echo` example, which seems to dead-lock itself in a race condition between multiple `accept` calls...

### Related issues

This PR is related to #207.

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -34,15 +34,9 @@ async fn server(conns: usize) {
 
     let mut servers = vec![];
     for _ in 0..conns {
-        let l = listener.clone();
+        let (stream, _) = listener.accept().await.unwrap();
         servers.push(Local::local(async move {
-            let (stream, _) = l.accept().await.unwrap();
-            loop {
-                let x = copy(&stream, &mut &stream).await.unwrap();
-                if x == 0 {
-                    break;
-                }
-            }
+            while 0 != copy(&stream, &mut &stream).await.unwrap() {}
         }));
     }
     join_all(servers).await;


examples/echo.rs
```diff
@@ -34,15 +34,9 @@ async fn server(conns: usize) {
 
     let mut servers = vec![];
     for _ in 0..conns {
-        let l = listener.clone();
+        let (stream, _) = listener.accept().await.unwrap();
         servers.push(Local::local(async move {
-            let (stream, _) = l.accept().await.unwrap();
```
glommer, line 39: Just so I understand:

this patch is doing two things:
- replacing a naked `loop` with a `while` loop.
- moving the accept call outside the Task.

Which one do you think is causing this issue?
Although I still have to do a final stabilization pass on the socket API as I said, I would like the code as is to work: IOW, we should be able to share a bound socket across many tasks and accept them locally.

zserik: I think that if `accept` is called in parallel, sometimes some of these calls are kind of "dropped" and the associated `accept().await` never returns.

zserik: The `loop` vs `while` thing shouldn't change behavoir at all, I just thought it would be a cleaner solution.

glommer: Is there any requirement from Linux that accept calls may be sequential?

From glommio's point of view indeed those calls are serial, but then can end up sequential in the same invocation of io_uring_submit.

zserik: As you have found out, this is because the other Waker's were simply dropped.

zserik: If we later replace `Async<TcpListener>` with an uring-specialized `TcpListener`, I think we should make it non-clonable to avoid the (Thundering Herd Problem)[https://en.wikipedia.org/wiki/Thundering_herd_problem] respectively waker ambiguity (e.g. should we wake *all* tasks waiting for the event, or only *one*, even if the "one one" might be the wrong one an the program might dead-lock if tasks depend on each other; I'm relatively sure that this possible via mixing of `futures::select` and channels....)

glommer:
This is superseded by https://github.com/DataDog/glommio/issues/214

Issue #194, opened 2020-11-30
Rust `Waker`s need to be `Send + Sync`


Matthias247:
It is currently not possible in Rust to safely define non thread-safe `Waker`s.  They are defined to be [Send](https://doc.rust-lang.org/std/task/struct.Waker.html#impl-Send) and [Sync](https://doc.rust-lang.org/std/task/struct.Waker.html#impl-Sync), and thereby the `RawWaker`s must fulfill those guarantees.

[This commit](https://github.com/DataDog/glommio/commit/26cd86c2a854ce875b1a6f79ffcdb8ab0d10bdab), as well as [the general Waker implementation for glommio tasks](https://github.com/DataDog/glommio/blob/035535a61d1d8e9dc3b64cf0044fb8d48a01354e/glommio/src/task/raw.rs#L180-L303) violate those properties.

Non thread-safe `Waker`s had been part of the earlier async/await design in form a so called `LocalWaker`. However they had been removed from the design in order to simplify it, which unfortunately makes life a bit harder (or slower) for projects like this.

In addition to this [Context is also `Sync`](https://github.com/rust-lang/rust/issues/66481), which is something that wasn't initially planned for and probably still should be fixed.

You might want to consider a different wakeup mechansism than directly referencing the tasks via non-thread safe pointers. [The original RFC](https://github.com/rust-lang/rfcs/blob/master/text/2592-futures.md#rationale-drawbacks-and-alternatives-to-the-wakeup-design-waker) mentioned an alternative implementation using thread-local storage and task IDs which could work. However there is also a drawback with this design: It would never allow for non-local wakeups, since the executor on the other thread would not be found (which might however be less of an issue for this project).

laa:
Hi @Matthias247 , thank you for such useful information both for this and the previous issue.  I also was worried by this misconception but did not have time yet to fully investigate this topic. I will definitely check it. That is not a top priority as for now, but IMHO has to be resolved.

laa:
@Matthias247 I have checked RFC which you linked. We actually already wake up tasks using local reference queue, though this probably is not so obvious from the code. As for executor id, I will do that shortly, so Wakers will not be possible to run on the foreign thread.

laa:
[...]

PR #210, opened 2020-12-07
Fix of issue #194. All changes in task status are guarded by the check of thread_id.

laa:
### What does this PR do?

Fix of issue #194. All changes in task status are guarded by the check of thread_id.

I have checked two implementations:
1. When Wakers are protected by executor_id.
2. When Wakers are protected by thread_id.

The last appeared to be more lightweight and personally, I think that it is less fragile from the point of view of the usage of API.
For both cases, I did not notice a noticeable slowdown in the execution of benchmarks.

To improve performance thread_id detection is wrapped in call of thread_local. Even if now there are some questions
to its performance, there is an ongoing issue which should solve them.

### Motivation

Satisfy the requirements of Waker's public API.

### Checklist

### DIFF
--- a/glommio/src/task/raw.rs
+++ b/glommio/src/task/raw.rs
@@ -14,6 +14,9 @@ use crate::task::header::Header;
 use crate::task::state::*;
 use crate::task::utils::{abort, abort_on_panic, extend};
 use crate::task::Task;
+use std::thread::ThreadId;
+
+thread_local!(static THREAD_ID : ThreadId = std::thread::current().id());
 
 /// The vtable for a task.
 pub(crate) struct TaskVTable {
@@ -110,6 +113,7 @@ where
 
             // Write the header as the first field of the task.
             (raw.header as *mut Header).write(Header {
+                thread_id: Self::thread_id(),
                 state: SCHEDULED | HANDLE | REFERENCE,
                 awaiter: None,
                 vtable: &TaskVTable {
@@ -130,6 +134,12 @@ where
 
             raw_task
         }
+    }
+
+    fn thread_id() -> ThreadId {
+        THREAD_ID
+            .try_with(|id| *id)
+            .unwrap_or_else(|_e| std::thread::current().id())
     }
 
     /// Creates a `RawTask` from a raw task pointer.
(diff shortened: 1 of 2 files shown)


glommio/src/task/raw.rs
```diff
@@ -14,6 +14,9 @@ use crate::task::header::Header;
 use crate::task::state::*;
 use crate::task::utils::{abort, abort_on_panic, extend};
 use crate::task::Task;
+use std::thread::ThreadId;
+
+thread_local!(static THREAD_ID : ThreadId = std::thread::current().id());
```
glommer, line 19: I don't understand from your comment why do you think this is beneficial as a TLS variable.

I already expect `std::thread::current()` to access a TLS variable to store its current pointer. Is that not the case?

Once you have the current thread info pointer, the id is just an u64 according to https://doc.rust-lang.org/src/std/thread/mod.rs.html#970

Creating the ID requires a bunch of atomics, but that's ok. Accessing it seems like fast.

laa: Hi @glommer here is the two simple benches which compare the performance of both approaches https://gist.github.com/laa/d71cb4bba211b1ecfc52bed0af3c91f6. 
In nutshell access through  std::thread::current() takes 10 ns, but access through try_with takes 0 ns. (obviously it takes some but it needs to be measure with higher precision).  I can add as the comment to the code. Sorry for not doing that it seems to be clear for me. But as I can see it is not clear enough :-)

glommer: It is not your fault. This is really puzzling!

I wonder why this is. Both options should be essentially a TLS access with an offset.

Do you have any explanation? I am ok with merging this - a comment close to the call site would definitely help.
But I wonder which code are both versions  generating?

laa: AFAIK, can not look at the code right now. Thread is protected by Arc which is cloned on the access of the thread. That is heavier obviously then getting id from the TLS. I will benchmark Arc performance out of curiosity, but unlikely I will get 10 ns. overhead, but who knows. Except for that, I do not have an explanation.  Maybe call overhead not 10 ns but less in the long run, for run15 min for example. The current bench is running about 2-3 min, which IMHO still long enough. But the difference is visible at least on my server.

laa: I will add a comment a bit later, but interesting that I always get the same consistent results several runs. So that is not a problem of tasks which are running at the moment on the server. Which was idle actually.

glommer: If the access to the current thread is protected by Arc that explains it.
But that's really surprising an unfortunate! =(

I'll merge as is and we can add a comment later.

PR #212, opened 2020-12-07
improve `executor.rs`

zserik:
### What does this PR do?

* flatten a chain of nested, convoluted `if` checks
* remove a redundant `Yield` implementation

### DIFF
--- a/glommio/src/executor.rs
+++ b/glommio/src/executor.rs
@@ -874,29 +874,24 @@ impl LocalExecutor {
                 // the opportunity to install the timer.
                 let duration = self.preempt_timer_duration();
                 self.parker.poll_io(duration);
-                if !self.run_task_queues() {
+                if self.run_task_queues() {
+                    spin_since = None
+                } else if let Poll::Ready(t) = future.as_mut().poll(cx) {
                     // It may be that we just became ready now that the task queue
                     // is exhausted. But if we sleep (park) we'll never know so we
                     // test again here. We can't test *just* here because the main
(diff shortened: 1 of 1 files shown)


glommio/src/executor.rs
```diff
@@ -1050,32 +1050,9 @@ impl<T> Task<T> {
             }
         });
 
-        if !need_yield {
-            return;
+        if need_yield {
+            futures_lite::future::yield_now().await;
```
glommer, line 1054: thank you!! When I found out about `yield_now` I figured it would do the same as this, and put it on my backburner to go back and verify.

Never did, though.

glommer (approved):

glommer:
hi @zserik -> thanks for this. I agree that the conditionals on spin were really hard to read. It unfortunately conflicted with the latest executor merge and I'd prefer not to touch that one.
It's otherwise approved, so as soon as you resolve the conflict I'll merge

Thanks again!

zserik:
I wasn't able to run the full testsuite against the rebased version because some tests are very timing/latency-sensitive, and thus stop working if my development system uses 100% cpu with load 5 (I'm currently upgrading incl. compiling thunderbird, that takes a while).

glommer:
yeah, I invested quite some work in making these tests robust against load, but there is only so much we can do for the scheduler tests...

I just checked CircleCI page and then still don't seem to have ubuntu 20.10 VM images =(

I tested this locally and it is all good

PR #218, opened 2020-12-11
Fix of order of semaphore waiting queue.

laa:
### What does this PR do?

Before the fix waiting queue was implemented as LIFO instead of FIFO.
Additional code clean up was performed.
Semaphore memory safety was ensured on close.

### Checklist

### DIFF
--- a/glommio/src/sync/semaphore.rs
+++ b/glommio/src/sync/semaphore.rs
@@ -65,7 +65,7 @@ impl Waiter {
             return;
         }
 
-        sem_state.waiters_list.push_front(self.node.clone());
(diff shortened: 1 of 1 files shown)


glommer:
@laa we should add a test of ordering.
Ordering is indeed important for the semaphore and this was a silent break.

The code itself is good, but please add a commit on top with a test that passes now but would fail without the fix

laa:
Sure, will do a bit later.

glommer:
Since this is currently broken I'll merge the current PR.

the test is still welcome but it can come in a separate one

laa:
@glommer thank you. I will do it for sure.

Issue #195, opened 2020-11-30
Consider using futures-intrusive


Matthias247:
This crate contains yet another implementation of async channels and semaphores. The implementation here comes with the downsides of the earlier `futures`-like implementation that an [unbounded list of waiting tasks needs to be stored in the sync primitives](https://github.com/DataDog/glommio/blob/035535a61d1d8e9dc3b64cf0044fb8d48a01354e/glommio/src/sync/semaphore.rs#L47-L48).  This has 2 downsides:
- A heap allocation is necessary
- This list might grow in some high-load situations, and might never be shrunk later on, leading to memory leakage

[futures-intrusive](https://docs.rs/futures-intrusive/0.4.0/futures_intrusive/) demonstrated how to build async sync primitives which use intrusive linkage instead of heap allocated task lists, and thereby doesn't require any dynamic memory allocation. The same approach was now also picked up by the tokio sync primitives.

You migth be interested in doing the same. And btw: `futures-intrusive` also offers non-`Sync` versions of those primitives which do not require any "real" OS mutex, e.g.
- [LocalSemaphore](https://docs.rs/futures-intrusive/0.4.0/futures_intrusive/sync/type.LocalSemaphore.html)
- [LocalChannel](https://docs.rs/futures-intrusive/0.4.0/futures_intrusive/channel/type.LocalChannel.html)

Those could be helpful to you out of the box

laa:
Hi @Matthias247 . Thank you very much for the tip. I will look on it.

glommer:
In particular I am *very* unhappy with the fact that we need an allocation for the `Source` of the io_uring operation and was thinking yesterday how we can make that better.

laa:
I have checked the code (not in all details though) and I suppose I will borrow the concept, not the library itself. There are a couple of reasons:
1.  Glommio has bigger a set of primitives then `futures-intrusive` and some more will be added.
2.   'futures-intrusive' and `glommio` have different design goals which intersect only partially. That may lead to different and contradictions design decisions. 
3. Because synchronization primitives are in the core of the design of `glommio` I prefer to have as more control as possible on such part of code.

But the overall idea is very good, I do like it. It will take a couple of weeks (2-3) for me to implement it.

laa:
@glommer sorry but I suppose this issue should be reoponed because it is not completely fulfilled. Rc type is boxed, so there are still some allocations once futures are paused. I will try to provide Rc-free versions in 2-3 days.

PR #219, opened 2020-12-11
 local_channel implementation was changed to use intrusive linked lists

laa:
Intrusive linked lists are used as an implementation of a waiting list and as the implementation of container for the items.

### What does this PR do?
 Fix of the issue #195. Final part.

 Heap allocations inevitable to happen inside of the collections
 were removed.
 Because intrusive collections work only with containers all nodes
 are wrapped to Rc (instead of Box, because that obviously will
 break the point of this change) which adds a bit of noise
 to implementation.

### Motivation

Reduce the number of heap allocations.

### Related issues

for this Pull Request.

### Additional Notes

Issue #195

### Checklist


glommer (approved):
This was easier for me to follow than the other two, perhaps because the channel code was fresher in my head.

Great work

Issue #220, opened 2020-12-12
Unresolved doc links


zserik:
I ran `cargo doc` against `glommio` and it returned 5 warnings:
```
warning: unresolved link to `Task::spawn_into`
    --> glommio/src/executor.rs:1136:25
     |
1136 |     /// [`spawn_into`]: Task::spawn_into
     |                         ^^^^^^^^^^^^^^^^ the struct `Task` has no field or associated item named `spawn_into`
     |
     = note: `#[warn(broken_intra_doc_links)]` on by default

warning: unresolved link to `Task::create_task_queue`
  --> glommio/src/shares.rs:18:20
   |
18 | /// [`TaskQueue`]: Task::create_task_queue
   |                    ^^^^^^^^^^^^^^^^^^^^^^^ no item named `Task` in scope

warning: unresolved link to `Task::create_task_queue`
  --> glommio/src/shares.rs:22:24
   |
22 |     /// [`TaskQueue`]: Task::create_task_queue
   |                        ^^^^^^^^^^^^^^^^^^^^^^^ no item named `Task` in scope

warning: unresolved link to `Task::create_task_queue`
  --> glommio/src/shares.rs:27:24
   |
27 |     /// [`TaskQueue`]: Task::create_task_queue
   |                        ^^^^^^^^^^^^^^^^^^^^^^^ no item named `Task` in scope

warning: unresolved link to `Task::create_task_queue`
  --> glommio/src/shares.rs:71:20
   |
71 | /// [`TaskQueue`]: Task::create_task_queue
   |                    ^^^^^^^^^^^^^^^^^^^^^^^ no item named `Task` in scope

warning: 5 warnings emitted
```

glommer:
Thanks. Do you know if there is a way to make those fatal? I would like to catch those at CI time

PR #226, opened 2020-12-12
doc: make unresolved links fatal

zserik:
### What does this PR do?

This PR makes the rustdoc warning "unresolved links" fatal.

### Motivation

https://github.com/DataDog/glommio/issues/220#issuecomment-743758354


zserik:
The "cargodoc" CI job fails. This is expected, it means that it worked.

glommer:
the fixes should be already in for those problems, though,
Can you merge your current branch with master ? this should fix the CI issue so we can merge green

Issue #180, opened 2020-11-19
Implement error system similar too the approach implemented in io::Error structure


laa:
This issue was discussed in community chat. Main issues with io::Error are:

1. Mapping between errors that happened inside the glommio library and presented in io::Errror is fuzzy and can not precisely describe the root cause of the issue.
2. io::Error lacks useful information as for example the path of the file where error happens. The error system should provide means to expose such information without allocation overhead for no-error situations. In general, error types should be very thin and do not request a substantial amount of memory.

Despite the visible simplicity of the task, it requires deep investigation of current implementations of error handling in Rust and a good understanding of a possible set of errors which could happen (will happen) inside of the library.

laa:
Interesting reading in the context of this issue https://blog.rust-lang.org/inside-rust/2020/11/23/What-the-error-handling-project-group-is-working-on.html

bryandmc:
I personally think a good pattern for a library like this could involve using `thiserror` from dtolnay that makes creating ergonomic, library errors much easier. Basically it lets us define one error type for this library (so that users of this library can work with a single type) and multiple enum variants for all the different error cases. It also makes it easier to encapsulate the `source` error inside of a particular variant (so the user can introspect what the actual cause of the error was). The thing that `thiserror` gives, is easy From/display implementations based on a proc macro. 

This seems to be the most idiomatic way to create library errors at the moment. I can provide an example of what this looks like, if anyone is interested. I actually used to use a hacky declarative macro to do basically the same thing before finding this crate, so this doesn't involve some sort of custom dynamic error type or anything like that. It's just an enum at the end of the day with some generated impl's.

Thoughts? It's pretty easy to implement and I'd be happy to mock up an example using whatever errors we currently have as the starting point.

glommer:
hello @bryandmc 

That looks great to me. My main concern is that I'd like to make sure that this interops well with `io::Error`, because that is everywhere : one good example is the Async I/O traits.

That means not only being able to convert, but keeping meaningful messages and causes whenever possible.
In other words, something that resembles a "Bad File" in Glommio should translate to "Bad File" in `io::Error`, and not just going the lazy route of marshalling everything under `Custom`

bryandmc:
@glommer That shouldn't be a problem. It can preserve and entire chain of "source" errors. Meaning it can hold the `io::Error` that caused the error in the first place, along with other information. I'll take a look at the current code base and see how it'll fit in. That said, I think we can accomplish your goals with this pretty easily.

[...]

PR #227, opened 2020-12-14
Addresses #180: Implement error system

bryandmc:
Here's my first stab at changing our error types to a single, composite error enum. 

### What does this PR do?

This implements a 'composite' error type to encompass all the errors glommio can produce. This is more convenient for users of the library, than to have many different types of errors that might need to be handled.

### Motivation

Issue #180 made a request for an error type similar to the `std::io::Error` type with many different `ErrorKind`'s. I suggested using `thiserror` because it got rid of some of the boilerplate of creating error enums without imposing weird restrictions such as alternative 'Error' traits (ex. `Failure` crate). 

Note: We could also change some of our custom error types over to just using this directly, instead of only wrapping our current error types. This might be a worthwhile simplification, but I can understand the other side as well so I left them be for now. This means the current errors will still exist, they will just be wrapped up into this type on any public API's.

### Related issues

#180 

### Additional Notes

YES! This pull request does not change over all the error types (yet). That's because I want to get feedback before going too far down that road. I did change over the Semaphore errors over to the new type as an example. If this method seems OK, I will modify this pull request (or do another) to complete the transition by moving all the current errors to this new `GlommioError` type.

### Checklist

[ X ] I have added unit tests to the code I am submitting
[ n/a ] My unit tests cover both failure and success scenarios
[ X ] If applicable, I have discussed my architecture
[ X ] The new code I am adding is formatted using `rustfmt`


glommio/src/error.rs
```diff
@@ -47,6 +98,17 @@ mod test {
     use super::*;
     use std::io;
 
+    #[test]
```
glommer, line 123: I would like to see tests making sure that we can convert easily to io::Error too.

For example, unless I am missing something it doesn't seem to me that this is providing a translation from the semaphore closed error to an io::Error / BrokenPipe.

glommio/src/error.rs
```diff
@@ -3,9 +3,60 @@
...
+    IoError(#[from] std::io::Error),
+
+    #[error("Lock closed error: {0}")]
+    LockClosedError(#[from] sync::LockClosedError),
+
+    #[error("Channel error: {0}")]
+    ChannelError(#[from] channels::ChannelError<T>),
+
+    #[error("Queue not found error: {0}")]
+    QueueNotFoundError(#[from] executor::QueueNotFoundError),
```
glommer, line 60: Isn't it cleaner to get rid of the original error and only use the user-visible error ?
if I understand this crate correctly this should be possible by adding a parameter to QueueNotFoundError

glommio/src/sync/semaphore.rs
```diff
@@ -359,9 +362,16 @@ impl Semaphore {
     /// This method does not suspend.
     ///
     /// # Errors
-    ///  If semaphore is closed `Err(io::Error(BrokenPipe))` will be returned.
+    ///  If semaphore is closed `Err(GlommioError::SemaphoreBroken)` will be returned.
```
glommer, line 365: SemaphoreClosed ?

glommio/src/error.rs
```diff
@@ -3,9 +3,60 @@
...
+/// Composite error type to encompass all error types glommio produces.
+pub enum GlommioError<T: Debug + 'static> {
+    #[error("IO error occurred: {0}")]
+    IoError(#[from] std::io::Error),
+
+    #[error("Lock closed error: {0}")]
+    LockClosedError(#[from] sync::LockClosedError),
+
+    #[error("Channel error: {0}")]
+    ChannelError(#[from] channels::ChannelError<T>),
```
glommer, line 56: I would prefer no to have a ChannelError.

What is special about a channel? The errors it can return are WouldBlock / BrokenPipe. That's very similar between a channel and a semaphore.

It seems to me that is is better to have just WouldBlock / BrokenPipe kinds of errors, and specify the source in the error parameters so the user knows where it comes from

glommer (changes requested):

glommer:
I left some comments on specific details on the implementation but ACK on the direction.

Thanks!

glommio/src/error.rs
bryandmc: Oh -- do you need that? I assumed it was just a made up error picked in that case for convenience. Largely you are only meant to go from io::Error -> GlommioError but it'll also be the cause of an error so it could be pulled out there. But is there need to go BACK to io::Error types?

glommio/src/error.rs
bryandmc: Yeah, I mention that somewhere in the description -- I agree that it would be cleaner.

glommio/src/sync/semaphore.rs
bryandmc: Yeah sounds good.

glommio/src/error.rs
bryandmc: I think one reason for ChannelError is returning the actual value in the case of an error so the user can do something with it after.. All that said, it would be nice if we didn't have to deal with it, tbh. In fact it's the reason we have an annoying type parameter is because of the channel error.

glommer: You are right, I overlooked that.

I will think about it a little bit more. I guess ChannelError is still generic enough because it serves all types of channel.

Another option is maybe to have those errors always returning a type, and `Semaphore` can use them but return `()`. Depends on how they would look like.

glommio/src/error.rs
glommer: sorry, missed that.
That's great!

glommio/src/error.rs
glommer: I think it is fundamental to be able to convert from glommio errors to io::Errors.
As I said before my main oposition to even having our own error is that every single other thing out there that deals with I/O (network and storage) tends to use `io::Error`. If we can't convert automatically, you lose the ability to do `?`, and the implementation of `From` usually needs to be crate local.

Once we accept that we have to implement `From` to convert between glommio errors to `io::Error` for interoperability, then there are two choices: we either stash everything into `io::Error::Other`, or we try our best to keep meaningful error kinds whenever possible.

I heavily prefer the latter, because it is easier for users to figure out what is going on.

Nothing of this is set of stone: if you see it differently by all means voice it.
I want to make it clear that the reason I am pushing for this is the use case where we are using, let's say, `hyper` with `glommio`: `hyper` uses `io::Error` everywhere. And not only that, but they are certainly testing for certain classes of error to see if a socket is disconnected, etc. If we are stashing everything into `Other`, aren't we making the integration harder?

bryandmc:
This is good feedback. Let me take that and see what I can do. The only sticking point would be the conversion to io::Error types.. It's perfectly possible, but only for ones that started out as an io::Error variant. I can show how you would get it out (I think it's just the `error_value.cause()` value). Now, if you want certain ones of the other errors to also convert to a specific io::Error  (BrokenPipe?) type, we can make custom code for that.. But it's not an included ability of this crate or anything like that. I guess I need to understand what your use case is for that to get a better idea of what you mean.

glommer:
I think the best way is to just implement `From` to allow to go from `GlommioError` to `io::Error`.
The ones that originated from `io::Error` will just be a simple conversion (and those are the ones I am most concerned about anyway).
The other ones we just do it manually for the ones that make sense (for instance, a closed channel is `BrokenPipe`, a full channel is `WouldBlock`, same as now) and for the ones that don't have an unambiguous conversion to an `ErrorKind`, we use `Other`

bryandmc:
FYI I have updated this. I can't promise I've actually migrated every last one of the public-facing error's to the GlommioError (though I took a whack at getting most of them). But I think this gives a good framework for the future, breaking down the error types into a few different areas, some of which are lightly used now, but can be integrated more deeply over time. 

Currently there's a lot of `into(..)` being used but in some places they could just be directly constructed as GlommioError's and I didn't tackle every one of those.

glommio/src/error.rs
```diff
@@ -24,66 +26,123 @@ pub enum ResourceType {
...
+                    requested, available
+                )),
+                ResourceType::RwLock => f.write_str("RwLock {{ .. }}"),
+                ResourceType::Channel(_) => f.write_str("Channel {{ .. }}"),
+            },
+        }
+    }
+}
+
+/// TODO: finish
```
glommer: The ones you added are coverage enough.
We should just add a catch-all statement after that and convert anything else to `Other`

glommio/src/error.rs
```diff
@@ -17,40 +17,75 @@ pub type Result<T, V> = std::result::Result<T, GlommioError<V>>;
...
+/// fn will_error() -> Result<(), GlommioError<()>> {
+///     Err(GlommioError::WouldBlock(ResourceType::File {
+///         msg: "Error reading a file".to_string(),
+///     }))?
+/// }
+/// assert!(will_error().is_err());
+///
+/// ```
+#[derive(Error)]
+#[non_exhaustive]
```
glommer, line 111: why are we marking this as non-exhaustive? I think it is healthy to have a catch-all statement in most error-related issues

glommer (commented):
I didn't review it all and unfortunately this will have to wait until 2021 (Have a happy new year!)

I did try to play with it a bit, though, and update our project using glommio to use your branch.

The biggest issues was that whereas I had this:
```
fn something() -> io::Result<()> {
  some_glommio_fn()
}
```

I had to change to:
```
fn something() -> io::Result<()> {
  some_glommio_fn()?;
  Ok(())
}
```

IOW, there is no automatic conversion until you `?`. I think this is totally ok.

From the little I read my biggest concern is with the error being non-exhaustive.

glommio/src/error.rs
bryandmc: So actually these can go.. but may I make a spirited defense of them if we ever release a 1.0 or this gains massive popularity.. In the ecosystem this is used to save yourself from making what would be considered a breaking change to add an error variant. That all said I am fine with removing it at the stage we are at... though we may want to consider it in the future if we start running into stronger stability requirements from the community. The rust community is maybe a little too obsessed with Sem-ver ;-).. for now going to remove.

bryandmc:
Yeah I don't know how much can be done regarding needing to use `?` and an `Ok(())` but what I did in a lot of places instead was to just do:

```
fn something() -> io::Result<()> {
  some_glommio_fn().map_err(Into::into)
}
```
It's not actually less typing I don't think but it sometimes seemed aesthetically better. I think in the end, it may make more sense to have the GlommioError's start deeper into the codebase or being OK with code using glommio to have the glommio error type instead of the io::Error one but I don't know your use case specifically. 

Will remove the non_exhaustives and take your time on the rest. Have a good holiday/new year as well.

glommer:
> Yeah I don't know how much can be done regarding needing to use `?` and an `Ok(())` but what I did in a lot of places instead was to just do:

This is perfectly ok. What I meant to imply was that the conversion was smooth: the only thing I had to do was this (which is perfectly ok).

Hope you had a great new year, Bryan.

There is a conflict now in the executor.
When you fix the conflict I'll merge

bryandmc:
Should be good to go now.

PR #228, opened 2020-12-15
Rc containers were removed from semaphore implementation.

laa:
### What does this PR do?

Because Rc is a boxed type, Rc containers which were used inside of semaphore implementation of the waiting queue

1. Futures lifetimes were restricted by a lifetime of the Semaphore.
 That is quite a minor restriction because that is typically how sync primitives are used anyway.
3. On Drop of the semaphore assertion that waiting queue is empty was added.
4. Permits also were bound to the Semaphore lifetime.

Also, test to ensure the order of execution of waiting tasks was added.

Implementation was check by ASAN.

### Motivation

Remove all heap allocations from semaphore implementation.

### Related issues

Issue #195

### Additional Notes

Unfortunately, ASAN reports that the Reactor is leaking (on code without any changes) so I bound to compare results of the execution of tests before and after the change.

Before change:
SUMMARY: AddressSanitizer: 671739684 byte(s) leaked in 1056 allocation(s).

After change:
SUMMARY: AddressSanitizer: 671739684 byte(s) leaked in 1056 allocation(s).

Leak reported by ASAN:

```
   #0 0x563be1e254fd in malloc /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:145:3
    #1 0x563be49d8bd5 in std::sys::unix::alloc::_$LT$impl$u20$core..alloc..global..GlobalAlloc$u20$for$u20$std..alloc..System$GT$::alloc::h4094fbb2da8d8ff0 /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/unix/alloc.rs:14:13
    #2 0x563be49cc6c1 in __rdl_alloc /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/alloc.rs:356:13
    #3 0x563be4009afe in alloc::alloc::alloc::hf3853c4e036525f8 /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:86:14
    #4 0x563be4009e61 in alloc::alloc::Global::alloc_impl::hd79981f77dd5de27 /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:166:73
    #5 0x563be400b724 in _$LT$alloc..alloc..Global$u20$as$u20$core..alloc..Allocator$GT$::allocate::h002ccb1a0bdb6212 /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:226:9
    #6 0x563be40097ea in alloc::alloc::exchange_malloc::h8efbecb76da3dc18 /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:316:11
    #7 0x563be39b958e in alloc::sync::Arc$LT$T$GT$::new::hed85b5c301f28c4c /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/sync.rs:330:25
    #8 0x563be3b6c8e7 in glommio::sys::uring::Reactor::new::hc0d3a6d8fa7d6649 /home/andrey/glommio/glommio/src/sys/uring.rs:1036:29
    #9 0x563be2a10a85 in glommio::parking::Reactor::new::hb5c81574e074abb3 /home/andrey/glommio/glommio/src/parking.rs:250:19
    #10 0x563be417448c in glommio::executor::LocalExecutor::new::h05fdf8ae72f3f90a /home/andrey/glommio/glommio/src/executor.rs:621:30
    #11 0x563be417279a in glommio::executor::LocalExecutorBuilder::make::h34c7f7dd1a78a261 /home/andrey/glommio/glommio/src/executor.rs:460:13
    #12 0x563be41740d7 in glommio::executor::LocalExecutor::make_default::h2629e83df605e8fb /home/andrey/glommio/glommio/src/executor.rs:612:9
```

### Checklist

### DIFF
--- a/glommio/src/sync/semaphore.rs
+++ b/glommio/src/sync/semaphore.rs
@@ -543,7 +543,7 @@ mod test {
     use crate::timer::Timer;
     use crate::{enclose, Local, LocalExecutor};
 
-    use futures::future::join3;
+    use futures_lite::future::or;
     use std::cell::Cell;
     use std::rc::Rc;
     use std::time::{Duration, Instant};
@@ -820,7 +820,7 @@ mod test {
             })
             .detach();
 
-            join3(t3, t2, t1).await;
+            or(or(t1, t2), t3).await;
         });
     }
 }
--- a/glommio/src/sync/semaphore.rs
+++ b/glommio/src/sync/semaphore.rs
@@ -141,6 +141,17 @@ impl<'a> Waiter<'a> {
     }
 }
 
+impl<'a> Drop for Waiter<'a> {
+    fn drop(&mut self) {
+        if self.node.link.is_linked() {
+            //if node is linked that is for sure is pinned so that is safe
+            //to make it pinned directly
+            let waiter_node = unsafe { Pin::new_unchecked(&mut self.node) };
+            Self::remove_from_waiting_queue(waiter_node, &mut self.semaphore.state.borrow_mut())
+        }
+    }
+}
+
 impl<'a> Future for Waiter<'a> {
     type Output = Result<()>;
 
--- a/glommio/src/sync/semaphore.rs
+++ b/glommio/src/sync/semaphore.rs
@@ -7,49 +7,115 @@ use std::cell::RefCell;
 use std::future::Future;
 use std::io::{Error, ErrorKind, Result};
 use std::pin::Pin;
-use std::rc::Rc;
 use std::task::{Context, Poll, Waker};
 
-use intrusive_collections::intrusive_adapter;
+use intrusive_collections::linked_list::LinkOps;
+use intrusive_collections::{container_of, offset_of, Adapter, PointerOps};
 use intrusive_collections::{LinkedList, LinkedListLink};
+use std::marker::PhantomPinned;
+use std::ptr::NonNull;
 
 #[derive(Debug)]
-struct Waiter {
-    node: Rc<WaiterNode>,
-    sem_state: Rc<RefCell<SemaphoreState>>,
+struct Waiter<'a> {
+    node: WaiterNode,
+    semaphore: &'a Semaphore,
 }
 
 #[derive(Debug)]
 struct WaiterNode {
     link: LinkedListLink,
-    state: RefCell<WaiterState>,
-}
-
-#[derive(Debug)]
-struct WaiterState {
     units: u64,
-    waker: Option<Waker>,
+    waker: RefCell<Option<Waker>>,
+
+    //waiter node can not be Unpin so its pointer could be used inside of intrusive
+    //collection, it also can not outlive the container which is guaranteed by the
+    //Waiter lifetime bound to the Semaphore which is container of all Waiters.
+    _p: PhantomPinned,
 }
 
-intrusive_adapter!(WaiterAdapter = Rc<WaiterNode> : WaiterNode {link : LinkedListLink});
+struct WaiterPointerOps;
 
-impl Waiter {
-    fn new(units: u64, sem_state: Rc<RefCell<SemaphoreState>>) -> Waiter {
+unsafe impl PointerOps for WaiterPointerOps {
+    type Value = WaiterNode;
+    type Pointer = NonNull<WaiterNode>;
+
+    unsafe fn from_raw(&self, value: *const Self::Value) -> Self::Pointer {
+        NonNull::new_unchecked(value as *mut Self::Value)
+    }
+
+    fn into_raw(&self, ptr: Self::Pointer) -> *const Self::Value {
+        ptr.as_ptr() as *const Self::Value
+    }
+}
+
+struct WaiterAdapter {
+    pointers_ops: WaiterPointerOps,
+    link_ops: LinkOps,
+}
+
+impl WaiterAdapter {
+    fn new() -> Self {
+        WaiterAdapter {
+            pointers_ops: WaiterPointerOps,
+            link_ops: LinkOps,
+        }
+    }
+}
+
+unsafe impl Adapter for WaiterAdapter {
+    type LinkOps = LinkOps;
+    type PointerOps = WaiterPointerOps;
+
+    unsafe fn get_value(
+        &self,
+        link: <Self::LinkOps as intrusive_collections::LinkOps>::LinkPtr,
+    ) -> *const <Self::PointerOps as PointerOps>::Value {
+        container_of!(link.as_ptr(), WaiterNode, link)
+    }
+
+    unsafe fn get_link(
+        &self,
+        value: *const <Self::PointerOps as PointerOps>::Value,
+    ) -> <Self::LinkOps as intrusive_collections::LinkOps>::LinkPtr {
+        let ptr = (value as *const u8).add(offset_of!(WaiterNode, link));
+        core::ptr::NonNull::new_unchecked(ptr as *mut _)
+    }
+
+    fn link_ops(&self) -> &Self::LinkOps {
+        &self.link_ops
+    }
+
+    fn link_ops_mut(&mut self) -> &mut Self::LinkOps {
+        &mut self.link_ops
+    }
+
+    fn pointer_ops(&self) -> &Self::PointerOps {
+        &self.pointers_ops
+    }
+}
+
+impl<'a> Waiter<'a> {
+    fn new(units: u64, semaphore: &'a Semaphore) -> Waiter<'a> {
         Waiter {
-            sem_state,
-            node: Rc::new(WaiterNode {
-                link: LinkedListLink::default(),
-                state: RefCell::new(WaiterState { units, waker: None }),
-            }),
+            node: WaiterNode {
+                link: LinkedListLink::new(),
+                units,
+                waker: RefCell::new(None),
+                _p: PhantomPinned,
+            },
+            semaphore,
         }
     }
 
-    fn remove_from_waiting_queue(&self, sem_state: &mut SemaphoreState) {
-        if self.node.link.is_linked() {
+    fn remove_from_waiting_queue(
+        waiter_node: Pin<&mut WaiterNode>,
+        sem_state: &mut SemaphoreState,
+    ) {
+        if waiter_node.link.is_linked() {
             let mut cursor = unsafe {
                 sem_state
                     .waiters_list
-                    .cursor_mut_from_ptr(self.node.as_ref())
+                    .cursor_mut_from_ptr(Pin::into_inner_unchecked(waiter_node) as *const _)
             };
 
             if cursor.remove().is_none() {
@@ -58,35 +124,44 @@ impl Waiter {
         }
     }
 
-    fn register_in_waiting_queue(&self, sem_state: &mut SemaphoreState, waker: Waker) {
-        self.node.state.borrow_mut().waker = Some(waker);
+    fn register_in_waiting_queue(
+        waiter_node: Pin<&mut WaiterNode>,
+        sem_state: &mut SemaphoreState,
+        waker: Waker,
+    ) {
+        *waiter_node.waker.borrow_mut() = Some(waker);
 
-        if self.node.link.is_linked() {
+        if waiter_node.link.is_linked() {
             return;
         }
 
-        sem_state.waiters_list.push_back(self.node.clone());
+        sem_state.waiters_list.push_back(unsafe {
+            NonNull::new_unchecked(Pin::into_inner_unchecked(waiter_node) as *mut _)
+        });
     }
 }
 
-impl Future for Waiter {
+impl<'a> Future for Waiter<'a> {
     type Output = Result<()>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let mut sem_state = self.sem_state.borrow_mut();
+        let mut sem_state = self.semaphore.state.borrow_mut();
+        let future_mut = unsafe { self.get_unchecked_mut() };
 
-        let units = self.node.state.borrow().units;
+        let waiter_node = unsafe { Pin::new_unchecked(&mut future_mut.node) };
+
+        let units = waiter_node.units;
         match sem_state.try_acquire(units) {
             Err(x) => {
-                self.remove_from_waiting_queue(&mut sem_state);
+                Self::remove_from_waiting_queue(waiter_node, &mut sem_state);
                 Poll::Ready(Err(x))
             }
             Ok(true) => {
-                self.remove_from_waiting_queue(&mut sem_state);
+                Self::remove_from_waiting_queue(waiter_node, &mut sem_state);
                 Poll::Ready(Ok(()))
             }
             Ok(false) => {
-                self.register_in_waiting_queue(&mut sem_state, cx.waker().clone());
+                Self::register_in_waiting_queue(waiter_node, &mut sem_state, cx.waker().clone());
                 Poll::Pending
             }
         }
@@ -132,8 +207,8 @@ impl SemaphoreState {
         while !cursor.is_null() {
             let node = cursor.remove().unwrap();
 
-            let waker = node.state.borrow_mut().waker.take();
-            if let Some(waker) = waker {
+            let node = unsafe { &*node.as_ptr() };
+            if let Some(waker) = node.waker.borrow_mut().take() {
                 waker.wake();
             } else {
                 panic!("Future is linked into the waiting list without a waker");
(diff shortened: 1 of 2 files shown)


laa:
I have submitted change only to the Semaphore, but the rest of sync primitives will be changed once and if this PR will be approved.

glommer:
@laa can you share the asan command line you used? I will look at it later. Best to run the exact same thing as you did to be sure.

glommer:
I suspect we're leaking the registered buffers pool on destruction, but will confirm once I can reproduce

laa:
Hi @glommer I used following command `cargo +nightly test -Zbuild-std --target x86_64-unknown-linux-gnu` with env. flags `export RUSTFLAGS=-Zsanitizer=address RUSTDOCFLAGS=-Zsanitizer=address`

glommio/src/sync/semaphore.rs
```diff
@@ -447,13 +520,21 @@ impl Semaphore {
...
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
     use crate::timer::Timer;
     use crate::{enclose, Local, LocalExecutor};
 
+    use futures::future::join3;
```
glommer, line 535: I am trying to avoid adding `futures` dependencies, and standardize on `futures_lite` (although there is no need to rush to remove the existing ones).

Worst case if you don't find anything in `futures_lite` you can just await them all separately. Better than increasing the dependency on `futures`

glommio/src/sync/semaphore.rs
```diff
@@ -7,49 +7,115 @@ use std::cell::RefCell;
...
+impl WaiterAdapter {
+    fn new() -> Self {
+        WaiterAdapter {
+            pointers_ops: WaiterPointerOps,
+            link_ops: LinkOps,
+        }
+    }
+}
+
+unsafe impl Adapter for WaiterAdapter {
```
glommer, line 65: are those unsafe signatures mandated by the intrusive containers implementation ?

In general there is a lot of `unsafe` usage now which I definitely consider a disadvantage. I would like to spend some time to understand each of them better

glommer (changes requested):

glommer:
Ok, that's a though one.

It's similar to the `Rc`. leaks we had in the reactor, but this time we are leaking the `Arc` used for the eventfd.
The `Rc` leaks we fixed by passing `downgrade`d versions and upgrading prior to use. But this won't work here because the cost of always having to upgrade an `Arc` before use is prohibitive.

We can probably open a separate issue about that for tracking and will have to think of a good way to fix.

glommer:
There are other ones as well. This one will be the harder to fix, though

glommio/src/sync/semaphore.rs
laa: Hi @glommer , unsafe adapter trait is imposed by the intrusive collections. Unfortunately, all implementations which want to use pinning property of the Future and to avoid allocation require a good portion of unsafe code.

But if you check it, the scope of its usage fails in two categories only.
1. Usage of pinned mutable reference. It is always unsafe because you have to be sure that you do not break pinning guarantees. So you do not move the object which you use outside of Pin. That is why moved is wrapped into Pin and there are some islands of unsafe when mutable reference is needed.
2. Adapter which converts pointers from to pointer to the link and pointer to the field. (At previous version it was hidden by the macros)

All mutations are controlled in code using RefCell to do not break mutability guarantees. Only links to the next times are controlled by intrusive list. 

Also for your consideration bench of usage of semaphore before and after the changes (numbers are quite consistent):

before 

cost of acquiring uncontended semaphore: 14ns
cost of acquiring contended semaphore: 190ns

after

cost of acquiring uncontended semaphore: 12ns (-14%)
cost of acquiring contended semaphore: 164ns (-14%)

P.S. To ensure the code safety I will add the Drop implementation for the waiter to ensure that it is removed from the waiting list for sure. And will replace join3 using features_light.

laa:
>We can probably open a separate issue about that for tracking and will have to think of a good way to fix.
Sure I will be curious to investigate it.

glommio/src/sync/semaphore.rs
glommer: To be clear I am very pro using this, and if we have to mark them as unsafe , so be it.
The only change I am requiring at this time was about the `join`.

My comment on `unsafe` is just meant to set expectations: I may take a bit longer to merge this because I want to understand the usage a bit better (I'll ask questions if needed).

glommio/src/sync/semaphore.rs
laa: Fixed

glommio/src/sync/semaphore.rs
```diff
@@ -7,49 +7,115 @@ use std::cell::RefCell;
...
+unsafe impl PointerOps for WaiterPointerOps {
+    type Value = WaiterNode;
+    type Pointer = NonNull<WaiterNode>;
+
+    unsafe fn from_raw(&self, value: *const Self::Value) -> Self::Pointer {
+        NonNull::new_unchecked(value as *mut Self::Value)
+    }
+
+    fn into_raw(&self, ptr: Self::Pointer) -> *const Self::Value {
+        ptr.as_ptr() as *const Self::Value
```
glommer, line 47: why isn't a non-null check needed here?

It's true that this looks just like a cast to me, but at the same time it has a self parameter, which indicates that the user of this data structure may expect a transformation to happen. If a transformation could happen, could it be that a null pointer is ever passed?

Keep in mind that I am just being too paranoid with this but I want to understand the unsafe usage better

glommer: Is it because we define `NonNull` as our pointer type?

glommio/src/sync/semaphore.rs
```diff
@@ -7,49 +7,115 @@ use std::cell::RefCell;
...
+
+unsafe impl Adapter for WaiterAdapter {
+    type LinkOps = LinkOps;
+    type PointerOps = WaiterPointerOps;
+
+    unsafe fn get_value(
+        &self,
+        link: <Self::LinkOps as intrusive_collections::LinkOps>::LinkPtr,
+    ) -> *const <Self::PointerOps as PointerOps>::Value {
+        container_of!(link.as_ptr(), WaiterNode, link)
```
glommer, line 73: can link be null? This doesn't seem to be a `NonNull`, at least it doesn't appear anywhere obvious

glommio/src/sync/semaphore.rs
```diff
@@ -7,49 +7,115 @@ use std::cell::RefCell;
...
+    }
 
-impl Waiter {
-    fn new(units: u64, sem_state: Rc<RefCell<SemaphoreState>>) -> Waiter {
+    unsafe fn get_link(
+        &self,
+        value: *const <Self::PointerOps as PointerOps>::Value,
+    ) -> <Self::LinkOps as intrusive_collections::LinkOps>::LinkPtr {
+        let ptr = (value as *const u8).add(offset_of!(WaiterNode, link));
+        core::ptr::NonNull::new_unchecked(ptr as *mut _)
```
glommer, line 81: that doesn't seem effective:
if `Value` is null, then the final ptr will be Null + offset which will pass the non_null check but still be invalid.

Why don't we have to check the original ptr for non-null ?
While we're at it, doesn't using `new_unchecked` defeats the whole purpose of using `NonNull` ?

glommio/src/sync/semaphore.rs
```diff
@@ -7,49 +7,115 @@ use std::cell::RefCell;
...
+    fn remove_from_waiting_queue(
+        waiter_node: Pin<&mut WaiterNode>,
+        sem_state: &mut SemaphoreState,
+    ) {
+        if waiter_node.link.is_linked() {
             let mut cursor = unsafe {
                 sem_state
                     .waiters_list
-                    .cursor_mut_from_ptr(self.node.as_ref())
+                    .cursor_mut_from_ptr(Pin::into_inner_unchecked(waiter_node) as *const _)
```
glommer, line 118: I am a bit concerned with this.
Right now usage seems safe, but I worry about future changes.

The reason this seems safe to me is that when we use node, it is only at the time we are ready to destruct it.
But to do that, we do dereference it, and it isn't hard to imagine someone makinga mistake and moving out of it.

Could we implement our our `Deref` for those that just panic, and use a helper function to convert node to its internal components that we are interested at?

Something like `fn consume_node(node: *const WaiterNode) -> (Option<Waker>, u64)`?

So we would never deref -> always safe on the unpin, and still extract the stuff we want ourselves when time comes to use

glommer (changes requested):
hi @laa 

Unless I am missing something - in which case let me know - I think that we should at least be using stronger guarantee versions of `NonNull` by using `new` instead of `new_unchecked`.
I don't mind paying the branch cost for this.

Also think we're missing some non-null checks -> I left comments in the specific places.

The only thing that still makes me uneasy with this is the Unpin. I think we could do something to improve safety there

glommio/src/sync/semaphore.rs
laa: Hi @glommer , those pointers can not be null because this trait implementation is used by intrusive_collections to convert pointers from one form to another (pointer to the link field to the pointer to the value field and vice versa). So those methods always receive valid pointers.

glommio/src/sync/semaphore.rs
laa: Hi @glommer , those pointers can not be null because this trait implementation is used by intrusive_collections to convert pointers from one form to another (pointer to the link field to the pointer to the value field and vice versa). So those methods always receive valid pointers. Those pointers are not pointers to the NonNull container, those are pointers to the value contained into the collection and to the link inside of this value contained into collation.

laa: Also, that is a part of the contract that methods can not panic, and actually, they are useless with null pointers.

glommio/src/sync/semaphore.rs
laa: Hi @glommer , in their terminology, Pointer is what you pass into the intrusive collection. In our case that is NonNull, so we do not pass a raw pointer there but valid object.

glommio/src/sync/semaphore.rs
laa: Hi @glommer that is also NonNull object, so this LinkPtr can not be null.

glommer: Do they document those things anywhere?
I am okay with that, but would be nice to point to some stated claims from the crate we are using.

glommio/src/sync/semaphore.rs
laa: The container itself guarantees validness of the pointer but I suppose additional check will not harm.

glommio/src/sync/semaphore.rs
laa: Sure that is part of the  trait code https://github.com/Amanieu/intrusive-rs/blob/master/src/linked_list.rs#L149 . We iplement LinkOps (it is in above type) and their LinkPtr is as NonNull<Link>

glommio/src/sync/semaphore.rs
laa: Hi @glommer , we use a pointer to the node here, only to remove from the list we do not access waker. But I suppose you mean that when we access the internal object pointed by the pointer, we should deref it. To mitigate which risks I am a bit confused. There are two risks there to have invalid pointers:
1. Node is moved to other location of memory, so the pointer becomes invalid. That is mitigated because of the object itself  !Unpin and wrapped into the PIn object. Code is unsafe there but intrusive_collections do not move objects. So it is logically safe.
2. Future is dropped from the stack or heap, in such case pointer also becomes invalid. To prevent that we implement Drop trait which removes a node from the list, ensuring that invalid pointers are not stored into the collection.

Could you elaborate what risks do you think will be mitigated by the introduction of this method, because I think only about null checks but they can not happen because we create Pin-s from references?

laa: Also this method `cursor_mut_from_ptr` accepts pointer not ref, it only looked like it accepts ref because of implicit conversion.

laa:
Hi @glommer sure, I agree that NonNull better to create with `new` it is merely slipped my mind during the change.

glommio/src/sync/semaphore.rs
glommer: the `Pin` implementation, especially for things that don't implement `Unpin`, require objects to be not movable.
Once you do this: 

```
let node = unsafe { &*node.as_ptr() }; 
```

You now turned your pinned object into a normal object from the point of view of the compiler. It could be moved and `Pin` would break.

Your implementation looks fine to me, because we only do that when we are exhausting the list.
My concern is more with the future proofing of this.

What I suggested would be a way to make sure that we can never dereference that pointer directly: if we can't dereference it, we can't break the `Pin` guarantees. We still need to manipulate it once it is going away and that is what the helper function would be for.

laa: Also when we access object, we either access it through the collection, in such case, it performs conversion itself from Pointer (NonNull in our case, that is why we implement PtrOps, to tell it how to convert Pointer to Node) or when we remove it from the node. In such a case, NonNull is returned. In both cases inside PtrOps or inside of the semaphore itself null check is performed by usage of NonNull contract.

laa:
>The only thing that still makes me uneasy with this is the Unpin 
That is why WaiterNode is implemented as !Unpin to prevent situations when Unpin objects are expected.  Which of course would be dangerous.

glommio/src/sync/semaphore.rs
laa: For example https://github.com/Amanieu/intrusive-rs/blob/master/src/linked_list.rs#L1041 push method accepts pointer. 
Also in PoiterOps we declared Pointer https://github.com/DataDog/glommio/pull/228/files/46aa06bd3f1f9cf978d2fdf94645591c168ae0ff#diff-9063a13a75caca6f9b4f70da0df9795d578d990677ae6c5ec0f916a43181d6c6R40 as NonNull defining all subsequent operations of LinkedList as such which will accept NonNull.

glommer:
> > The only thing that still makes me uneasy with this is the Unpin
> 
> That is why WaiterNode is implemented as !Unpin to prevent situations when Unpin objects are expected. Which of course would be dangerous.

The problem is that they are `Unpin`, but then we use `into_inner_unchecked`. At that point we could violate the pin guarantee.

To be clear, I think the current implementation is safe because we only dereference the object from the pinned pointer when it is time to destroy it. But would like to make it safer.

I just thought that maybe instead of all that dance I suggested: Can we maybe transform the object back into a `Pin` before using it, with `Pin::new_unchecked` ?

glommio/src/sync/semaphore.rs
laa: LinkOps are defined there https://github.com/DataDog/glommio/pull/228/files/46aa06bd3f1f9cf978d2fdf94645591c168ae0ff#diff-9063a13a75caca6f9b4f70da0df9795d578d990677ae6c5ec0f916a43181d6c6R66 as result they determine that LinkPtr is NonNull<Link>

laa:
But they are `!Unpin` not `Unpin` that is guaranteed because there is `PhantomPinned` member inside of WaitingNode, that makes both node and Future !Unpin to prevent to move them without any guarantees. We call into_inner_unchecked only to pass pointer of the object to the intrusive collection which does not move objects. `into_inner_unchecked` is unsafe because it can not guarantee on compiler level that we will not move that pointer after conversion. But because collections do not move objects inside memory such usage is safe. Every time we call into_inner_uncheced it is already Pin, because this method only accept pinned objects.

glommer:
I mispoke and confused you, my bad. I get that they are `!Unpin`.
My problem is with this sentence:

```
into_inner_unchecked is unsafe because it can not guarantee on compiler level that we will not move that pointer after conversion. But because collections do not move objects inside memory such usage is safe.
```

I agree it is safe now. What I am looking for is for ways to make sure it keeps being safe in the future when we make changes to it. 

Because this is confusing, let me restate: I judge the code as is to be fine and I found no problems.

What I am looking for is ways to make it more future proof.
As an example, you derefence the `!Unpin` object back into a normal object with `&*node`. The usage you have of it is *now* safe but it becomes too easy to make a mistake in a patch to this later and end up moving the object.

Would it be safer if we used `Pin::new_unchecked` to move it back to its pinned state?

laa:
Ah got it, finally. Yes, that makes sense.

glommer:
It's my fault. I kept saying `Unpin` when I meant `!Unpin`.

if `Pin::new_unchecked` does not require deref we can implement a panic version of it. But I don't know if it does.

Also the null stuff is all fine but the extra checks would reduce our dependency of a future version of intrusive doing something weird. It's good to have. They may want those functions not to panic, but if they passed a null pointer that means they violated their contract with us. So we are okay to violate our contract with them.

zserik:
@glommer It would be a really bad idea to "just" implement a `Deref` which panics, as this could lead to introduction of errors into code without noticing it (at compile time, and maybe at runtime if it happens in an seldom used code branch, e.g. error handling/recovery), as the compiler wouldn't catch it; and you should take note that `Deref` is automatically called in a lot of places ([`Deref` coercion](https://doc.rust-lang.org/stable/core/ops/trait.Deref.html#more-on-deref-coercion))...
Can we just *not* implement `Deref`?

glommer:
That was just an idea. If we don't implement `Deref`, don't we get an automatic implementation ?

zserik:
@glommer about which structure exactly do we talk about? I don't think we *generally* get an automatic implementation.

laa:
We did not implement Deref which panics, we merely add some safety checks to ensure that pointers never null.  and that only pins are used.

zserik:
ok, that is good.

glommer:
I am happy with the current implementation.

Thanks Andrey

laa:
@glommer  thank you :-) RwLock in nutshell but the channel is a bit different, it is interesting that on non contended case performance of the local channel was improved a little bit more than 2 times :-) I have provided results of the benches.

glommer:
quick note: this failed the build because the semaphore test was still using `make_default`. 
I'll fix that, it's no big deal.

Speaking of that, take a look at https://github.com/DataDog/glommio/issues/241.

I never liked the fact that `run()`, because it runs a single future to completion doesn't run in a task queue: that exposes us to all sorts of issues.

Now that I understand Rust a bit more I think it would be doable to get rid of that special case. Any suggestions on how are welcome. We depend too much on things being in queues. When they are not, problems like that can easily arise.

PR #232, opened 2020-12-16
Rc containers were removed from RwLock implementation.

laa:
### What does this PR do?
Because Rc is a boxed type, Rc containers were used inside of
RwLock implementation of the waiting queue

1. Futures lifetimes were restricted by a lifetime of the RwLock.
 That is a quite minor restriction because
that is typically how sync primitives are used anyway.
3. On Drop of the RwLock assertion that waiting queue is empty
was added.
4. Guards also were bound to the RwLock lifetime.
5. On Drop Futures remove itself from the waiting queue.

Code was tested by ASAN.
### Motivation
Removal of heap allocations inside of sync primitives.

Because the first PR was generally good accepted I decided to submit another one.

### Related issues

Issue #195

### Checklist

### DIFF
--- a/glommio/src/sync/rwlock.rs
+++ b/glommio/src/sync/rwlock.rs
@@ -31,7 +31,6 @@
 //! });
 //! ```
 //!
-use alloc::rc::Rc;
 use core::fmt::Debug;
 use std::cell::{Ref, RefCell, RefMut};
 use std::error::Error;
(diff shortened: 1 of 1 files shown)


laa:
null pointer safety checks were added

Waiter node references converted into the Pin-ed whenever possible.

PR #234, opened 2020-12-16
Rc containers were removed from LocalChannel implementation.

laa:
### What does this PR do?

Rc containers were removed from LocalChannel implementation.

Because Rc is a boxed type, Rc containers which were used inside of 
LocalChannel implementation of the waiting queue 
1. Futures lifetimes were restricted by lifetime of the LocalChannel.
 That is quite a minor restriction because 
that is typically how sync primitives are used anyway.
3. On Drop of the LocalChannel assertion that waiting queues are empty 
was added.
4. On Drop Futures remove itself from waiting queue.
5. LocalChannel does not implement Stream itself but returns related Stream 
instance through LocalReceiver#stream() method. 
That makes usage less convenient but a) It is more canonical because single 
Stream should work with single Task instance the same as Future. 
b) It makes implementation simpler.
6. On Drop of stream associated waiting node is removed from the waiting queue. 

Results of benchmark

before 

cost of sending contended local channel: 137ns
cost of sending uncontended local channel: 40ns

after 

cost of sending contended local channel: 138ns 0%
cost of sending uncontended local channel: 17ns  -58%

Code is checked by ASAN.

### Motivation

Decrease heap allocations during the scheduling of tasks

### Related issues

Issue #195

### Checklist

### DIFF
--- a/glommio/benches/local_channel.rs
+++ b/glommio/benches/local_channel.rs
@@ -10,7 +10,7 @@ fn main() {
             let runs: u32 = 10_000_000;
             let (sender, receiver) = local_channel::new_bounded(1);
             let receiver = Local::local(async move {
-                receiver.fold(0, |acc, x| acc + x).await;
+                receiver.stream().fold(0, |acc, x| acc + x).await;
             })
             .detach();
 
@@ -34,7 +34,7 @@ fn main() {
             let runs: u32 = 10_000_000;
             let (sender, receiver) = local_channel::new_bounded(10_000_000);
             let receiver = Local::local(async move {
-                receiver.fold(0, |acc, x| acc + x).await;
+                receiver.stream().fold(0, |acc, x| acc + x).await;
             })
             .detach();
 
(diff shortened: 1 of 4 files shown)


laa:
null pointer safety checks were added

Waiter node references converted into the Pin-ed whenever possible.

glommer:
@laa 

clippy from beta is complaining about this now:

```
warning: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
   --> glommio/src/channels/local_channel.rs:732:5
    |
732 |     pub fn stream<'a>(&'a self) -> impl Stream<Item = T> + 'a {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(clippy::needless_lifetimes)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
```

I am already using the beta stream locally, but not mandating in CI yet.

glommer:
this is FYI. It's a trivial fix, so I will push a fix.

PR #240, opened 2020-12-17
Improve ReadResult

zserik:
### What does this PR do?

This PR tries to improve the `ReadResult` code, it introduces an `Deref` impl, moves the `read_at` code to `crate::io::ByteSliceExt::read_at` and introduces the function `ReadResult::slice_unchecked`, which omits the (out-of-)range checks in release mode.

### Motivation

https://glommio.zulipchat.com/#narrow/stream/262273-general/topic/Where.20to.20help.3F/near/220295672

### Related issues

None.

### Additional notes

It is probably a good idea to squash this before merging.

### Checklist

[X] The new code I am adding is formatted using `rustfmt`

### DIFF
--- a/glommio/src/io/byte_slice_ext.rs
+++ b/glommio/src/io/byte_slice_ext.rs
@@ -1,0 +1,125 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the
+// MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
+
+/// Utility methods for working with byte slices/buffers.
+pub trait ByteSliceExt {
+    /// Copies the contents of `self` into the byte-slice `destination`.
+    ///
+    /// The copy starts at position `offset` into the `source` and copies
+    /// as many bytes as possible until either we don't have more bytes to copy
+    /// or `destination` doesn't have more space to hold them.
+    ///
+    /// ## Returns
+    ///
+    /// The amount of bytes written.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use glommio::LocalExecutor;
+    /// use glommio::io::DmaFile;
+    /// use glommio::io::ByteSliceExt;
+    ///
+    /// let ex = LocalExecutor::default();
+    /// ex.run(async {
+    ///     let file = DmaFile::create("test.txt").await.unwrap();
+    ///     let buf = file.alloc_dma_buffer(4096);
+    ///     let mut vec = vec![0; 64];
+    ///     let n = buf.read_at(0, &mut vec);
+    ///     assert_eq!(n, 64); // read 64 bytes, the size of the buffer
+    ///
+    ///     let n = buf.read_at(4090, &mut vec);
+    ///     assert_eq!(n, 6);  // read 6 bytes, as there are only 6 bytes left from this offset
+    ///     file.close().await.unwrap();
+    /// });
+    /// ```
+    fn read_at<D: AsMut<[u8]> + ?Sized>(&self, offset: usize, destination: &mut D) -> usize;
+}
+
+/// Utility methods for working with mutable byte slices/buffers.
+pub trait ByteSliceMutExt {
+    /// Writes data to this buffer from a user-provided byte slice, starting at a particular
+    /// offset
+    ///
+    /// ## Returns
+    ///
+    /// The amount of bytes written.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use glommio::LocalExecutor;
+    /// use glommio::io::DmaFile;
+    /// use glommio::io::ByteSliceMutExt;
+    ///
+    /// let ex = LocalExecutor::default();
+    /// ex.run(async {
+    ///     let file = DmaFile::create("test.txt").await.unwrap();
+    ///     let mut buf = file.alloc_dma_buffer(4096);
+    ///     let vec = vec![1; 64];
+    ///     let n = buf.write_at(0, &vec);
+    ///     assert_eq!(n, 64); // wrote 64 bytes.
+    ///
+    ///     let n = buf.write_at(4090, &vec);
+    ///     assert_eq!(n, 6);  // wrote 6 bytes, as there are only 6 bytes left from this offset
+    ///     file.close().await.unwrap();
+    /// });
+    /// ```
+    fn write_at<S: AsRef<[u8]> + ?Sized>(&mut self, offset: usize, src: &S) -> usize;
+
+    /// Writes the specified value into all bytes of this buffer
+    fn memset(&mut self, value: u8);
+}
+
+#[inline]
+fn slice_read_at(source: &[u8], offset: usize, destination: &mut [u8]) -> usize {
+    if offset > source.len() {
+        0
+    } else {
+        let len = core::cmp::min(destination.len(), source.len() - offset);
+        destination[0..len].copy_from_slice(&source[offset..][..len]);
+        len
+    }
+}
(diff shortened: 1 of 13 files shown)


glommio/src/io/read_result.rs
```diff
@@ -2,104 +2,139 @@
...
+        self.inner.as_ref().map(|inner| &**inner).unwrap_or(&[])
+    }
+}
+
+#[derive(Clone, Debug)]
+struct ReadResultInner {
+    buffer: Rc<DmaBuffer>,
     offset: usize,
-    end: usize,
+    len: NonZeroUsize,
```
zserik, line 30: This (usage of `NonZeroUsize`) is probably needed to make sure that `rustc` doesn't need to reserve additional memory for the surrounding `Option` enum tag.

glommio/src/io/read_result.rs
```diff
@@ -2,104 +2,138 @@
...
-    pub fn read_at(&self, offset: usize, dst: &mut [u8]) -> usize {
-        let offset = self.offset + offset;
-        self.buffer.as_ref().unwrap().read_at(offset, dst)
+    /// Returns `None` if either offset or offset + len would not fit in the original buffer.
+    pub fn slice(this: &Self, extra_offset: usize, len: usize) -> Option<Self> {
+        if let Some(len) = NonZeroUsize::new(len) {
+            ReadResultInner::slice(this.inner.as_ref()?, extra_offset, len)
+                .map(|inner| Self { inner: Some(inner) })
+        } else {
+            Some(Self::empty_buffer())
```
zserik, line 64: This is needed to make sure that calls to `slice` with `len = 0` are valid. If they aren't valid, then `len` should be changed to `NonZeroUsize`.

glommio/src/io/read_result.rs
bryandmc: Are we sure that the `NonZeroUsize` in the `len` field is actually gonna help optimize the layout? The example given in the docs refer to cases where it's wrapped in an option directly but not a case like this.. So do you have any documentation or even measured example of this working to optimize the layout and remove the need for the extra space for the tag?

zserik: I added an unit test for this, thanks.

glommer: Shouldn't those things go into comments in the code?

For github, once the PR is closed the information is lost

glommio/src/io/read_result.rs
```diff
@@ -137,3 +137,14 @@ impl ReadResultInner {
         }
     }
 }
+
+#[cfg(test)]
```
glommer, line 128: I'm a fan of testing expected struct sizes, although it is a bit sad in my opinion that rust doesn't seem to have the equivalent of `static_assert`, which is how I'd do this in c++

glommio/src/sys/dma_buffer.rs
```diff
@@ -177,13 +177,7 @@ impl DmaBuffer {
     /// });
     /// ```
     pub fn read_at(&self, offset: usize, dst: &mut [u8]) -> usize {
-        if offset > self.size {
```
glommer, line 180: for consistency, we could consider moving `write_at` to common code as well.
It's not a must since there is only one user currently.

glommio/src/io/read_result.rs
```diff
@@ -95,9 +95,6 @@ impl ReadResultInner {
         }
     }
 
-    /// Creates a slice of this ReadResult with the given offset and length.
```
glommer, line 54: they are unnecessary, but don't hurt either. Maybe we can leave them in place with `//`

zserik: which? The `slice*` methods?

glommio/src/sys/dma_buffer.rs
zserik: Huh, I think I've done that already?

glommer: I must have missed it. Will look again, thanks!

glommio/src/io/read_result.rs
glommer: the comments, I mean.
Doc comments may not be necessary, but comments are helpful and there is no harm in leaving them be.

Overall the series looks very good.

glommio/src/io/byte_slice_ext.rs
```diff
@@ -0,0 +1,120 @@
...
+        0
+    } else {
+        let len = std::cmp::min(source.len(), destination.len() - offset);
+        destination[offset..offset + len].copy_from_slice(&source[0..len]);
+        len
+    }
+}
+
+fn slice_memset(destination: &mut [u8], value: u8) {
+    unsafe { std::ptr::write_bytes(destination.as_mut_ptr(), value, destination.len()) }
```
zserik, line 97: This can be replaced with a safe `for` loop, which shouldn't cause any regressions (rustc is probably good enough to optimize that to `memset` or inline it), but reduce the amount of `unsafe` code.

glommer: should be easy to verify if this is indeed the case with `llvm-objdump`.

zserik: That may be the case, but I have no experience in using `llvm-objdump`, so I'm unable to verify this.

glommer: there's a way to cheat that which I find very handy. You can use rust playground, select release mode and instead of "build", you change the big red button to "show assembly"

https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=3d184f0f5c56d5bf1b7c722728f41623

The unsafe code seems slightly more efficient:

```
playground::slice_memset:
	movq	%rsi, %rax
	movl	%edx, %esi
	movq	%rax, %rdx
	jmpq	*memset@GOTPCREL(%rip)
```

The code using for does generate a memset - which is good news - but it also wraps it in some boundary tests:
```
playground::slice_memset_for:
	testq	%rsi, %rsi
	je	.LBB1_2
	pushq	%rax
	movq	%rsi, %rax
	movzbl	%dl, %esi
	movq	%rax, %rdx
	callq	*memset@GOTPCREL(%rip)
	addq	$8, %rsp

.LBB1_2:
	retq
```

My reading is that it is just special casing the zero-size array first, and then the rest is about the same with one extra stack push.

It's a price worth paying to remove unsafe. memset is unlikely to be a super hot path and if it is a user can roll out theirn own

zserik: ok, that also mean that currently, calling:
```rust
use glommio::io::ByteSliceMutExt;
[].memset(0);
```
introduces UB. that is bad.

zserik: it's probably enough to just let the compiler inline it when appropriate (which means marking with `#[inline]` to support inlining across crates, which is really useful for these low-level, public methods).

zserik:
I wasn't sure if `glommio/src/io/byte_slice_ext.rs` is the correct location for `ByteSlice*Ext`... Maybe it should be placed inside of `glommio/src/sys/` instead.

glommio/src/io/read_result.rs
zserik: I didn't want to remove the doc comments as then `clippy` would complain about missing doc comments...

glommer:
It looks great to me now. Do you feel this is ready to go ?
I'll probably squash the result if you don't mind

glommer:
> I wasn't sure if `glommio/src/io/byte_slice_ext.rs` is the correct location for `ByteSlice*Ext`... Maybe it should be placed inside of `glommio/src/sys/` instead.

sys is mostly for low level linux things.
byte slice being quite generic, maybe can go just at the top level `src/`

PR #250, opened 2020-12-25
+DmaOpenOptions

### LINKED ISSUE #249: DmaFile: provide a `create_new` method
The `create_new` method should work similarly to [create_new](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create_new).

Alternative: provide a `DmaFileBuilder` struct with a `create_new` method.

zserik:
### What does this PR do?

This PR adds a struct `DmaOpenOptions`, which allows to more freely choose `open_at` arguments via an already established builder pattern taken from libstd (e.g. it is modeled after [std::fs::OpenOptions](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html))

### Motivation

Previously, it was not possible afaik to open a file exclusively (`create_new`).

### Related issues

Fixes #249 (`create_new`)

### Additional Notes

This PR changes the default DmaFile opening mode from 0o644 to 0o666, to behave more similiar to the rust standard library.
Additionally, it changes the type of the `mode` argument of `read_at` in several places from `libc::c_int` to the more appropriate `libc::mode_t`.

I omitted `append` from the `DmaOpenOptions` methods. I don't think it makes much sense for unbuffered files.

Unit tests are still missing, which might be a bit challenging, at it is really hard to test every use case with such an kinda "open-end" builder pattern.

### Checklist

* [ ] I have added unit tests to the code I am submitting
* [ ] My unit tests cover both failure and success scenarios
* [X] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/src/io/dma_open_options.rs
+++ b/glommio/src/io/dma_open_options.rs
@@ -3,7 +3,7 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
-use crate::io::dma_file::DmaFile;
+use crate::io::dma_file::{DmaFile, Result};
 use std::io;
 use std::path::Path;
 
@@ -114,18 +114,18 @@ impl DmaOpenOptions {
         self
     }
 
-    pub(super) fn get_access_mode(&self) -> io::Result<libc::c_int> {
+    pub(super) fn get_access_mode(&self) -> Result<libc::c_int> {
         Ok(match (self.read, self.write) {
             (true, false) => libc::O_RDONLY,
             (false, true) => libc::O_WRONLY,
             (true, true) => libc::O_RDWR,
-            (false, false) => return Err(io::Error::from_raw_os_error(libc::EINVAL)),
+            (false, false) => return Err(io::Error::from_raw_os_error(libc::EINVAL).into()),
         })
     }
 
-    pub(super) fn get_creation_mode(&self) -> io::Result<libc::c_int> {
+    pub(super) fn get_creation_mode(&self) -> Result<libc::c_int> {
         if !self.write && (self.truncate || self.create || self.create_new) {
-            Err(io::Error::from_raw_os_error(libc::EINVAL))
+            Err(io::Error::from_raw_os_error(libc::EINVAL).into())
         } else {
             Ok(match (self.create, self.truncate, self.create_new) {
                 (false, false, false) => 0,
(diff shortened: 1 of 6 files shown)


glommer:
I agree about this pattern being the right one.

glommio/src/io/dma_open_options.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // the following methods are directly taken from std/sys/unix/fs.rs
+    // NOTE(zserik): I omitted `append` from the following methods.
+    //               I don't think it makes much sense for unbuffered files.
+
+    /// Sets the option for read access.
+    ///
+    /// This option, when true, will indicate that the file should be read-able if opened.
+    pub fn read(&mut self, read: bool) -> &mut Self {
```
glommer, line 58: shouldn't those be `mut self` returning `Self` instead ? (consuming self)

glommio/src/io/dma_open_options.rs
```diff
@@ -0,0 +1,192 @@
...
+            // system-specific
+            custom_flags: 0,
+            // previously, we defaulted to 0o644, but 0o666 is usedd by libstd
+            mode: 0o666,
+        }
+    }
+
+    // the following methods are directly taken from std/sys/unix/fs.rs
+    // NOTE(zserik): I omitted `append` from the following methods.
+    //               I don't think it makes much sense for unbuffered files.
```
glommer, line 53: I agree about the lack of append, but that's mostly because it rarely makes sense to write to the middle of a file - (even a buffered one, if you ask me).

Then append and write kind of becomes the same.

glommer:
Good direction, let me know when this is ready to go.

glommio/src/io/dma_open_options.rs
zserik: No, the pattern used seems correct, and is directly copied from `libstd`, so it's correct, because the actual `open` fn takes a `&self`, which makes this work. It has the benefit that it allows both classic chaining and simple method calling, which is good for usability.

glommio/src/io/dma_open_options.rs
zserik: `append` has an important difference to `write`: it is not possible to change the write position inside of a file to anything except the end of the file, and writes always append to the end of the file, plus writes are atomic.

zserik:
The only thing that is missing is unit tests for this... The code should work, tho, as it is mostly directly copied from `libstd`, and the common use cases are implicitly tested, as `DmaOpenOptions` is used in `DmaFile::open`/`DmaFile::create`.

glommer:
there is also some commented out code that would be good to see gone in the final version (either by deleting the comments or making them official)

glommio/src/io/dma_open_options.rs
glommer: All reads and writes to our `DmaFile` take a position anyway (true for both buffered and dma)

zserik:
I think I've mostly fixed that in 0050d0d.

glommer:
merging this - sorry it took me a while to get to it

PR #252, opened 2020-12-26
Add an example of using TcpStream in parallel

thirstycrow:
### What does this PR do?

Add an example demonstrating how to use glommio TcpStream in multiple tasks in parallel.

### Motivation

I'm working on a server that talks redis' pipeline protocol but will do some IO when processing the requests. So the requests would have to be processed in parallel for performance.

### Additional Notes

I spent a day just to make it work, but I'm not sure it is what it could be best. The `AsyncIO` is something like future_lite's ReadHalf/WriteHalf, with the inner `Mutex` removed, because in glommio we are always in the same thread, so `Mutex` is not necessary. This make `AsyncIO` trivial to write, and may be better be part of glommio. (Or make glommio TcpStream ready be used in parallel by default?) Another difference is that there can be as many `AsyncIO` for a TcpStream, while with future_lite we can just get a single `ReadHalf` and `WriteHalf`.

### Checklist

### DIFF
--- a/examples/pipeline.rs
+++ b/examples/pipeline.rs
@@ -1,0 +1,180 @@
+use std::cell::RefCell;
+use std::io::Result;
+use std::net::SocketAddr;
+use std::ops::{Deref, DerefMut};
+use std::pin::Pin;
+use std::rc::Rc;
+use std::time::Duration;
+
+use futures::io::IoSliceMut;
+use futures::task::{Context, Poll};
+use futures::{AsyncBufReadExt, AsyncWriteExt};
+use futures_lite::{AsyncBufRead, AsyncRead, AsyncWrite};
+
+use glommio::channels::local_channel;
+use glommio::channels::local_channel::{LocalReceiver, LocalSender};
+use glommio::net::{TcpListener, TcpStream};
+use glommio::task::JoinHandle;
+use glommio::timer::sleep;
+use glommio::{enclose, Local, LocalExecutor, Task};
+
+type Request = Rc<Vec<u8>>;
+type Response = Rc<Vec<u8>>;
+
+pub fn server(port: u16) -> SocketAddr {
+    let listener = Rc::new(TcpListener::bind(("0.0.0.0", port)).unwrap());
+
+    Local::local(enclose!((listener) async move {
+        while let Ok(stream) = listener.accept().await {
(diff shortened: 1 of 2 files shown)


glommer:
a general comment before I get started on the review:

Anybody can write a thread per core system with any language and any framework. The question becomes: how easy it is, and how efficient the result turns out to be.

We got the efficiency part at least theoretically handled (we'll find bugs here and there that hinder it, but such is life).

I am very interested in making it very easy, very ergonomic, and very consumable. All of that to say that as long as we can make interfaces generic enough, I am all for making lock-free parallelism  inherent capabilities of our data structures, and/or trivially achievable with traits on top.

examples/pipeline.rs
```diff
@@ -0,0 +1,180 @@
...
+                pipeline(stream).await
+            })).detach();
+        }
+    }))
+    .detach();
+
+    listener.local_addr().unwrap()
+}
+
+async fn pipeline(stream: Rc<RefCell<TcpStream>>) {
```
glommer, line 41: Is the idea that messages are received sequentially at the TCP level but then processed in parallel and potentially out-of-order? This is a great example of where a semaphore can be useful to limit parallelism.

I think we can merge such structure in a `sink`-`source` paradigm to glommio.

In general terms, we can require a `sink` to implement an adaptor to connect to the `source`, which will be native to glommio. The example then can be massively simplified.

I think a nice example could even be writing the stream to a file, using one of our file data structures - but up to you.

examples/pipeline.rs
```diff
@@ -0,0 +1,180 @@
+use std::cell::RefCell;
```
glommer, line 1: I will leave it up to you whether you want to merge this example as is or spend some effort in merging more core infrastructure to glommio first.

I am fine with whatever decision you make!

However if this is to merged as an example as is, either temporarily or definitely, it would be nice to have more comments explaining to the reader what is the example doing, and how they can make use of a similar thing.

Examples are, after all, a way for people to learn.

examples/pipeline.rs
thirstycrow: > I think we can merge such structure in a `sink`-`source` paradigm to glommio.

It's a good idea to have something like that in glommio! It reminds me `akka-stream`, which is funny to work with. I'll spend some time trying to make a prototype of this.

Issue #243, opened 2020-12-18
Advanced options on networking are not exposed, but should be


glommer:
If we look at the API defined by `net::TcpStream`, `net::UdpSocket`, and others there are many options we are not implementing (setting timeouts, multicast, etc).

They shouldn't be hard to implement -> essentially just passing options down sockets, but I will require that they come with unit testing unless it is really not possible to.

glommer:
Timeouts are now implemented for sockets. The issue is still open because it is mostly an umbrella issue. There's still work to do, especially around Udp.

Suyash458:
@glommer I'd like to add options for `UdpSocket` if this is not something being worked on currently

glommer:
Hello @Suyash458 ! To the best of my knowledge there's no one working on this at the moment.

PR #257, opened 2021-01-11
TcpStream nodelay methods

mxxo:
### What does this PR do?

Adds the `TcpStream::nodelay` getter and changes `TcpStream::set_nodelay`'s signature to match [`std::net::TcpStream`](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_nodelay).

### Motivation

Continue work on #243 

### Related issues

#243 

### Additional Notes

For 735251c, I included the standard library rationale for `set_nodelay` (and most other socket setters) taking an immutable reference, i.e. 
> RFC 1145: "This aspect mirrors the ability of the underlying system to deal with concurrent modification of socket options."

This PR changes the `set_nodelay` signature from `&mut` to `&`. The opposite change, moving from `&` to `&mut`, breaks semver since old code may fail to compile. On the other hand I am relatively sure `&mut` to `&` doesn't count as API breakage, since passing `&mut` to a method expecting `&` will still compile, although there might be new warnings about needless mutability: [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e4096a90e9fcf367875dcc680a580547).
 
### Checklist

### DIFF
--- a/glommio/src/net/tcp_socket.rs
+++ b/glommio/src/net/tcp_socket.rs
@@ -419,6 +419,27 @@ impl TcpStream {
         self.stream.stream.set_nodelay(value).map_err(Into::into)
     }
 
+    /// Gets the `TCP_NODELAY` option on this socket.
+    ///
+    /// For more information about this option, see [`TcpStream::set_nodelay`].
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use glommio::net::TcpStream;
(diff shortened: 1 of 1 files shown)


glommer:
I am not overly concerned about semver as we are way below 1.0 at the moment - but we're getting there!!

glommer (approved):

Issue #229, opened 2020-12-15
Asan not returning cleanly


glommer:
@laa hit this recently, but our `cargo test` with asan does not return cleanly.

Command line:
```
cargo +nightly test -Zbuild-std --target x86_64-unknown-linux-gnu
```

with flags:
```
export RUSTFLAGS=-Zsanitizer=address
```

They seem all related to `Rc` / `Arc` lifetime but some will be tricky to fix
In particular, the way those dependencies are usually broken is by moving things to `Weak`. But we don't want to do this with the `Arc` we use for eventfd: passing a `Weak` pointer would make us have to upgrade it before use, and that's full of atomics which would tank the shared channel performance.

The others I found are related to the submission queue -> I tried to move the `EnqueuedSource` to Weak and it didn't fix it, so needs more investigation, and the memory allocator for registered buffers.

glommer:
Ok, I think the biggest issue is that `file` now has a reactor copy inside. That should have been a `Weak` copy.
We'll face the same issue with `timers` and `tcp_streams` but it should be mostly a mechanical fix

glommer:
https://github.com/DataDog/glommio/pull/231 handles most of those cases. Some tests still report leaks, and I don't know at this moment why. But the PR should make the situation better.

laa:
Asan reported on the latest commit:

```
Indirect leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x5635224d051d in malloc /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:145:3
    #1 0x56352290abb5 in std::sys::unix::alloc::_$LT$impl$u20$core..alloc..global..GlobalAlloc$u20$for$u20$std..alloc..System$GT$::alloc::h4094fbb2da8d8ff0 /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/unix/alloc.rs:14:13
    #2 0x5635228fe6a1 in __rdl_alloc /home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/alloc.rs:356:13
    #3 0x56352250b64e in alloc::alloc::alloc::h34b4eb3c37d61eba (/tmp/rustdoctestWMhJMD/rust_out+0x12464e)
[...]

PR #258, opened 2021-01-14
Issue #229 has been fixed. 

laa:
### What does this PR do?

An issue that caused program abort, in the presence of not finished timer in the executor,  has been fixed.

Issue #229 has been caused by the fact that the timer kept instance of Waker, which was not used to wake the future. That drove cyclic reference between the not finished timer and future itself through waker instance.

Program abort was caused by the drop of the future that triggered drop of the waker.  As a result, waker tries to reschedule the local executor task, which causes panic because the local executor is absent. The panic inside of local executor triggers drop of the timer, but the reactor is already absent at this moment which triggers panic inside panic and as a result triggers abort of the program.

### Motivation

Bug fixing.

### Related issues

Issue #229

### Checklist

### DIFF
--- a/glommio/src/multitask.rs
+++ b/glommio/src/multitask.rs
@@ -129,7 +129,7 @@ impl LocalExecutor {
     pub(crate) fn spawn<T>(
         &self,
         tq: Rc<RefCell<TaskQueue>>,
-        future: impl Future<Output=T>,
+        future: impl Future<Output = T>,
     ) -> Task<T> {
         let tq = Rc::downgrade(&tq);
 
--- a/glommio/src/multitask.rs
+++ b/glommio/src/multitask.rs
@@ -113,6 +113,7 @@ pub(crate) struct LocalExecutor {
 }
 
 impl UnwindSafe for LocalExecutor {}
+
 impl RefUnwindSafe for LocalExecutor {}
 
 impl LocalExecutor {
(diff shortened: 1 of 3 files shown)


glommio/src/multitask.rs
```diff
@@ -128,18 +129,21 @@ impl LocalExecutor {
...
-            {
-                let queue = tq.borrow();
-                queue.ex.local_queue.push(runnable);
+            let tq = tq.upgrade();
+
+            if let Some(tq) = tq {
+                {
+                    let queue = tq.borrow();
+                    queue.ex.local_queue.push(runnable);
+                }
```
glommer, line 144: Fantastic work!

I am wondering if this one is really necessary, though ?
If I understood this correctly, we no longer remove the timer if the reactor is not present from the timer side. So why do we have to test for queues being still in existence here ?

laa: Otherwise, we would panic, on the test above because the task queue is absent. We schedule task on waker drop, but we can not because the queue is already dropped.

laa: But in the case of the "background" task, the only thing which is left during the drop of the executor is waker. If such a task is paused of course.

glommer: ok, that makes sense.

PR #260, opened 2021-01-16
Add a benchmark for submitting nop events to the ring

gardnervickers:
- Add a new feature for benchmark-related code, "bench".
- Add a new NopSubmitter type for submitting nop operations to the ring,
  gated behind the "bench" feature.
- Changes to Source and Source handling to support nop submissions.
- Add a new benchmark which measures nop submission rate for a variety of
  task counts.

### What does this PR do?

This PR adds a new `NopSubmitter` reactor handle which can submit nops to the ring. This is to support a new nop benchmark.

### Motivation

This benchmark was created to compare the current performance of submission/completion in Glommio against another set of proposed changes.

### Checklist

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -41,6 +41,9 @@ futures = "0.3.5"
 fastrand = "1.4.0"
 tokio = { version = "0.3.5", default-features = false, features = ["rt", "macros", "rt-multi-thread", "net", "io-util", "time"] }
 
+[features]
+bench = []
+
 [[bench]]
 name = "executor"
 harness = false
@@ -73,3 +76,6 @@ harness = false
 name = "udp"
 harness = false
 
+[[bench]]
+name = "nop"
+harness = false
--- a/glommio/benches/nop.rs
+++ b/glommio/benches/nop.rs
@@ -1,13 +1,11 @@
-//! Benchmark the noop performance of Glommio. This is a stress test for the executor/reactor
-//! infrastructure.
-
+//! Benchmark the performance of the submission ring by timing tasks which submit and wait on noop requests.
 use glommio::{LocalExecutorBuilder, Task};
 use std::fmt;
 use std::time::{Duration, Instant};
 
 struct Bench {
-    num_tasks: u64,
-    num_events: u64,
+    num_tasks: u32,
+    num_events: u32,
 }
 
 const BENCH_RUNS: &'static [Bench] = &[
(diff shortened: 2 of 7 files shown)


glommio/src/nop.rs
```diff
@@ -0,0 +1,31 @@
+//! Provides a [`NopSubmitter`] which can be used to submit nop operations to the ring. This is mainly useful for benchmarking Glommio.
```
gardnervickers, line 1: Do we need a license header for these?

glommer: We don't have license headers is any of our files. There is the datadog snippet but that's not a license. I don't expect non-DD people to add that - although you can, for consistency, if you want.

glommio/benches/nop.rs
```diff
@@ -0,0 +1,99 @@
+//! Benchmark the noop performance of Glommio. This is a stress test for the executor/reactor
```
glommer, line 1: this isn't measuring the performance "of glommio", but rather just a part of it (the submission to the ring)
Just a nitpick

glommio/benches/nop.rs
```diff
@@ -0,0 +1,99 @@
...
+    },
+];
+
+struct Measurement {
+    total_events: u64,
+    total_tasks: u64,
+    duration: Duration,
+}
+
+impl fmt::Display for Measurement {
```
glommer, line 42: not a request for change, but I am wondering if this code couldn't be made generic for the other benchmarks.
The creation of the executor is potentially generic as well. We'd have a facility to create a measuring executor (where we pin to CPU0, and count the loops + time) and a helper executor (where we pin to some other CPU, and don't count).

See the TCP benchmarks or shared channel to see how those would be handy.

glommio/Cargo.toml
```diff
@@ -40,6 +40,9 @@ futures = "0.3.5"
 fastrand = "1.4.0"
 tokio = { version = "0.3.5", default-features = false, features = ["rt", "macros", "rt-multi-thread", "net", "io-util", "time"] }
 
+[features]
```
glommer, line 43: I am surprised we need a feature for this.

glommio/benches/nop.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            r"completed submission of {} noop events across {} tasks, {} per task
```
glommer, line 46: @gardnervickers A small detail here: I prefer reporting the average cost per individual task -- see the other benchmarks for that.

gardnervickers: That makes more sense, thanks!

glommer:
I have no idea why, but CI hasn't run here.
Does anybody know a way to kick it ?

glommer (approved):

glommer:
@gardnervickers maybe rebase this on top of current master to see if CI runs ?

gardnervickers:
@glommer Has the `Build forked pull requests` option been enabled in CircleCI? 
https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories

glommer:
every contributor including me always send PRs from forked repositories.
I do not allow pushing directly to the main branch.

glommer:
Maybe you wronged someone at Gitlab and now they hate you.

I will test this manually and merge today.

glommer:
I merged because what's the worst that can happen... However the --build-all step fails:


```
cargo build --all --all-targets
   Compiling glommio v0.3.0 (/home/glauber/glommio/glommio)
   Compiling examples v0.0.0 (/home/glauber/glommio/examples)
error[E0433]: failed to resolve: could not find `nop` in `glommio`
  --> glommio/benches/nop.rs:85:38
   |
85 |             let submitter = glommio::nop::NopSubmitter::new();
   |                                      ^^^ could not find `nop` in `glommio`

error: aborting due to previous error
```

PR #263, opened 2021-01-22
Fix 261

### LINKED ISSUE #261: Shared channel benchmark fails/hangs with probability

glommer:
### What does this PR do?

Fixes #261.

### Motivation

Honestly, stubbornness. I got this close to quitting and going to live in a farm, planting potatoes, never getting any close to a computer ever again.

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -35,6 +35,7 @@ buddy-alloc = "0.4.1"
 ahash = "0.5.7"
 intrusive-collections = "0.9.0"
 thiserror = "1.0"
+membarrier = "0.2.2"
 
 [dev-dependencies]
 futures = "0.3.5"
--- a/glommio/benches/shared_channel.rs
+++ b/glommio/benches/shared_channel.rs
@@ -11,7 +11,7 @@ fn test_spsc(capacity: usize) {
         .pin_to_cpu(0)
         .spin_before_park(Duration::from_millis(10))
         .spawn(move || async move {
-            let sender = sender.connect();
+            let sender = sender.connect().await;
             let t = Instant::now();
             for _ in 0..runs {
                 sender.send(1).await.unwrap();
@@ -29,7 +29,7 @@ fn test_spsc(capacity: usize) {
         .spin_before_park(Duration::from_millis(10))
         .pin_to_cpu(1)
         .spawn(move || async move {
-            let receiver = receiver.connect();
+            let receiver = receiver.connect().await;
             let t = Instant::now();
             for _ in 0..runs {
                 receiver.recv().await.unwrap();
--- a/glommio/benches/udp.rs
+++ b/glommio/benches/udp.rs
@@ -12,8 +12,8 @@ fn main() {
     let server_ex = LocalExecutorBuilder::new()
         .pin_to_cpu(0)
         .spawn(move || async move {
-            let addr_sender = server_sender.connect();
-            let addr_receiver = client_receiver.connect();
+            let addr_sender = server_sender.connect().await;
+            let addr_receiver = client_receiver.connect().await;
 
             let receiver = UdpSocket::bind("127.0.0.1:0").unwrap();
             addr_sender
@@ -46,8 +46,8 @@ fn main() {
     let client_ex = LocalExecutorBuilder::new()
         .pin_to_cpu(1)
         .spawn(move || async move {
-            let addr_receiver = server_receiver.connect();
-            let addr_sender = client_sender.connect();
+            let addr_receiver = server_receiver.connect().await;
+            let addr_sender = client_sender.connect().await;
 
             let addr = addr_receiver.recv().await.unwrap();
             let client = UdpSocket::bind("127.0.0.1:0").unwrap();
--- a/glommio/src/channels/mod.rs
+++ b/glommio/src/channels/mod.rs
@@ -78,6 +78,12 @@ pub mod local_channel;
 /// respond back, other executors that are not really connected to anybody, cliques of
 /// executors, you name it.
 ///
+/// However, note that connecting a channel will involve a blocking operation as we need to
+/// synchronize and exchange information between peers. That can block the executor, although in
+/// practice that will only happen if it races with the creation of a new executor. Because of
+/// that, The best performance pattern when using shared channels is to create both the
+/// executors and channels early, connect the channels as soon as they start, and keep them alive.
+///
 /// Data that goes into the channels need to be [`Send`], as well as the channels themselves
 /// (otherwise you would not be able to really pass them along to the actual executors). But to
 /// prevent accidental use from multiple producers, the channel endpoints have to be `connected` before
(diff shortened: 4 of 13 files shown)


gardnervickers:
@glommer Has the `Build forked pull requests` option been enabled in CircleCI?

thirstycrow:
Got "too many open files" with the following code:

```rust
fn main() {
    for i in 0..10000 {
        println!("==========");
        println!("Round {}", i);
        test_spsc(1);
    }
}
```

```
...
==========
Round 506
==========
Round 507
thread 'unnamed-1015' panicked at 'cannot initialize I/O event notification: Os { code: 24, kind: Other, message: "Too many open files" }', glommio/src/parking.rs:259:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'unnamed-1016' panicked at 'cannot initialize I/O event notification: Os { code: 24, kind: Other, message: "Too many open files" }', glommio/src/parking.rs:259:14
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any', glommio/benches/shared_channel.rs:32:19
error: bench failed
```

thirstycrow:
A latest run of the above code hangs at round 492.

thirstycrow:
https://github.com/glommer/glommio/blob/8d768f1313a292047332c75cdcdb0dd83bebde22/glommio/src/channels/spsc_queue.rs#L143

```
            self.shadow_head.set(self.head.load(Ordering::Relaxed));
                                                ~~~~~~~~~~~~~~~~~
                                                Should this be Ordering::Acquire?
```

glommer:
Ok, it is likely that the eventfd files are not being released. It is strange, because we do have a drop implementation that should release them.

About the memory ordering, that code uses cached versions of head and tail to avoid always issuing potentially expensive operations. However I do suspect that this is not doing us any favors and am considering reverting back to a simpler version (this is originally 3rd party code).

thirstycrow:
I tried changing the load ordering from relaxed to acquire, but that does not fix it. 

The hanging problem is quite easy to reproduce on my laptop. With `test_spsc(1)` and `let runs: u32 = 100_000`, the bench never goes beyond 80 rounds.

I found that by removing the `self.notifier.wake_up()`, the bench will always reach round 507. Another solution seems to be adding some statement such as a `println`, between `self.notifier.prepare_to_sleep()` and `if process_remote_channels(wakers) == 0`

```rust
            self.notifier.prepare_to_sleep();
            // println!("hello");  <=== Add a simple print here seems also work
            if process_remote_channels(wakers) == 0 {
                self.link_rings_and_sleep(&mut main_ring, &self.eventfd_src)?;
                self.notifier.wake_up();  <===  // The HANGING problem seems disappeared
                                                // with this line commented out
                consume_rings!(into wakers; lat_ring, poll_ring, main_ring);
            }
```

For the above code to work properly, it is required that a thread to set its notifier memory before processing remote channels and going to sleep. Is it possible that the 2 steps be reversed as result of instruction reordering?

glommer:
Right. This is very clear an issue with the wake ups still, so the memory ordering of the queue should not be affecting us much.

Indeed my suspicions lie in the place you highlighted, so what you claim is possible. However `prepare_to_sleep` stores in the sleep notifier's `memory` with `Release` ordering , and `must_notify()` loads the same variable with `Acquire` ordering. It is my understanding that this should make sure that all writes up until this point are visible.

The idea of that snippet is that first, we must announce and commit that we are going to sleep.
After that, if the other side of the channel is activated `must_notify` will return true so a notification will be sent.

glommer:
maybe what's going on is that the size of the channel is not yet seen as updated. Can you try changing the memory order in `prepare_to_sleep` to `SeqCst` ? I don't think stronger serialization in both the reader and writer should be needed.

glommer:
I just pushed a commit that should fix the eventfd issue so we can get it out of the way.
I can reproduce the issue as well, just not as often as you, apparently. SeqCst definitely helps and using a simplified version of the spsc queue without those cached variables, I am yet to see a hang.

With the cached versions I still see hangs. I will work on it a bit more to make sure this is not because of changed timing.

glommer:
The new version should fix all issues.

glommer:
This has been running for a while for me, plus it actually makes sense that it would hang with the lack of the `membarrier` call. So I am merging this. Happy to reopen if the problems continue.

Thanks a lot for throughly testing this!

PR #264, opened 2021-01-27
add a version of the semaphore permit for complex lifetimes

glommer:
Sometimes the lifetimes associated with the permits are complex. We may
move them, pass around, and keep them alive to signal that a long-lived
operation is still in progress.

In such sitations we need a more complex permit, that is able to keep
the semaphore alive for the lifetime of the operation.

In reality I expect most permits to be like this. But recognizing that
sometimes simple is good, I am adding a separate permit acquisition
for these cases instead of requiring that permits are always used with
an Rc.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/sync/semaphore.rs
+++ b/glommio/src/sync/semaphore.rs
@@ -15,6 +15,7 @@ use intrusive_collections::{container_of, offset_of, Adapter, PointerOps};
 use intrusive_collections::{LinkedList, LinkedListLink};
 use std::marker::PhantomPinned;
 use std::ptr::NonNull;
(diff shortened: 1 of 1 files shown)


glommer:
Quick heads up that I tried this in my project and it says that this borrow is already taken:
```
            Self::remove_from_waiting_queue(waiter_node, &mut self.semaphore.state.borrow_mut())
 ```

I haven't investigated further

laa (approved):

PR #265, opened 2021-01-27
Require the "bench" feature for the nop benchmark

gardnervickers:
This add a `required-feature` to the nop benchmark, requiring the `"bench"` feature. 

This should fix `cargo build --all --all-targets` which was broken as part of #260


glommer:
and CI even likes you again!

PR #268, opened 2021-02-03
Build fully connected mesh of channels 

thirstycrow:
### What does this PR do?

This PR adds a utility to build a mesh of channels connecting each pair of participating executors.

### Motivation

Make it easier to implement sharding

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -81,3 +81,7 @@ harness = false
 name = "nop"
 harness = false
 required-features = ["bench"]
+
+[[bench]]
+name = "sharding"
+harness = false
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -17,6 +17,7 @@ readme = "../README.md"
 log = "0.4"
 concurrent-queue = "1.1.2"
 futures-lite = "1.11.1"
+futures = "0.3.12"
 libc = "0.2.73"
 socket2 = { version = "0.3.18", features = ["unix", "reuseport"] }
 iou = { git = "https://github.com/glommer/iou", tag = "glommio-2020-11-30" }
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -17,7 +17,6 @@ readme = "../README.md"
 log = "0.4"
 concurrent-queue = "1.1.2"
 futures-lite = "1.11.1"
-futures = "0.3.12"
 libc = "0.2.73"
 socket2 = { version = "0.3.18", features = ["unix", "reuseport"] }
 iou = { git = "https://github.com/glommer/iou", tag = "glommio-2020-11-30" }
--- a/glommio/src/channels/channel_mesh.rs
+++ b/glommio/src/channels/channel_mesh.rs
@@ -77,6 +77,7 @@ impl<T: Send + Copy> Senders<T> {
     }
 }
 
+/// Receiver side
 #[derive(Debug)]
 pub struct Receivers<T: Send + Copy> {
     peer_id: usize,
@@ -120,6 +121,10 @@ impl<T: Send + Copy> Receivers<T> {
         }
     }
 
+    /// Returns a vec of [`ConnectedReceiver`]s with the id of their upstream
+    /// producers.
+    ///
+    /// [`ConnectedReceiver`]: ../shared_channel/struct.ConnectedReceiver.html
     pub fn streams(&mut self) -> Vec<(usize, ConnectedReceiver<T>)> {
         self.receivers
             .iter_mut()
@@ -184,6 +189,7 @@ pub trait MeshAdapter: Clone {
     fn is_full(&self) -> bool;
 }
 
+/// Adapter for full mesh
 #[derive(Clone, Debug)]
 pub struct Full;
 
@@ -197,8 +203,10 @@ impl MeshAdapter for Full {
     }
 }
 
+/// Alias for full mesh builder
 pub type FullMesh<T> = MeshBuilder<T, Full>;
 
+/// Adapter for partial mesh
 #[derive(Clone, Debug)]
 pub struct Partial;
 
@@ -212,6 +220,7 @@ impl MeshAdapter for Partial {
     }
 }
 
+/// Alias for partial mesh builder
 pub type PartialMesh<T> = MeshBuilder<T, Full>;
 
 /// A builder for channel mesh
@@ -281,6 +290,7 @@ impl<T: 'static + Send + Copy, A: MeshAdapter> MeshBuilder<T, A> {
         }
     }
 
+    /// Returns number of all peers in the mesh
     pub fn nr_peers(&self) -> usize {
         self.nr_peers
     }
--- a/glommio/src/channels/channel_mesh.rs
+++ b/glommio/src/channels/channel_mesh.rs
@@ -12,8 +12,9 @@ use std::sync::{Arc, RwLock};
 use crate::channels::shared_channel::{self, *};
 use crate::{GlommioError, Local, Result, Task};
 
+/// Sender side
 #[derive(Debug)]
-struct Senders<T: Send + Copy> {
+pub struct Senders<T: Send + Copy> {
     peer_id: usize,
     producer_id: Option<usize>,
     senders: Vec<Option<ConnectedSender<T>>>,
@@ -36,21 +37,39 @@ impl<T: Send + Copy> Senders<T> {
     }
 
     /// Send a message to the idx-th consumer
+    ///
+    /// It returns a [`GlommioError::IoError`] encapsulating a [`InvalidInput`] if the idx is out of
+    /// the range of available senders, or the sender is a placeholder in the case of full mesh.
+    ///
+    /// See [`ConnectedSender.send`] for how the underlying sender works.
+    ///
+    /// [`GlommioError::IoError`]: ../../error/enum.GlommioError.html#variant.IoError
+    /// [`InvalidInput`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.InvalidInput
+    /// [`ConnectedSender.send`]: ../shared_channel/struct.ConnectedSender.html#method.send
     pub async fn send_to(&self, idx: usize, msg: T) -> Result<(), T> {
-        match self.senders[idx].as_ref() {
-            Some(consumer) => consumer.send(msg).await,
-            None => Err(GlommioError::IoError(Error::new(
+        match self.senders.get(idx) {
+            Some(Some(consumer)) => consumer.send(msg).await,
+            _ => Err(GlommioError::IoError(Error::new(
                 ErrorKind::InvalidInput,
                 "Local message should not be sent via channel mesh",
             ))),
         }
     }
 
-    /// Try sending a message to the idx-th consumer
+    /// Send a message to the idx-th consumer
+    ///
+    /// It returns a [`GlommioError::IoError`] encapsulating a [`InvalidInput`] if the idx is out of
+    /// the range of available senders, or the sender is a placeholder in the case of full mesh.
+    ///
+    /// See [`ConnectedSender.try_send`] for how the underlying sender works.
+    ///
+    /// [`GlommioError::IoError`]: ../../error/enum.GlommioError.html#variant.IoError
+    /// [`InvalidInput`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.InvalidInput
+    /// [`ConnectedSender.try_send`]: ../shared_channel/struct.ConnectedSender.html#method.try_send
     pub fn try_send_to(&self, idx: usize, msg: T) -> Result<(), T> {
-        match self.senders[idx].as_ref() {
-            Some(consumer) => consumer.try_send(msg),
-            None => Err(GlommioError::IoError(Error::new(
+        match self.senders.get(idx) {
+            Some(Some(consumer)) => consumer.try_send(msg),
+            _ => Err(GlommioError::IoError(Error::new(
                 ErrorKind::InvalidInput,
                 "Local message should not be sent via channel mesh",
             ))),
@@ -59,7 +78,7 @@ impl<T: Send + Copy> Senders<T> {
 }
 
 #[derive(Debug)]
-struct Receivers<T: Send + Copy> {
+pub struct Receivers<T: Send + Copy> {
     peer_id: usize,
     consumer_id: Option<usize>,
     receivers: Vec<Option<ConnectedReceiver<T>>>,
@@ -82,14 +101,31 @@ impl<T: Send + Copy> Receivers<T> {
     }
 
     /// Receive a message from the idx-th producer
+    ///
+    /// It returns a [`GlommioError::IoError`] encapsulating a [`InvalidInput`] if the idx is out of
+    /// the range of available receivers, or the receiver is a placeholder in the case of full mesh.
+    ///
+    /// See [`ConnectedReceiver.recv`] for how the underlying sender works.
+    ///
+    /// [`GlommioError::IoError`]: ../../error/enum.GlommioError.html#variant.IoError
+    /// [`InvalidInput`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.InvalidInput
+    /// [`ConnectedReceiver.recv`]: ../shared_channel/struct.ConnectedReceiver.html#method.recv
     pub async fn recv_from(&self, idx: usize) -> Result<Option<T>, ()> {
-        match self.receivers[idx].as_ref() {
-            Some(producer) => Ok(producer.recv().await),
-            None => Err(GlommioError::IoError(Error::new(
+        match self.receivers.get(idx) {
+            Some(Some(producer)) => Ok(producer.recv().await),
+            _ => Err(GlommioError::IoError(Error::new(
                 ErrorKind::InvalidInput,
                 "Local message should not be received from channel mesh",
             ))),
         }
+    }
+
+    pub fn streams(&mut self) -> Vec<(usize, ConnectedReceiver<T>)> {
+        self.receivers
+            .iter_mut()
+            .enumerate()
+            .flat_map(|(idx, recv)| recv.take().map(|recv| (idx, recv)))
+            .collect()
     }
 }
 
@@ -116,10 +152,14 @@ type SharedChannel<T> = (
 
 type SharedChannels<T> = Vec<Vec<SharedChannel<T>>>;
 
+/// The role an executor plays in the mesh
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum Role {
+    /// The executor produces message
     Producer,
+    /// The executor consumes message
     Consumer,
+    /// The executor produces and consumes message
     Both,
 }
 
@@ -133,13 +173,19 @@ impl Role {
     }
 }
 
-pub trait MeshAdapter {
+/// An adapter for MeshBuilder
+pub trait MeshAdapter: Clone {
+    /// Determine whether a channel should be created between a pair of peers
+    /// considering their roles.
     fn connect(&self, from: &Role, to: &Role) -> bool;
+
+    /// Determine whether we are building a full mesh, when placeholders should
+    /// be inserted into the list senders/receivers
     fn is_full(&self) -> bool;
 }
 
-#[derive(Clone)]
-struct Full;
+#[derive(Clone, Debug)]
+pub struct Full;
 
 impl MeshAdapter for Full {
     fn connect(&self, _: &Role, _: &Role) -> bool {
@@ -151,8 +197,10 @@ impl MeshAdapter for Full {
     }
 }
 
-#[derive(Clone)]
-struct Partial;
+pub type FullMesh<T> = MeshBuilder<T, Full>;
+
+#[derive(Clone, Debug)]
+pub struct Partial;
 
 impl MeshAdapter for Partial {
     fn connect(&self, from: &Role, to: &Role) -> bool {
@@ -164,7 +212,9 @@ impl MeshAdapter for Partial {
     }
 }
 
-#[derive(Clone)]
+pub type PartialMesh<T> = MeshBuilder<T, Full>;
+
+/// A builder for channel mesh
 pub struct MeshBuilder<T: Send + Copy, A: MeshAdapter> {
     nr_peers: usize,
     channel_size: usize,
@@ -174,6 +224,18 @@ pub struct MeshBuilder<T: Send + Copy, A: MeshAdapter> {
 }
 
 unsafe impl<T: Send + Copy, A: MeshAdapter> Send for MeshBuilder<T, A> {}
+
+impl<T: Send + Copy, A: MeshAdapter> Clone for MeshBuilder<T, A> {
+    fn clone(&self) -> Self {
+        Self {
+            nr_peers: self.nr_peers,
+            channel_size: self.channel_size,
+            peers: self.peers.clone(),
+            channels: self.channels.clone(),
+            adapter: self.adapter.clone(),
+        }
+    }
+}
 
 impl<T: Send + Copy, A: MeshAdapter> Debug for MeshBuilder<T, A> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
@@ -217,6 +279,10 @@ impl<T: 'static + Send + Copy, A: MeshAdapter> MeshBuilder<T, A> {
             channels: Arc::new(Self::placeholder(nr_peers)),
             adapter,
         }
+    }
+
+    pub fn nr_peers(&self) -> usize {
+        self.nr_peers
     }
 
     fn placeholder(nr_peers: usize) -> SharedChannels<T> {
@@ -400,6 +466,7 @@ mod tests {
 
     #[test]
     #[should_panic]
+    #[allow(unused_must_use)]
     fn test_join_more_than_once() {
         let mesh_builder = MeshBuilder::<u32, _>::full(2, 1);
 
@@ -410,6 +477,7 @@ mod tests {
 
     #[test]
     #[should_panic]
+    #[allow(unused_must_use)]
     fn test_join_more_than_capacity() {
         do_test_channel_mesh(1, 1, 2);
     }
--- a/glommio/src/channels/channel_mesh.rs
+++ b/glommio/src/channels/channel_mesh.rs
@@ -5,105 +5,22 @@
 //
 use std::cell::Cell;
 use std::fmt::{self, Debug, Formatter};
-use std::result::Result as StdResult;
+use std::io::{Error, ErrorKind};
+use std::iter::repeat_with;
 use std::sync::{Arc, RwLock};
 
-use futures::future;
-
-use crate::channels::local_channel::{self, *};
 use crate::channels::shared_channel::{self, *};
-use crate::{Local, Result};
-
-#[derive(Debug)]
-pub enum Sender<T: Send + Copy> {
-    Local(LocalSender<T>),
-    Connected(ConnectedSender<T>),
-}
-
-impl<T: Send + Copy> Sender<T> {
-    pub async fn send(&self, item: T) -> Result<(), T> {
-        match self {
-            Self::Local(sender) => sender.send(item).await,
-            Self::Connected(sender) => sender.send(item).await,
-        }
-    }
-
-    pub fn try_send(&self, item: T) -> Result<(), T> {
-        match self {
-            Self::Local(sender) => sender.try_send(item),
-            Self::Connected(sender) => sender.try_send(item),
-        }
-    }
-}
-
-enum SenderProvider<T: Send + Copy> {
-    Local(LocalSender<T>),
-    Shared(SharedSender<T>),
-    Taken,
-}
-
-impl<T: Send + Copy> Default for SenderProvider<T> {
-    fn default() -> Self {
-        Self::Taken
-    }
-}
-
-impl<T: Send + Copy + 'static> SenderProvider<T> {
-    async fn get(self) -> Sender<T> {
-        match self {
-            Self::Local(sender) => Sender::Local(sender),
-            Self::Shared(sender) => Sender::Connected(sender.connect().await),
-            Self::Taken => unreachable!(),
-        }
-    }
-}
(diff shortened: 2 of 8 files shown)


glommio/Cargo.toml
```diff
@@ -17,6 +17,7 @@ readme = "../README.md"
 log = "0.4"
 concurrent-queue = "1.1.2"
 futures-lite = "1.11.1"
+futures = "0.3.12"
```
glommer, line 20: We have been trying to standardize on `futures-lite` and avoid using `futures`.
See 6405b59d0de6604d28ac7b857653b3b972a9c6cb for details

glommio/src/channels/mod.rs
```diff
@@ -146,6 +146,44 @@ pub mod local_channel;
 /// [`StreamExt`]: https://docs.rs/futures-lite/1.11.1/futures_lite/stream/trait.StreamExt.html
 pub mod shared_channel;
 
+/// A channel mesh consists of a group of participating executors (peers) and the channels between
+/// each pair of them.
+///
+/// Within a mesh, peers are identified by unique peer ids, which are used to specify the
+/// source/destination which messages are sent to/received from. Peer ids are assigned when the last
+/// peer joins by the indexes to the list of all peers, which is sorted by their executor ids. So
+/// multiple meshes can be built over the same set of peers, and the peer ids are guaranteed to be
```
glommer, line 155: but does that mean that all meshes have to be built at once too ?

glommio/src/channels/channel_mesh.rs
```diff
@@ -0,0 +1,312 @@
...
+            let receiver_provider = ReceiverProvider::Shared(receiver);
+            (Cell::new(sender_provider), Cell::new(receiver_provider))
+        }
+    }
+
+    async fn join(self) -> (Senders<T>, Receivers<T>) {
+        let mut notifiers = self.notifiers.write().unwrap();
+
+        if notifiers.len() == self.nr_peers {
+            panic!("The channel mesh is full.");
```
glommer, line 215: best to error here than to panic.

glommio/src/channels/channel_mesh.rs
```diff
@@ -0,0 +1,312 @@
...
+            notifiers.insert(index, Notifier::new(None));
+            for (peer_id, notifier) in notifiers.iter_mut().enumerate() {
+                if let Some(sender) = notifier.sender.take() {
+                    sender.connect().await.send(peer_id).await.unwrap();
+                }
+            }
+            drop(notifiers);
+            index
+        } else {
+            let (sender, receiver) = shared_channel::new_bounded(1);
```
glommer, line 233: if I understand correctly, this effectively means that only one message at a time can be sent, right ?

Full disclosure, I have a similar implementation in my project using glommio internally. It's fairly specific to what we are doing so I didn't consider moving to glommio. Yours is more generic so that's good!! (I don't need a full mesh)

However it is hard to achieve  good throughput with a single slot in the sender.

glommio/src/channels/channel_mesh.rs
```diff
@@ -0,0 +1,312 @@
...
+
+use futures::future;
+
+use crate::channels::local_channel::{self, *};
+use crate::channels::shared_channel::{self, *};
+use crate::{Local, Result};
+
+#[derive(Debug)]
+pub enum Sender<T: Send + Copy> {
+    Local(LocalSender<T>),
```
glommer, line 19: Why do we need a `LocalSender`? I supposed I will have understood by the time I am don reading, but it would be nice to have a description of the channel architecture for new readers anyway.

The description should include: the role of the local sender, how do messages flow, how do responses return, etc.

glommio/src/channels/channel_mesh.rs
```diff
@@ -0,0 +1,312 @@
...
+    Connected(ConnectedSender<T>),
+}
+
+impl<T: Send + Copy> Sender<T> {
+    pub async fn send(&self, item: T) -> Result<(), T> {
+        match self {
+            Self::Local(sender) => sender.send(item).await,
```
glommer, line 26: Okay, so are you using the `LocalChannel` just to have the option of sending this both locally or remotely?

What would the value be for using this over local channels?

glommio/src/channels/channel_mesh.rs
```diff
@@ -0,0 +1,312 @@
...
+            nr_peers,
+            notifiers: Arc::new(RwLock::new(Vec::new())),
+            providers: Arc::new(Self::build_providers(nr_peers, channel_size)),
+        }
+    }
+
+    fn build_providers(nr_peers: usize, channel_size: usize) -> ChannelProviders<T> {
+        (0..nr_peers)
+            .map(|i| {
+                (0..nr_peers)
```
glommer, line 190: does this mean we are also constructing a channel to ourselves? nr_peers x nr_peers.

Should probably exclude the diagonal of the matrix.

glommio/src/channels/channel_mesh.rs
glommer: Ok, this seems to exist only for notifications. I still haven't understood the role of notifications, though.

glommio/src/channels/channel_mesh.rs
```diff
@@ -0,0 +1,312 @@
...
+    #[should_panic]
+    fn test_join_more_than_capacity() {
+        do_test_channel_mesh(1, 1, 2);
+    }
+
+    fn do_test_channel_mesh(nr_peers: usize, channel_size: usize, nr_executors: usize) {
+        let mesh_builder = MeshBuilder::new(nr_peers, channel_size);
+
+        let executors = (0..nr_executors).map(|_| {
+            LocalExecutorBuilder::new().spawn(enclose!((mesh_builder) move || async move {
```
glommer, line 316: What I think I am missing is the relationship between the mesh peer id with the executor id.

How am I supposed to use this in a real life scenario? Let's say that I am implementing a sharded key-value store.
Usually I will hash the key with some hashing algorithm, then % by the number of shards, then send to "shard 2" if `x % y == 2`.

But I see no translation happening between the mesh peer id and any id that the user could refer to.

glommer: Btw, a very simple memcached-like sharded key-value is a great thing to add to examples if you are up to it!

glommer (commented):
Overall it's a great direction, thanks!

It's great to add primitives that simplify building of applications like that.
I have some questions about how this is supposed to be used.

I've been also thinking if we should support creating other types of architectures aside of a mesh, like trees (with N senders and M receivers). What is your opinion on that? I wouldn't force you to implement other architectures, but if this is something we could reasonably have, then the API should include type-selection (with only mesh supported) from the start.

glommio/src/channels/mod.rs
thirstycrow: Meshes could be built at any time. Any peer will have the same peer id across all meshes, as long as they have the same set of peers, because peer id is determined by the sorted order of its executor id.

glommio/src/channels/channel_mesh.rs
thirstycrow: The notification exists because peers need to wait for the last one to join to get their final order (peer id) in the mesh. Another way to implement this could be a polling loop to check the fullness of the mesh with a short sleep between each iteration. It could be simpler than the notification way.

glommio/src/channels/channel_mesh.rs
thirstycrow: > What I think I am missing is the relationship between the mesh peer id with the executor id.
> 
> How am I supposed to use this in a real life scenario? Let's say that I am implementing a sharded key-value store.
> Usually I will hash the key with some hashing algorithm, then % by the number of shards, then send to "shard 2" if `x % y == 2`.
> 
> But I see no translation happening between the mesh peer id and any id that the user could refer to.

Peer ids are intended to be used for sharding directly. Translation is not necessary. I'll add some documentation to make it clear.

glommio/src/channels/channel_mesh.rs
thirstycrow: My intention was to provide a uniform interface in cases when users may want a local channel to implement staged event-driven architecture. But indeed those are minor cases and it introduces complexity with all the enums to unify the local and remote cases. I'm now considering removing the local support,  or at least make it optional. Which way would be better?

thirstycrow:
> I've been also thinking if we should support creating other types of architectures aside of a mesh, like trees (with N senders and M receivers). What is your opinion on that? I wouldn't force you to implement other architectures, but if this is something we could reasonably have, then the API should include type-selection (with only mesh supported) from the start.

I would suggest a design of deciding whether to create a channel by the result of a custom filter on tags pre-assigned to peers on each side, for this purpose.

glommio/Cargo.toml
thirstycrow: Do you mind if I change the `connect` methods on `SharedSender`/`SharedReciever` a little bit like this, so that connections could be fired at once, and be waited one by one without potential deadlock, thus the dependency on `join` could be removed.

```rust
    pub fn connect(mut self) -> impl Future<Output = ConnectedSender<T>> {
        let state = self.state.take().unwrap();
        let reactor = Local::get_reactor();
        state.buffer.connect(reactor.id());
        let id = reactor.register_shared_channel(Box::new(enclose! {(state) move || {
            if state.buffer.consumer_disconnected() {
                state.buffer.capacity()
            } else {
                state.buffer.free_space()
            }
        }}));

        let reactor = Rc::downgrade(&reactor);
        let peer = Connector::new(state.buffer.clone(), reactor.clone());
        
        async move {
            let notifier = peer.await;
            ConnectedSender {
                state,
                id,
                reactor,
                notifier,
            }
        }
    }

```

glommio/src/channels/channel_mesh.rs
thirstycrow: > Btw, a very simple memcached-like sharded key-value is a great thing to add to examples if you are up to it!

There's a problem when I'm trying to write such an example. Messages passed through shared channels needs to be `Copy `, which is not possible for things like `String`. A walkaround  would be to wrap it with `Arc`. But I'm afraid it would introduce another memory indirection without benefit, which I'm not sure whether the compiler will optimize away. Would it be better to support non-copyable messages with a little bit of unsafe code (`ptr::read`, like the way in `VecDeque`) when trying to pop messages from the spsc_queue buffer?

glommio/Cargo.toml
glommer: another alternative is to fire the connections `Task::local`, detach, and wait for the result. That is what I am doing in my N:M non-mesh implementation.

glommio/src/channels/channel_mesh.rs
glommer: I don't mind the notification system. Just wanted to understand its role.

glommio/src/channels/channel_mesh.rs
glommer: For now I think we should keep it simple and remove the local case.
By using the same interface we're forcing `Send` on types that will pass through the local channel which is not good anyway.

glommio/src/channels/channel_mesh.rs
glommer: Because it is just an example we can use integers.
Passing strings is the wrong thing to do anyway because you will incur a copy which can be expensive (imagine you have a 1MB blob as a value, you don't want to be copying that around in the channel).

What you do in this situation is pass a pointer, and have some facility to make sure that the original data structure is alive.

My implementation of the 1:M ( I said N:M, but my use case really is more like 1:M) is a request-reply channel. I can send requests, and receive replies.

First, you should consider adding that to your implementation, if not always, at least optionally: If we want to be generic, in most times where we pass data to the other executor we will like to know if the operation on the data was successful. (Imagine you are passing buffers for another executor to do I/O on: you want to know if the I/O succeeded).

But now because the channel is request-reply, I keep the  big object (String) until the reply comes.
So you can pass a pointer through the channel, and the receiver knows that it will be alive until the reply is sent.

glommer:
> > I've been also thinking if we should support creating other types of architectures aside of a mesh, like trees (with N senders and M receivers). What is your opinion on that? I wouldn't force you to implement other architectures, but if this is something we could reasonably have, then the API should include type-selection (with only mesh supported) from the start.
> 
> I would suggest a design of deciding whether to create a channel by the result of a custom filter on tags pre-assigned to peers on each side, for this purpose.

Sounds good in principle. I would like to see how does that end up

glommio/src/channels/channel_mesh.rs
glommer: An even more obvious example of why request-reply is beneficial:
- you send a `GET` request to one executor
- now you need to reply =)

glommio/Cargo.toml
thirstycrow: `Task::local` requires it argument have a static lifetime, which is quite annoying. I tried removing it and glommio get compiled without problem. So why is it necessary for the future to be 'static?
```rust
    pub fn local(future: impl Future<Output = T> + 'static) -> Task<T>
    where
        T: 'static,
    {
        LOCAL_EX.with(|local_ex| local_ex.spawn(future))
    }
```

laa: Yes, good question. I wanted to discuss that too. For thread::spawn that is understandable. But in our case we copy future into the heap during the spawn, so we support static lifetime ourselves without requiring it to be static on the language level (we use reference counting). So IMHO it can be removed. @glommer WDYT ?

glommer: it needs to be static because you can `.detach` it. If you don't `.detach` it you can probably figure out a lifetime for it but at task creation you don't know if it will be detached or not.

You can experiment with that for yourself: it is easy to get into a situation where you remove the `'static` requirement, which allows you to now pass an `async {}` closure that uses a reference from the outside. Add a timer to the task and access that reference after the timer. Outside the `Task`, the reference goes out of scope and gets freed -> boom.

thirstycrow: Simplified a little bit to remove dependency on self, now I'm able to use `Task::local`.

glommer:
@thirstycrow have you given some thought about making the channels request-reply?
I can upstream my implentation of a bidirectional channel if it helps you to use as a building block

glommio/src/channels/channel_mesh.rs
thirstycrow: The N:M model is now supported as partial mesh, where a role of either producer or consumer can be specified when joining the mesh.

thirstycrow:
I was planning to use 2 mesh channels

> @thirstycrow have you given some thought about making the channels request-reply?
> I can upstream my implentation of a bidirectional channel if it helps you to use as a building block

Yes, please. I wonder how that is implemented. My original plan is to use two channel meshes for request and reply respectively. Does the bidirectional channel have a advancement to a pair of one-way channels in performance or other aspects?

glommio/src/channels/channel_mesh.rs
thirstycrow: > Passing strings is the wrong thing to do anyway because you will incur a copy which can be expensive (imagine you have a 1MB blob as a value, you don't want to be copying that around in the channel).

Rust `String` just wraps `Vec<u8>`, which is then implemented as a memory block pointed by a `Unique`, with an additional field for length. By using `ptr::write` and `ptr::read`, we can move a shallow copy of a string to the receiving thread without copying the contents, and lift the weight of managing raw pointers in client code. And since the ownership is also passed to the receiving thread, resources will be released there after operation, so no reply will be necessary in this scenario.

glommer:
> > @thirstycrow have you given some thought about making the channels request-reply?
> > I can upstream my implentation of a bidirectional channel if it helps you to use as a building block
> 
> Yes, please. I wonder how that is implemented. My original plan is to use two channel meshes for request and reply respectively. Does the bidirectional channel have a advancement to a pair of one-way channels in performance or other aspects?

I don't think there is any performance advantage of using a bidirectional channel. The main advantage is that you can tie the lifetime of the objects on both sides, as I described, because you know that the reply is coming through the same channel object.

glommer:
FYI: It will take me some time to push the bidirectional channel.

the main reason is - as I said - that my design was fairly specific to my problem as I didn't think there would be broad interest here. The channel I need is fairly low throughput so it is implement as a vector of N channels, each of them with capacity 1. This simplifies the request -> reply tremendously because you can just send, and wait on receive.

For a general purpose implementation we should probably be indexing the requests on top of a channel with more slots which is way, way faster. However that creates a couple of complexities: There needs to be a `Task` constantly running that is pulling from the reply channel, checking the index and pushing requests accordingly.

thirstycrow:
> FYI: It will take me some time to push the bidirectional channel.
> 
> the main reason is - as I said - that my design was fairly specific to my problem as I didn't think there would be broad interest here. The channel I need is fairly low throughput so it is implement as a vector of N channels, each of them with capacity 1. This simplifies the request -> reply tremendously because you can just send, and wait on receive.
> 
> For a general purpose implementation we should probably be indexing the requests on top of a channel with more slots which is way, way faster. However that creates a couple of complexities: There needs to be a `Task` constantly running that is pulling from the reply channel, checking the index and pushing requests accordingly.

@glommer After trying to implement lifetime tying myself in several ways in the past few days, I'm now aware of how much complexity it would bring. I give up finally, and chose the other way I mentioned, to make it possible to pass non-copyable messages via shared channel using `ptr::write` and `ptr::read`. In this way, the ownership of the message is passed to the receiving thread, and it will be dropped there after operation, unless it is send back to the thread in which it is created. So we don't have to track the state of a messege between thread. Referencing to a message from multiple threads can be achieved with `Arc`, which was not possible because `Arc` is not `Copy`able.

I also add and example for implementing sharding with channel mesh, however it fails with high probability, and the cuase seem to be the consuming side being disconnected unexpectedly by some reason not known yet.

glommio/Cargo.toml
laa: But that is possible to lower restrictions any way by providing a version of the method which prohibits detaching of the task.
For example 
```
pub async fn spawn(future: impl Future<Output = T>) 
```
such a task can not be detached and as result does not require a static lifetime. Borrow checker is a pain point for many newcomers so lowering restrictions worth to consider. It is helpful even for more experienced developers. @glommer WDYT ?

BTW local IMHO is not the best name for the method. Would be nice to have `spawn_detachable` and `spawn` instead.

laa: If join handle dies and the task is not stored in weaker holders (I am considering the case that user doesn't call await) then the task is dropped on Drop so it will not outlive internal data. So such an approach seems safe for me.

glommer:
@thirstycrow there was an interesting discussion yesterday in our Zulip channel (Sad to see you're not there, btw!)

Long story short, another user came discussing the `Copy` requirement, and after some discussion we converged on trying to remove it. It would be possible to remove it by passing ownership to the channel: the sender would consume the object and `mem::forget` it, and the receiver can reconstruct it. This way you would be able to pass a string just fine.

Your work is good work, and I find it sad that it gets blocked by other things. I'd like to find a way to reasonably unblock it.

How about you provide a simple example passing integers (I think an example is really good, so people can know how to use it), without worrying with request-reply, copy, and anything like that? With some luck we'll have made some progress on that by the time you are done
(@bryandmc said he would take a stab at the non-copy shared channels )

glommio/Cargo.toml
glommer: I am trying to understand what exactly it is that you are proposing, @laa . Is the idea that you would be able to spawn a task that would start running in the background?

If we can do:
```
let a = spawn(); // tasks starts running
... more async code here

a.await; // waits for a to finish, might have finished already
```

I see value on that.

About `local`, I agree. I took it from `async-task`, that had `spawn` and `local` for a local spawn. However for us, everything is local and it is a bad name.

We're being very lax prior to 1.0 with the API, but `local` is a bit different because it is used a lot by everyone. I would be okay to remove that only after deprecation.

laa: Hi @glommer , let us continue later in community chat. Now I am quite busy with allocator and can not spend enough time to completely analyze the idea.  As for the current moment, I incline to think that better to leave all as is. I need to think more about that topic :-) .

thirstycrow:
> @thirstycrow there was an interesting discussion yesterday in our Zulip channel (Sad to see you're not there, btw!)

Good to know this. I had problems the last time trying to visit Zulip, and it works well right now. I'll check it more often in the future. 

> I find it sad that it gets blocked by other things. I'd like to find a way to reasonably unblock it.

Can you elaborate?

> How about you provide a simple example passing integers (I think an example is really good, so people can know how to use it), without worrying with request-reply, copy, and anything like that? With some luck we'll have made some progress on that by the time you are done

It turns out to be not trivial to build well behaving sharding on top of full meshs, so I added another reusable utility for it. A simplified example is add in `examples/sharding.rs`.

glommio/src/channels/sharding.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        Ok(Self {
+            shard,
+            consumers: Vec::new(),
+            forward_tasks,
+            closed: false,
+        })
+    }
+
+    /// Consume messages from a stream. It will return a [`GlommioError`] if this [`Sharded`] is
```
glommer, line 72: those will result in broken links if I am not mistaken.

You need to add `[GlommioError`]: ...` to the end of the help string.
See other instances of that in the shared channels for example.

It is also helpful to run `cargo docs --all-targets` locally, skim through the documentation, and see if the end result is sound.

glommio/src/channels/sharding.rs
```diff
@@ -0,0 +1,177 @@
...
+use crate::task::JoinHandle;
+use crate::{GlommioError, Local, ResourceType, Result};
+
+/// Trait for handling sharded messages
+///
+/// * `msg` - The message to handle.
+/// * `src_shard` - Id of the shard where the msg is from.
+/// * `cur_shard` - Id of the local shard.
+pub trait Handler<T>: Clone {
+    fn handle(&self, msg: T, src_shard: usize, cur_shard: usize);
```
glommer, line 17: Doesn't this have to be async?
The way this is now works well for integers, but it shouldn't be hard to come up with an example where you want to do some I/O processing in this handler.

Spawning a task results in a memory allocation (although using async-trait does too... oh well)

glommio/src/channels/channel_mesh.rs
```diff
@@ -165,7 +165,7 @@ impl Role {
 }
 
 /// An adapter for MeshBuilder
-pub trait MeshAdapter {
+pub trait MeshAdapter: Clone {
     /// Determine whether a channel should be created between a pair of peers
     /// considering there roles.
```
glommer, line 179: there -> their

glommio/src/channels/spsc_queue.rs
```diff
@@ -115,7 +129,8 @@ impl<T: Copy> Buffer<T> {
             }
         }
 
-        let v = self.buffer_storage[current_head % self.ccache.capacity].get();
+        let cell = self.buffer_storage[current_head % self.ccache.capacity].as_ptr();
+        let v = unsafe { ptr::read(cell) };
```
glommer, line 133: Hi @thirstycrow 

This is a great change, as I mentioned before. But because it is very impactful I'd like to see it in a separate PR.
We should also make @bryandmc aware since he was working on a proposed implementation of that too.

Would you be able to clean this up and send just this part as a separate PR?
Ack the direction!

However I am concerned about `ptr::read` and `ptr::write` here. I just double checked the docs and they say, respectively:

```
Reads the value from src without moving it. This leaves the memory in src unchanged.
```

```
Overwrites a memory location with the given value without reading or dropping the old value.
```

So does that perform a copy? Bryan's approach was without this copy, passing ownership of the object into the channel (although I have not seen the code). For a high throughput channel, that is preferable

glommio/src/channels/sharding.rs
thirstycrow: Either spawning a task or using async-trait is expensive for a single message, I'm afraid. The idiomatic way, IMO, would be unbounded local channels with sync `send` method.

glommer: This is still the main contention point in my view.

I don't see how we can realistically not support async function in the receiver end. The biggest problem is that there is no way to do backpressure with this.

Imagine there is a sender that is 2x as fast as the receiver: how does the receiver throttles the sender? The receiver gets the messages, puts into an unbounded local channel, but the messages keep coming and the channel will eventually run out of memory.

Since the handler consumes like this:

```
                while let Some(msg) = stream.recv().await { <== stream is good to send a new request
                    handler.handle(msg, src_shard, cur_shard)  <== request ends up using more memory in the local channel anyway, but no backpressure
                }
```

I am inclined to think we should just use `async trait` here. It is an allocation, but pushing things into a local channel may allocate memory as well.

For very high throughput use cases where the cost of the allocation is prohibitive batching and sending a vector through the channel seems like a better strategy
(you can't send vectors now, but we already agreed we'll fix it, so let's assume it will be fixed)

glommer:
The example is very good, thanks. Consider adding the following improvements, though:
 https://github.com/glommer/glommio/commit/e374749f3193690a35510304a9c61e2bdff03105

I find it easier to consume for a user interested in examples if we deal with fewer elements, and prints then. Helps figure out what is going on.

I am also adding a `shard_id` function as it is useful here - and in other situations - to know who we are.

glommio/src/channels/spsc_queue.rs
bryandmc: Hey I am on vacation so I haven't had a chance to catch up on all this but I will mention that ptr read and write could be used in the right circumstances to help accomplish the goal. But I'd have to see exactly how it's being used.  Mostly I just wanted to mention that I'm away and won't be able to finish up that code or look into this stuff till Tuesday.

glommer: hope you had fun!

thirstycrow:
@glommer The example is updated following your suggestion.

glommer:
Awesome! Hope you had a great Chinese New Year.
Please pay special attention to my comment about the use of async in the handler. That's the most salient point that we still need to resolve.

thirstycrow:
I made the `Handler` async by returning a boxed future.

glommio/src/channels/spsc_queue.rs
bryandmc: TLDR: Change Cell -> UnsafeCell in the SPSC queue, or enum with destructured data from heap allocated types, and directly copying for copy types. 

Ok so I've gone done a few different paths to see what works best.. First, I went down the path where I left alone the actual queue and it's structure (namely, the copy bound) and managed to make it _more_ ergonomic.. But it has drawbacks. The other method, is replacing the `Cell` with an `UnsafeCell` s/t we can remove the `T: Copy` bound entirely, while still doing internal mutability in a sane way. The drawbacks of the `UnsafeCell` method is that you can have very different sized entries, based on what the user decides to use. But, on the other hand, this is what the standard library does. It gives you that option to put a large struct into one of these data structures, if you wanted to. The other method is pretty simple in theory, but heavy with lots of glue code to make it ergonomic. What the enum method does is create a structure like this: 

```rust
pub enum Shared<T: Send, V: Send + Copy> {
    /// Boxed data
    Box(NonNull<T>),
    /// Strings
    String(NonNull<u8>, usize, usize),
    /// Vectors
    Vec(NonNull<T>, usize, usize),
    /// Copy types that are ok to actually just copy
    Copy(V),
}
```
That includes many of the most common types of heap allocated structures and any data needed to deconstruct/reconstruct, as well as a variant for any type that is Copy, to be stored directly, as is. With a little bit of trait and impl magic you can get to the point where the user's code looks like:

```rust
try_send(String::from("some string data here.."));
try_send(vec!["a".to_string(), "b".to_string()]);
try_send(Box::new(String::from("something here..")));
try_send(100usize);
```
So it can probably be made relatively ergonomic.. That's good. The try_send or any method that needs to take these values can just use a generic or the impl Trait form. (Note: ConvertInto is the trait I had to make to get around the issue of overlapping impls. Basically, can't be don't with normal From/Into.. sadly..)

For example:
```rust
fn try_send<T, V>(i: impl ConvertInto<Shared<T, V>>)
where
    T: Send,
    V: Send + Copy,
{
    // ... do stuff ...
}
```
The UnsafeCell method is actually probably gonna end up being a lot simpler.. All an UnsafeCell has from an API standpoint is `UnsafeCell::get(&self) -> *mut T` which will give you a mutable pointer to it's internals. Then we can safely just move values into that position, and get them back out with `ptr::read`.  The truth is there will probably always need to be ptr::read somewhere in this process.. In any of these options you need the data contained in a Vec struct (for example) to be able to produce the value on the other side.. whether that's with the UnsafeCell or with my custom enum.. either way we probably won't be able to just move out of the Buffer once it gets put there. At least I don't see a way. To be fair, this is how most standard library structures are implemented in rust. For example `Vec::pop(&mut self) -> Option<T>`:

```rust
pub fn pop(&mut self) -> Option<T> {
    if self.len == 0 {
        None
    } else {
        unsafe {
            self.len -= 1;
            Some(ptr::read(self.as_ptr().add(self.len())))
        }
    }
}
```
The only way (that I know) to produce an owned value from a pointer in rust, when the `T` isn't Copy is ptr::read. 

So I'm going to try and finish implementing the UnsafeCell version and see how that goes, first. I've begun implementing the enum version as well, but I'm gonna put that one on the backburner for now because of the increased simplicity of the UnsafeCell version.

NOTE: My original idea _seemed_ to have less copies, but it also proved to be unsound and had ergonomic issues that couldn't be resolved. After studying some other containers in std and in the community, this seemed like the way to go.

thirstycrow:
@glommer I did some benchmark for the average delay to handle 400 million messages from a mesh channel. By returning a boxed local future, the delay increased 4~5 ns per message compared to the sync trait version, which I think is acceptable. Delays for the `async-trait` version are a little bit higher than those for the boxed-local-future version, but still acceptable. I think the boxed-local-future version should be the winner for both a lower overhead and not having to introduce another dependency. Code for the benchmark can be found in `benches/sharding.rs`

| implementation | average delay |
| --- | --- |
| sync trait | 37~44 ns |
| with FutureExt::boxed_local | 43~47 ns |
| with #[async_trait(?Send)] | 48~50 ns |

glommer:
Very good results and cheers for the benchmark!
One small comment I have is that it is very useful for the benchmark to report the cost per operation (you currently only print totals).

I am happy with the current architecture. I'll leave this baking for another day in case someone else has any further comments (or you have last minute fixes) and should be merging this tomorrow.

Thanks!

PR #275, opened 2021-02-17
Support passing non-copyable messages through shared channels

thirstycrow:
### What does this PR do?

Support passing non-copyable messages through shared channels

### DIFF
--- a/glommio/src/channels/spsc_queue.rs
+++ b/glommio/src/channels/spsc_queue.rs
@@ -1,7 +1,7 @@
 use std::cell::Cell;
-use std::fmt;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+use std::{fmt, ptr};
 
 #[repr(align(64))]
 struct ProducerCacheline {
@@ -34,14 +34,14 @@ struct ConsumerCacheline {
 /// Buffer holds a pointer to allocated memory which represents the bounded
 /// ring buffer, as well as a head and tail atomicUsize which the producer and consumer
 /// use to track location in the ring.
-pub(crate) struct Buffer<T: Copy> {
+pub(crate) struct Buffer<T> {
     buffer_storage: Arc<Vec<Cell<T>>>,
 
     pcache: ProducerCacheline,
     ccache: ConsumerCacheline,
 }
 
-impl<T: Copy> fmt::Debug for Buffer<T> {
+impl<T> fmt::Debug for Buffer<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let head = self.ccache.head.load(Ordering::Relaxed);
         let tail = self.pcache.tail.load(Ordering::Relaxed);
@@ -68,36 +68,50 @@ impl<T: Copy> fmt::Debug for Buffer<T> {
     }
 }
 
-unsafe impl<T: Sync + Copy> Sync for Buffer<T> {}
+unsafe impl<T: Sync> Sync for Buffer<T> {}
 
 /// A handle to the queue which allows consuming values from the buffer
-#[derive(Clone)]
-pub(crate) struct Consumer<T: Copy> {
+pub(crate) struct Consumer<T> {
     pub(crate) buffer: Arc<Buffer<T>>,
 }
 
 /// A handle to the queue which allows adding values onto the buffer
-#[derive(Clone)]
-pub(crate) struct Producer<T: Copy> {
+pub(crate) struct Producer<T> {
     pub(crate) buffer: Arc<Buffer<T>>,
 }
 
-impl<T: Copy> fmt::Debug for Consumer<T> {
+impl<T> Clone for Consumer<T> {
+    fn clone(&self) -> Self {
+        Self {
+            buffer: self.buffer.clone(),
+        }
+    }
+}
+
+impl<T> Clone for Producer<T> {
+    fn clone(&self) -> Self {
+        Self {
+            buffer: self.buffer.clone(),
+        }
+    }
+}
+
+impl<T> fmt::Debug for Consumer<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "Consumer {:?}", self.buffer)
     }
 }
 
-impl<T: Copy> fmt::Debug for Producer<T> {
+impl<T> fmt::Debug for Producer<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "Producer {:?}", self.buffer)
     }
 }
 
-unsafe impl<T: Send + Copy> Send for Consumer<T> {}
-unsafe impl<T: Send + Copy> Send for Producer<T> {}
+unsafe impl<T: Send> Send for Consumer<T> {}
+unsafe impl<T: Send> Send for Producer<T> {}
 
-impl<T: Copy> Buffer<T> {
+impl<T> Buffer<T> {
     /// Attempt to pop a value off the buffer.
     ///
     /// If the buffer is empty, this method will not block.  Instead, it will return `None`
@@ -115,7 +129,8 @@ impl<T: Copy> Buffer<T> {
             }
         }
 
-        let v = self.buffer_storage[current_head % self.ccache.capacity].get();
+        let cell = &self.buffer_storage[current_head % self.ccache.capacity];
+        let v = unsafe { ptr::read(cell.as_ptr()) };
         self.ccache.head.store(current_head + 1, Ordering::Release);
         Some(v)
     }
@@ -140,7 +155,8 @@ impl<T: Copy> Buffer<T> {
             }
         }
 
-        self.buffer_storage[current_tail % self.pcache.capacity].set(v);
+        let cell = &self.buffer_storage[current_tail % self.pcache.capacity];
+        unsafe { ptr::write(cell.as_ptr(), v) }
         self.pcache.tail.store(current_tail + 1, Ordering::Release);
         None
     }
@@ -175,7 +191,7 @@ impl<T: Copy> Buffer<T> {
 }
 
 /// Handles deallocation of heap memory when the buffer is dropped
-impl<T: Copy> Drop for Buffer<T> {
+impl<T> Drop for Buffer<T> {
     fn drop(&mut self) {
         // Pop the rest of the values off the queue.  By moving them into this scope,
         // we implicitly call their destructor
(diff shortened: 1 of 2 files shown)


glommio/src/channels/spsc_queue.rs
```diff
@@ -140,7 +155,8 @@ impl<T: Copy> Buffer<T> {
             }
         }
 
-        self.buffer_storage[current_tail % self.pcache.capacity].set(v);
+        let cell = &self.buffer_storage[current_tail % self.pcache.capacity];
+        unsafe { ptr::write(cell.as_ptr(), v) }
```
glommer, line 159: will that not copy the entire data pointed to by cell ?

The documentation for `ptr::write` says:

```
Overwrites a memory location with the given value without reading or dropping the old value.
```

The impression I have from reading the documentation is that this will involve a `memcpy`

thirstycrow: `Cell` by itself is not a pointer type. In the case of `Cell<Arc<T>>`, the `ptr::write` here will copy the a `Arc<T>` to the memory hold by the cell, but will not copy the value of type `T` pointed to by the `Arc`

bryandmc: I didn't originally see this (so I opened one too) but this method is pretty similar to mine. It does copy all the data pointed to by the cell. But usually that's not most of the data, at all. For a Vec that's just the (ptr, len, cap) values in the struct you'd be writing or copying. Not the main data pointed to by that pointer. I would suggest this is just about as efficient as any other method short of being OK with returning &mut T's or raw pointers from the channel. In that case you could avoid even those copies, probably. I thought about a simple deref wrapper around a pointer but at the end of the day, you wouldn't be actually getting what you wanted to send through the channel. You'd get some other type.. which might be ok in some situations. Definitely less versatile if we went that route and it would require the data to be boxed either by the user or by us to get an actual valid pointer we can use across threads. 

It's a little weird they let you write a non-copy T to that cell, even though I guess it's just writing the pointer, or the value the pointer is pointing to (it's the same T, though, isn't it?!).. Still funky IMO. I used UnsafeCell in mine because it's what most the other std library containers use for this exact purpose.

glommer: this is non-obvious and deserves a comment but it's a good direction.

One question I have (which is a bit rhetorical, admittedly): is the memory that represents those data structures always guaranteed to be contiguous? I can't fathom why we would represent something like a string or a vector with non-contiguous memory. Even a B-Tree, the internal pages may be spread, but the BTree object itself will be contiguous. 

Another potential issue: what happens if you pass a type `T` (not `Arc<T>`) that is no `Copy`? somehow this feels to me as potentially dangerous.

thirstycrow: The only thing I was considering when I wrote this PR is to minimize the amount of modifications to the existing code to get a quick solution to passing `Vec<u8>` by which we were blocked when building our sharded cache server. Despite the difference in details, @bryandmc's solution and mine are essentially the same in this sense. However, neither of them is the way it should be eventually, IMO.

As pointed by @bryandmc:
> It's a little weird they let you write a non-copy T to that cell, even though I guess it's just writing the pointer,

I guess that `Cell` was introduced to make the spsc_queue `Buffer` with only safe code. Now, since we agree that some unsafe code is unavoidable for non-copy messages, why should we still keep it? Instead of `Vec<Cell<T>>` or `Vec<UnsafeCell<T>>`, the spsc_queue `Buffer` should be able to be built around `RawVec<T>`, just like how `VecDeque` is implemented. The spsc_queue `Buffer` is a specialization of `VecDeque` optimized for SPSC scenarios, IMO.

thirstycrow: > One question I have (which is a bit rhetorical, admittedly): is the memory that represents those data structures always guaranteed to be contiguous? I can't fathom why we would represent something like a string or a vector with non-contiguous memory. Even a B-Tree, the internal pages may be spread, but the BTree object itself will be contiguous.

I don't get the point why the data structure is contiguous or not matter, can you elaborate?
 
> Another potential issue: what happens if you pass a type `T` (not `Arc<T>`) that is no `Copy`? somehow this feels to me as potentially dangerous.

If `T` is not `Copy`, then it must refer to some non-shareable resource underneath, and since we're going to using it in different threads, it should be wrapped in `Arc`, right?

bryandmc: There's no reason, from a safety standpoint, that it matters if there is an Arc around the T. Reason being, we aren't keeping around the original T. Whether it's contiguous, sure maybe not. But I just honestly don't think it's possible to move the value from one thread to the other, using this queue without this technique. The only other technique would be to de-structure all these heap allocated values manually with impls. I can show what that would look like but it's messy and not completely generic. You need an impl for each kind of container, box, string. Only other idea I came up with was getting the pointer using an &mut T, forgetting T, and then attempting to read that T out again, but there's just no other way short of maybe returning &mut T or transmuting potentially to get an owned value back on the receiver side. 

Now the RawVec<T> idea is interesting for sure... and I do think it's a good point about the VecDeque being implemented that way. I'd have to look into it again to see what the benefits would be.

bryandmc: to kinda mark out the flow to show why you don't need Arc for lets say.. a String..

String -> ptr::write(src, String) (doesn't call drop for String, or src) -> ptr::read(src) -> String. 

NOTE: There is risk of leaking / not calling destructors for items that are left inside the channel, when the channel is dropped (with this impl). So that might need to be addressed in the drop Impl.  You are gonna need to iterate through remaining items in the drop of the Channel and properly drop them. But also don't wanna drop items that aren't technically there anymore, but the pointer remains in the cell (as it does with this design).

glommer: Good point about the destructors -> we should have unit tests to make sure they are called as well.

Both implementations are very similar as you noticed, and I think the differences were ironed out well in this thread.
In particular I like the direction with `RawVec`.

It's up to the two of you, but my proposal is that @bryandmc takes this to the finish line. Two main reasons:
1. his implementation has a slightly broader test coverage (should get even better with tests for Drop) and is already using unsafe accessors.
2. This will free your time for us to figure out the last remaining details of the mesh.

Let me know if that works for the two of you

bryandmc: I should mention regarding the contiguous thing.. For a Vec (for example), the struct that controls the vec is frequently just on the stack unless put inside another container like a Box/Rc/Arc. So that stuff wouldn't be contiguous anyways, would it?

bryandmc: I'm fine with that, if @thirstycrow is on board. Does that mean you are OK with UnsafeCell @glommer ? I'm seeing here that RawVec is unstable and requires a feature flag. Hey -- we never actually expand this ring buffer do we? Cause we can just use a _boxed_ array in this case. This is what I'm thinking: 

### Code:
```rust
pub(crate) struct Buffer<T: Copy> {
    buffer_storage: Arc<[UnsafeCell<Option<T>>]>,

    pcache: ProducerCacheline,
    ccache: ConsumerCacheline,
}
```

glommer: I am okay with the `UnsafeCell` for this case.
Bummer about the `RawVec` - the feature flag is a deal breaker because I want glommio to be always usable with stable.

We don't expand the array once it is created.

thirstycrow: @bryandmc Please go ahead with your work. I'll be glad to switch to your final version once it's done.

thirstycrow: > ```rust
> pub(crate) struct Buffer<T> {
>     buffer_storage: Arc<[UnsafeCell<Option<T>>]>,
> 
>     pcache: ProducerCacheline,
>     ccache: ConsumerCacheline,
> }
> ```

@bryandmc Please also consider a `RawVec` like implementation with `Unique<T>`, to save the space for discriminants introduced by `Option<T>`.

bryandmc:
I'm wondering if you get to save on the discriminant because of the null
pointer optimization rust does in certain cases.. I know it does it for raw
pointers. But yeah I can easily swap it out for Unique, it's a good thought
and might be better for the optimizer.

https://stackoverflow.com/questions/46557608/what-is-the-null-pointer-optimization-in-rust#:~:text=The%20null%20pointer%20optimization%20basically,space%20as%20that%20associated%20value%2C

This link makes it seem possible but I'll look into it further and if not,
just try and use Unique alone.

thirstycrow:
> I'm wondering if you get to save on the discriminant because of the null pointer optimization rust does in certain cases.. I know it does it for raw pointers. But yeah I can easily swap it out for Unique, it's a good thought and might be better for the optimizer. https://stackoverflow.com/questions/46557608/what-is-the-null-pointer-optimization-in-rust#:~:text=The%20null%20pointer%20optimization%20basically,space%20as%20that%20associated%20value%2C This link makes it seem possible but I'll look into it further and if not, just try and use Unique alone.

Sorry, I didn't make it clear. What I wanted to say is that, a custom container like `RawVec` which operates on pointers should be used instead of array. The problem of array is that when it's destroyed, destructors for each of the remaining elements will be called. Since we are build a ring buffer, elements get popped by copying and moving the tail cursor, but will still stay in the underlying array. To prevent double freeing, it's necessary to replace the popped element with default value. That's what the `Option` is for. Besides the discriminant, an additional write will have to be done each time an element is popped.

glommer:
We don't run destructor on vector destruction because the buffer `Drop` implementation is careful enough to do `set_len(0)` under an unsafe block.

I don't understand where the extra write is coming from when elements are popped.

thirstycrow:
> We don't run destructor on vector destruction because the buffer `Drop` implementation is careful enough to do `set_len(0)` under an unsafe block.

It will not be the case any more if `Arc<[UnsafeCell<Option<T>>]>` is used for buffer storage.

glommer:
This is fine. I am just trying to understand where does the cost come from when using the `Vec` approach.
* I understand why the double free is a problem, but argued that it is not an issue if you `set_len`
* your other statement claims that there is an extra write on pop with `Vec`: I don't see where this write is coming from.

bryandmc:
To be clear I ended up not going with the array method because there's no
good way without a bunch more unsafe to directly create an array on the
heap (there's unstable features, but none stable yet w/ out excessive
unsafe). So the option was able to be eliminated by using the drop impl we
already had that @glommer is referring to. I understand the position with
the options cause it means you don't need a drop impl but that isn't a big
enough of a benefit to really bother with so I think where it ended up is
probably good.

thirstycrow:
> your other statement claims that there is an extra write on pop with Vec: I don't see where this write is coming from.

It's also base on the assumption that array of generic type is used directly for buffer storage. In that case, something like 'Cell<Option<T>>' will have to be used as the element type to prevent double freeing, and elements will be popped by `Cell::take` which will write a `None` to the position of the popped value.

PR #276, opened 2021-02-17
Non copy shared channel

bryandmc:
### What does this PR do?
Removes `T: Copy` bound from shared channel. Discussed this in another pull request where conversation started but basic idea is we change the SPSC queue's `Vec<Cell<T>>` to a `Vec<UnsafeCell<Option<T>>>` s/t we can remove the need for the copy bound. These require copies, but if you have a large struct, that's on the stack, it'll need to be put somewhere to give to another thread anyways, so you should do what you are recommended to do in the std library: stick it in a box, then send that. So despite some efforts to somehow eliminate these copies, it didn't turn out to be as possible as it seemed originally. This seems like the preferred pattern, as it's used inside the rust std library for it's containers. This doesn't end up being materially different from the method where we de-structure the heap allocated and container types to their parts (ptr, len, cap) for String/Vec for example isn't different from just copying that small struct, containing the same items, into the UnsafeCell. The process is the same but in reverse for the de-structure method, which again is the same as copying that struct out of the UnsafeCell, and returning that owned value to the user on the other side.

### Motivation
Discussion on another issue and worries that the Copy constraint is hard for newcomers and just annoying for everyone else.

### Related issues
#275   - not sure if this method would be preferred. 

### Additional Notes
It looks like this is addressed similarly in another pull request. Not sure there's any difference in the safety or performance or anything like that between the two options.. They are fundamentally pretty similar, but I'm submitting this anyways because I spent time doing it and figure we might as well have both options to compare ;-p

### Checklist


glommer:
Thanks Bryan.

In general if something works without unsafe / unsafecell it is a preferred direction. Aside from that both implementations seem very similar.

It would be good if you would reach consensus on the details. On my side I am just trying to make sure we are not doing anything wrong in any of the corner cases. Perhaps unit tests are granted for those?

Examples: what happens if you pass a `non-Copy` type without an `Arc`?  What happens if you pass a pointer object? Does it end up doing a deep copy ?

bryandmc:
Yeah so that's fine. Though there's nothing safer about using a raw pointer in the cell vs UnsafeCell, though I actually just didn't know it was possible in a cell. I assumed the T: Copy would be enforced across the board in the Cell. 

I pass non-copy types without Arc's in my tests, but I can add further ones... But if the Cell seems better I can close this one. I also prefer not to use unsafe, but I didn't know this could be done with the Cell. I guess we just have to make sure there aren't any gotchas, but I don't really think there would be.

bryandmc:
Incorporated many of your suggestions.. Managed to remove an unnecessary layer of `Arc` and of course the Option. Added a bunch of tests that exercise and verify we are dropping everything properly. There might be more optimizations we can do, but frankly I think it's a good, safe place to start. Because all the optimizations would likely require more unsafe.

glommio/src/channels/spsc_queue.rs
```diff
@@ -36,7 +38,7 @@ struct ConsumerCacheline {
 /// use to track location in the ring.
 #[repr(C)]
 pub(crate) struct Buffer<T> {
-    buffer_storage: Arc<Vec<UnsafeCell<Option<T>>>>,
+    buffer_storage: Vec<UnsafeCell<T>>,
```
glommer, line 41: good catch, this is indeed redundant.

glommio/src/channels/shared_channel.rs
```diff
@@ -762,4 +762,179 @@ mod test {
...
+        let ex1 = LocalExecutorBuilder::new()
+            .spawn(move || async move {
+                let sender = sender.connect().await;
+                for x in 0..1000 {
+                    let val = WithDrop(send_count.clone(), x);
+                    drop_count.fetch_add(1, Ordering::Relaxed);
+                    match sender.send(val).await {
+                        Ok(_) => {}
+                        Err(_) => {
+                            // dbg!(err);
```
glommer, line 792: if this is an unexpected condition we should panic. Tests should be extremely reliable and as few surprises as possible

glommer:
Very minor comment on tests and the `dbg` comment (should be nothing or assert)

Replacing the `Vector` with another more efficient data structure is just polishing at this point, and can be done later.
I am also not convinced it will buy us much anyway. Won't block the merge.

Let me know your thoughts on the test nitpick above, and I'll merge right after.

glommio/src/channels/shared_channel.rs
bryandmc: Oh the errors are actually OK in these tests because we aren't interested in if it's closed or not yet, we just wanna make sure everything drops, no matter where the items are at the time of the drops. So I'll clean up those match statements cause they take up too much space.

glommer (approved):

PR #277, opened 2021-02-17
add yolo accept()

fujita:
### What does this PR do?

Firstly, try to accept in non-blocking way, if it fails, let's use io_uring.

### Motivation

improve the performance of accepting new connections.

### Related issues

for this Pull Request.

### Additional Notes

Probably I should not touch backlog of accept. I'll revert the change if you prefer.

### Checklist


glommer:
hi! Notice that our CI requires `cargo clippy` to pass, and this doesn't.
Aside from that - as we spoke on twitter - patch is good

glommer:
thanks!

PR #278, opened 2021-02-18
Add SQE flags and try implementing TcpStream::connect_timeout on top

mxxo:
### What does this PR do?

This PR adds a `SQE` flags field to `UringDescriptor` and attempts to use it to implement `TcpStream::connect_timeout`. However, the timeout test hangs forever. I think there is an problem with how I am submitting two SQEs for the new `SourceType::ConnectTimeout` source [(link)](https://github.com/DataDog/glommio/commit/6b54d25b0a73a8105a5f0747554d0f3bec671db7#diff-d8d2d815ed457d48361ae92c5976d152059d3f04d59dea4ad148d1ea8a75f21aR1410-R1433), or the completion queue handler has to be adjusted. I would appreciate any suggestions.

The implementation is a little hacky, but even if we decide to not go ahead with adding (an actually working) `connect_timeout`, I thought I'd submit this anyways in case similar issues surface for other SQE linking cases.

### Motivation
Having a way to link SQEs would be useful.

### Related issues
#243

### Additional Notes

[Link to Zulip discussion](https://glommio.zulipchat.com/#narrow/stream/262273-general/topic/connect_timeout/near/222465933)

### Checklist


glommio/src/sys/mod.rs
```diff
@@ -365,6 +365,7 @@ pub(crate) enum SourceType {
     Statx(CString, Box<RefCell<libc::statx>>),
     Timeout(TimeSpec64),
     Connect(SockAddr),
+    ConnectTimeout(SockAddr, TimeSpec64),
```
glommer, line 368: This is unrelated, but FYI: I. have a pending patch that removes `SockAddr` from `SourceType` and replaces it with `Box<SockAddr>`. Those larger fields are bloating the size of `SourceType` for everybody.

No action needed now, but keep an eye out for changes in the near future as this may conflict with this change.

glommer:
What do you find hacky about this implementation?

The biggest problem I see is that you are queueing the two operations in the right order, but you are not necessarily sending them in the right order. io_uring can fill up (we should have tests covering this scenario), so you may not be able to fetch an sqe from the ring. Usually this is fine: what is left to submit is just submitted in the next period. However for linked SQEs you have to submit both.

One possible way of implementing this is when grabbing a sqe, if you notice the link flag immediately grab another before filling up the first. If you fail to grab the next, fill the first with a NOP and give up.

In theory chains can be arbitrarily long, so just apply the same logic recursively.

One way of testing this is to carefully pre-fill the queue with N-1 requests (say, long expiring timeouts) and then submit your linked request. io_uring allows (although we don't use that) for the number of sqe entries to be configurable so we could add that too to make sure the test is robust against future default changes.

Hope you're having fun!

mxxo:
Thanks for taking a look! By hacky I just meant the special case for `ConnectTimeout` in `queue_request_into_ring`. Even just running this single test hangs, so I thought maybe there was an issue with reaping the timeout result and returning an error to the source type, but first I will make sure it's getting submitted correctly.

bryandmc:
I don't know that I'd characterize it as hacky either. I'm wondering if we can't find way to build the linked-timeout feature in a way where it can be applied to _any_ operation, generically. Almost like a wrapper we can add to any operation. Linking in general is even more generic and probably a very important feature that's worth looking into. 

I haven't dug into exactly why your implementation of it isn't working -- I'd suggest doing some tracing, logging, debugging, whatever to understand what's happening on the submission, and then on the completion, side of things and go from there. 

If ask on the Zulip, maybe someone can help a little more in real-time.

glommer:
The way he built it was per my advice , and I think it is the way to go: 
- the uring layer is generic and supports linking for any pair of sqe operations
- the upper layer is specific, so we can enforce different semantics for different operations.

bryandmc:
Oh, I wasn't saying his SHOULD have that. More just brainstorming the ideal api for timeouts and linking.. and it would be neat if we could just wrap operations generically, or something like that.

(very vague example)
```rust
let read_result = tcpstream.read(&mut buf);
let linked = Link::timeout(read);
linked.await?;

let linked = Link::read_write(read_stream, write_stream);
let write_result: usize = linked.await?;
```
Just spit-balling API's. Cause having different functions for everything is doable but might actually become more work eventually. That said, this also looks like it might be tricky to implement..

Linking would be nice so that if you just wanna pipe data around from different sources without touching it at all.

I'll make an issue for linking if you want, when my plate clears a little more

mxxo:
> I haven't dug into exactly why your implementation of it isn't working -- I'd suggest doing some tracing, logging, debugging, whatever to understand what's happening on the submission, and then on the completion, side of things and go from there.

I did some testing at the `iou` level and think this is almost certainly from not submitting both sqes at the same time.

> One possible way of implementing this is when grabbing a sqe, if you notice the link flag immediately grab another before filling up the first. If you fail to grab the next, fill the first with a NOP and give up.

It might be necessary to wait until we get a more recent `iou` to implement this (I don't think you can have two live `SQE` references obtained from `ring.next_sqe()` at the same time because they are mutable refs to the submission queue). However there is a new [`prepare_sqes` API](https://github.com/ringbahn/iou/blob/045f8d44c235e35b3731625356a11e6ee05966f9/src/submission_queue.rs#L155-L170) that lets us request a number of sqes at once. Then, we could have something like: 

```rust 
    fn submit_one_event(&mut self, queue: &mut VecDeque<UringDescriptor>) -> Option<bool> {
        if queue.is_empty() {
            return Some(false);
        }
        
        // find position of first element without a link
        let chain_len = queue.iter()
            .position(|&sqe| !sqe.flags.intersects(SubmissionFlags::IO_LINK | SubmissionFlags::IO_HARDLINK))
            .unwrap();
        
        // catch links off the end of the queue -- internal source prep bug
        assert!(chain_len < queue.len() - 1);

        if let Some(mut sqes) = self.ring.prepare_sqes(chain_len + 1) { 
            for mut sqe in sqes { 
                self.submitted += 1;
                let op = queue.pop_front().unwrap();
                let allocator = self.allocator.clone();
                fill_sqe(&mut sqe, &op, move |size| allocator.new_buffer(size));
            }
            Some(true)
        } else { 
            None
        }
    }
```

glommer:
Thanks! Liburing should have a new release this week, and there is an open discussion [here](https://github.com/DataDog/glommio/issues/283) about what to do with `iou` (we may be forced to fork it).

One way or another we should bring in a new version so that API should be available soon.

glommer:
@mxxo I just merged `iou` into glommio - thanks again for your review.
This should unlock the API you need. If you need any fixes for the `iou` API to make that happen I'll take them too - as long as we make a good effort to upstream them (which I know you will)

mxxo:
Fantastic, thank you for the integration work :)

mxxo:
Just force-pushed the revised implementation and updated the PR description to be more optimistic.

glommio/src/sys/uring.rs
```diff
@@ -1545,4 +1567,23 @@ mod tests {
             None => {}
         }
     }
+
+    #[test]
+    #[should_panic(expected = "Unterminated SQE link chain")]
```
glommer, line 1590: what about a test for successful chaining ? 

Another interesting case is what happens when the chain is bigger than the the amount of SQEs the ring is configured for.

If I read your code correctly we'll end up waiting forever and hanging in that case. Erroring out the entire chain would be better in that case (granted this is very, very unlikely especially right now. But once eBPF lands into io_uring, I expect user-defined chains that can get arbitrarily long)

glommio/src/net/tcp_socket.rs
```diff
@@ -398,6 +399,68 @@ impl TcpStream {
...
+        let inet = InetAddr::from_std(&addr);
+        let addr = SockAddr::new_inet(inet);
+
+        let reactor = Local::get_reactor();
+        let source = reactor.connect_timeout(socket.as_raw_fd(), addr, duration);
+
+        // connect_timeout submits two sqes to io_uring: a connect sqe soft-linked
+        // with a LINK_TIMEOUT sqe. If the timeout fires, the connect sqe fails with
+        // ECANCELED. We map that error to TimedOut to match the standard library's API.
+        source
```
glommer, line 442: fantastic, very happy with the result.

glommer (changes requested):
Canadian grade AAA work Max, really good!

A few more tests suggested for corner cases. 
There are conflicts now, and I suspect it is because of the PR I just merged from Bryan.
We are experimenting with using nightly for `cargo fmt` because it is like, 100x more useful. (it formats even the example code).

Please rebase on that and then it will be ready to go.

glommio/src/sys/uring.rs
mxxo: For successful chaining, I was having trouble coming up with a good small test case (and I hoped the `connect_timeout` tests would partially cover it). The too-large chain is a good one, I think you're right about it hanging since `prepare_sqes` should fail over and over. I'm not sure if there's an easy way to get queue sizes.

glommer: the queue size is determined at creation. Right now I am open coding it, you can just make it into a const.

mxxo: Thank you for reviewing, I added this in ca28b7e!

glommio/src/sys/uring.rs
```diff
@@ -683,12 +683,22 @@ impl UringCommon for PollRing {
...
-            let op = queue.pop_front().unwrap();
-            let allocator = self.allocator.clone();
-            fill_sqe(&mut sqe, &op, move |size| allocator.new_buffer(size));
+        // find position of first element without a link
+        let chain_len = match queue.iter().position(|sqe| {
+            !sqe.flags
+                .intersects(SubmissionFlags::IO_LINK | SubmissionFlags::IO_HARDLINK)
+        }) {
+            Some(pos) => pos,
+            None => panic!("Unterminated SQE link chain: internal source prep bug"),
```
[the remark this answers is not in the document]
mxxo: Thank you, fixed as part of ca28b7e

glommio/src/sys/uring.rs
mxxo: There is now a test for successful links in 66e61fa

glommer:
The code looks good to me and architecturally this is what we're after. But some tests fail:

```
failures:
    net::tcp_socket::tests::connect_and_ping_pong
    net::tcp_socket::tests::connect_local_server
    net::tcp_socket::tests::multi_executor_accept
    net::tcp_socket::tests::multibuf_fill
    net::tcp_socket::tests::overconsume
    net::tcp_socket::tests::parallel_accept
    net::tcp_socket::tests::peek
    net::tcp_socket::tests::repeated_fill_before_consume
    net::tcp_socket::tests::stream_of_connections
    net::tcp_socket::tests::test_lines
    net::tcp_socket::tests::test_read_line
    net::tcp_socket::tests::test_read_until
    net::udp_socket::tests::sendto_blocking
    net::unix::tests::connect_local_server
    net::unix::tests::read_until
```

Please note that unfortunately because of kernel version in the CI boxes we can't really run the tests as part of CI (and I didn't yet have the time to search for alternatives), so we have to run them manually.

Let me know if you can reproduce them

mxxo:
I can (and sorry about that: I've been running the new unit tests but had some existing tests lock up, blamed it on my kernel, etc). I did bisect and it looks like d171963 is the bad one. In particular it looks like there's some double mutable borrows failing in `queue_request_into_ring` (e.g. from `sys::Reactor::accept`): 
https://github.com/DataDog/glommio/blob/d17196385be413d2748abeb4ea6e0bfd61bfda2a/glommio/src/sys/uring.rs#L1478-L1486
 I will have to rework that part.

PR #281, opened 2021-02-23
Gate

thirstycrow:
### What does this PR do?

Add a `Gate` similar to the same named facility in seastar to help achieving graceful shutdown of things like tcp connections.

### DIFF
--- a/glommio/src/sync/gate.rs
+++ b/glommio/src/sync/gate.rs
@@ -1,0 +1,188 @@
+use std::cell::{Cell, RefCell};
+use std::rc::Rc;
+
+use futures_lite::Future;
+
+use crate::channels::local_channel::{self, LocalSender};
+use crate::{GlommioError, Local, ResourceType, Task, TaskQueueHandle};
+
+#[derive(Debug)]
+enum State {
+    Closing(LocalSender<bool>),
+    Closed,
+    Open,
+}
+
+/// Facility to achieve graceful shutdown by waiting for the dependent tasks
+/// to complete.
+#[derive(Clone, Debug)]
+pub struct Gate {
+    inner: Rc<GateInner>,
+}
+
+impl Gate {
+    /// Create a new [`Gate`](crate::sync::Gate)
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
+        Self {
+            inner: Rc::new(GateInner {
+                count: Cell::new(0),
+                state: RefCell::new(State::Open),
+            }),
+        }
+    }
+
+    /// Spawn a task for which the gate will wait on closing
+    pub fn spawn<T: 'static>(
+        &self,
+        future: impl Future<Output = T> + 'static,
+    ) -> Result<Task<T>, GlommioError<()>> {
+        self.spawn_into(future, Local::current_task_queue())
+    }
+
+    /// Spawn a task for which the gate will wait on closing
+    pub fn spawn_into<T: 'static>(
+        &self,
+        future: impl Future<Output = T> + 'static,
+        handle: TaskQueueHandle,
+    ) -> Result<Task<T>, GlommioError<()>> {
+        let inner = self.inner.clone();
+        inner.enter()?;
+        Task::<T>::local_into(
+            async move {
+                let result = future.await;
+                inner.leave();
+                result
+            },
+            handle,
+        )
+    }
+
+    /// Close the gate, and wait for all spawned tasks to complete
+    pub async fn close(&self) -> Result<(), GlommioError<()>> {
+        self.inner.close().await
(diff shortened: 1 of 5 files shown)


glommer:
My friend, I don't know where you got your thirsty nickname from, but you sure are hungry!

reviewing.

glommer:
I have no meaningful comments on the gate implementation. It is very good, thanks!

I am struggling a bit to understand why you need the `Pass`. In general all those `Send` structs need a bit of extra scrutiny.
Let's chat about the pass a bit more, the rest is good

bryandmc:
I would just wanna understand what led you to use the raw pointer.. Any reason in particular? Seems like it could be accomplished other ways but I would accept it if the reason was good.. Maybe I'm just missing it from reading through the code. Looks like you just need shared access? Is the issue that you'd need an arc for `Send` and that's overkill? Just wanna understand the tradeoff being made.

Beyond that I think this looks great. Really solid.

thirstycrow:
@bryandmc My original plan was to send a Pass with the request, and send it back with the response. After rethought I realized that it might never be guaranteed that a response will be sent back. So I removed the `Send` trait from `Pass`, and replaced the pointer with an `Rc`.

bryandmc:
All my safety concerns are satisfied, so I say merge it. If we want that functionality across executors we'll just have to come up with a way to do that separately (or by later modifying this). @thirstycrow can you resolve the git conflicts?

thirstycrow:
@bryandmc Code rebased.

glommio/src/sync/gate.rs
```diff
@@ -0,0 +1,188 @@
...
+/// Facility to achieve graceful shutdown by waiting for the dependent tasks
+/// to complete.
+#[derive(Clone, Debug)]
+pub struct Gate {
+    inner: Rc<GateInner>,
+}
+
+impl Gate {
+    /// Create a new [`Gate`](crate::sync::Gate)
+    #[allow(clippy::new_without_default)]
```
glommer, line 38: why not add a `Default` ?
The new implementation looks like what a good default would be. We can probably implement `Default` on `State` (I think it is a good practice to implement `Default` for enums, since you can't derive it, and then likely just `Derive` it for `Gate`.

I will not consider this a blocker because the code is very good otherwise, but please consider sending a follow up

glommer (approved):

glommio/src/sync/gate.rs
thirstycrow: IMO, people would always create `Gate`s explicitly with `new`. Is there a case in which a default value of `Gate` would be used via the `Default` trait?

glommer: There are a variety of functions in rust that expect a default implementation, and may be more convenient to use.
For example, `std::mem::take`

If something can be thought to have an implementation that is a reasonable default, I think we should expose  it.

thirstycrow: There's a risk of the default `Gate` created by `std::mem::take` being used unconscious, and left unclosed at shutdown. If it's real necessary to take an old `Gate` and leave a new one in-place, then `std::mem::replace(&gate, Gate::new())` should be better used to express the users intention explicitly, IMO.

glommer: The risk of leaving the gate unclosed is still there in your example, if the programmer forgets to close after replacing. Sure, it is a better pattern and it is the one I expect to see, but if we are going to force a specific pattern I would expect that pattern to be really fool proof, and this isn't. At least not enough.

Consider the downsides:

Imagine a user has a temporary struct with 100 members, one of them being a gate. User wants to add `[derive(Default)]` and is now unable to as `Gate` doesn't have a default implementation. Is the guarantee given by forbidding `mem::take` enough to justify this hassle? I think that's the question that really matters

thirstycrow: > Imagine a user has a temporary struct with 100 members, one of them being a gate. User wants to add `[derive(Default)]` and is now unable to as `Gate` doesn't have a default implementation. Is the guarantee given by forbidding `mem::take` enough to justify this hassle? I think that's the question that really matters

I really recommend a refactoring to split the struct into smaller pieces in this circumstance and move the gate member to a part where is `Default` is not necessary. If not, a slightly verbose implementation of `Default` could be done, not as simple as `[derive(Default)]` though: 
```rust
impl Default for MyStruct {
    fn default() -> Self {
        Self { gate: Gate::new(), ..Defaults::default() }
    }
}
```

Anyway, I'm not purticularly against the idea of making the `Gate` a `Default`. I'm just not convinced it would be necessary in a real project (not imaginary) with proper design. I think it could be done when a user send a request for it.

Issue #270, opened 2021-02-10
error information is lost


glommer:
Part of the information we encode in errors is lost when we print them.

Take for example this:
```

        if start_id != end_id {
            return Err(GlommioError::<()>::WouldBlock(ResourceType::File(format!(
                "Reading {} bytes from position {} would cross a buffer boundary (Buffer size {})",
                len, self.current_pos, buffer_size
            ))));
        }
```

This very helpful message was supposed to tell us how much we are reading, by how much we failed, etc.

However due to the way we are implementing `Display` for `GlommioError`, the only information that make the output is:

```
thread 'test-1' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: WouldBlock, error: "File operation would block" }',
```

glommer:
@bryandmc can you please take a look ?

bryandmc:
Yeah I'll look at it as soon as I get back tomorrow it's probably an easy
fix. Just gotta tweak the display impl as you suggest.

-Bryan

bryandmc:
Ok so after looking at this I'm starting to think maybe it's worth just getting rid of `thiserror` and doing our own display impl's manually instead of contorting this library to do it. We already did our own impls for other things that it would sometimes offer as well because we had more specific needs so I'm just gonna do that. Bonus: we get to remove a dependency!

PR #282, opened 2021-02-25
Error lost context

bryandmc:
### What does this PR do?
Should be a sufficient fix for issue #270, where we had lost information in the error message. In doing so, I removed the `thiserror` crate, because we ended up having to do all the impls by hand anyways to get the exact behavior we wanted. On the plus side, we probably shaved off some compile time as well by removing all the derive/custom attribute macros. 

### Motivation
Addresses #270 

### Related issues
#270 

### Checklist

[ X ] I have added unit tests to the code I am submitting
[ - ] My unit tests cover both failure and success scenarios (somewhat irrelevant in this case b/c it's about errors)
[ X ] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -34,7 +34,6 @@ smallvec = "1.4.2"
 buddy-alloc = "0.4.1"
 ahash = "0.5.7"
 intrusive-collections = "0.9.0"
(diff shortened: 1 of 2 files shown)


glommer:
Thanks Bryan!

PR #292, opened 2021-03-09
Add rustfmt.toml plus simple defaults

bryandmc:
Use rustfmt to also format comments (doc, and normal), doctests, and imports.
This is a bold PR to make, but the inconsistent formatting was driving me
crazy (as silly as that might be).

### What does this PR do?

I made a rustfmt.toml that'll include comments and other things in the formatting. This may SEEM like it changed a lot but I think it makes the code look a lot more consistent.

### Motivation
_What inspired you to submit this pull request?_

OCD? A sadness inside? it can't be known for sure, but it's definitely a sickness of some sort...

### Related issues
I think it's related to ALL issues and NO issues..

### Additional Notes
_Anything else we should know when reviewing?_ 

This is somewhat subjective, so I won't be offended or hurt if the PR isn't accepted. But I think it uses safe defaults for rustfmt that keep the comments and imports as idiomatic as the rest of the code already is. 

### Checklist


glommer:
Great work @bryandmc . However current CI is complaining that this breaks it.
Any changes that we must do to CI to accompany this?

bryandmc:
I looked at these on CI and I'll have to fiddle with some stuff. I may have accidentally been using nightly rustfmt (yeah, that's a thing!?) features but I can probably change whatever that is, as long as we still get the comment formatting I'll be happy.. Will fix these later today when I  have a minute.

glommer:
I think it is okay to use nightly in CI for this. It's just not ok to depend on nightly code features.

glommer:
if this gets too annoying and people start complaining we can revert. But I think it is a good change and we should give it a try.

PR #302, opened 2021-03-20
Local exec pool builder

trtsl:
### What does this PR do?

Creates a `LocalExecutorPoolBuilder` in preparation for a more complete solution to issue #32.

### Motivation

Address issue #32.

### Related issues

None.

### Additional Notes

Happy to get feedback and make further changes as necessary.  At a minimum, documentation could likely be improved--happy to do so once the interface is fully agreed upon.  There are a few other `TODO`s in the code that should be resolved.

### Checklist

[~] My unit tests cover both failure and success scenarios


glommio/src/executor.rs
```diff
@@ -559,6 +559,276 @@ impl Default for LocalExecutorBuilder {
...
+    /// For the `Unbound` variant, the [`LocalExecutor`]s created by a
+    /// [`LocalExecutorPoolBuilder`] are not bound to any CPU.
+    Unbound,
+    /* MaxSpread,
+     * MaxPack,
+     * Custom, */
+}
+
+/// A description of the CPUs location in the machine topology.
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
```
glommer, line 575: I'd be careful with those traits.
What does it mean to say that a location is greater or lesser than another?

If you need to put them on a map that needs a comparison, it is probably better to use one of the `_with` functions to avoid surprising results.

glommio/src/executor.rs
```diff
@@ -559,6 +559,276 @@ impl Default for LocalExecutorBuilder {
...
+    // /// [`LocalExecutorBuilder::spin_before_park`] for details.  The setting
+    // /// is applied to all executors in the pool.
+    // pub fn spin_before_park(mut self, spin: Duration) -> Self {
+    //     self.spin_before_park = Some(spin);
+    //     self
+    // }
+
+    /// Please see documentation under [`LocalExecutorBuilder::name`] for
+    /// details. The setting is applied to all executors in the pool.
+    pub fn name(mut self, name: &str) -> Self {
```
glommer, line 663: that's usually not what the user wants. Usually we want each thread to have a unique name, so for an executor pool this is likely a function, taking an `id` as a parameter.

glommer (commented):
It looks like a strong and solid start.

glommio/src/executor.rs
```diff
@@ -559,6 +559,276 @@ impl Default for LocalExecutorBuilder {
...
+    ///     println!("hello");
+    /// });
+    ///
+    /// handles.join_all();
+    /// ```
+    #[must_use = "This spawns executors on multiple threads, so you may need to call \
+                  `PoolThreadHandles::join_all()` to keep the main thread alive"]
+    pub fn on_all_shards<G, F, T>(self, fut_gen: G) -> PoolThreadHandles
+    where
+        G: FnOnce(usize, CPUSet) -> F + Clone + Send + 'static,
```
glommer, line 738: what is your intention by passing the `cpuset` inside the function for `on_all_shards` ?

glommio/src/executor.rs
```diff
@@ -559,6 +559,276 @@ impl Default for LocalExecutorBuilder {
...
+    /// Spawn a pool of [`LocalExecutor`]s in a new thread according to the
+    /// [`Placement`] policy, which is `Unbound` by default.
+    ///
+    /// This method is the pool equivalent of [`LocalExecutorBuilder::spawn`].
+    ///
+    /// It takes a closure `fut_gen` which will be executed on each new thread
+    /// to obtain the [`Future`] to be executed there.  The `fut_gen`
+    /// closure is passed two arguments: a `usize` and a [`CPUSet`].  The
+    /// `usize` provides a sequential id of the threads created by this
+    /// [`LocalExecutorPoolBuilder`] (i.e. numbered in the range `0..nr_shards`
```
glommer, line 708: I think we should avoid making promises about what the ID will be.
Once we connect channels to this (a certainly good add), we may want to use an ID that makes using channels simpler.

I'd take inspiration on the `std::thread` IDs, where we promise that an executor id is unique, but don't say anything about what to expect from it.

glommio/src/executor.rs
trtsl: Agreed, this is arbitrary.  I used this in parsing `sysfs` but will remove it.

glommio/src/executor.rs
trtsl: This comment is actually somewhat misleading anyway.  As with the `LocalExecutorBuilder`, the name will ultimately be combined with a notifier id, so it would be something like `name-5`.  I'm happy to modify this--maybe storing a `Box<dyn Fn(usize) -> String>` in the builder (I'm not aware of an ergonomic way to use a closure as an optional field with static dispatch).

glommio/src/executor.rs
trtsl: This is purely for users' convenience / information.  It would allow the user to modify execution based on where the thread will be running (e.g. maybe a particular package/socket has a larger cache so the user might take advantage of that in some way, or threads running on the same core have some preferred work-stealing arrangement, or the user has some method of detecting panics and would use this information to quickly restart a thread that went down ... just spitballing).  I can remove this if you believe it just clutters the interface.

glommio/src/executor.rs
trtsl: I'll tie this to the notifier / reactor id (`sys::ReactorGlobalState`) for now.  Let me know if you would rather have a totally unique id and I'll create a new global.

trtsl: ... and i'll change the thread id visible to the user from `usize` to some opaque `ThreadId` type

glommer: Take a look at how the `mesh_channels` work. They assign a peer_id based on the order in which the executors were connected to the mesh.

Long term, I would like to provide a mesh by default with this group of executors and the Ids of the mesh and executor should match.

I actually think it is okay if they are of different types, if that gives us flexibility in implementation, as long as they compare to the same thing.

glommio/src/executor.rs
glommer: Without an example of how this would be used I think it is clutter.

glommio/src/executor.rs
glommer: "name-x" is okay.
No need for a function in this case. This is a convenience interface, so there is no need to go for ultra flexibility.

Thanks!

glommio/src/executor.rs
trtsl: I was still a bit unsure what the approach here should be since `std::thread::ThreadId` are globally unique but the ids created by `mesh_channels` would only be unique for each builder.  I've revised this, for now to just use a `usize`.

trtsl:
General comment: I was hoping to button down the hardware layout work a little more, because that might drive additional changes to e.g. `CpuLocation`, but I might need some feedback there (I'll check in on zulip).

glommio/src/executor.rs
glommer: A usize is fine, we just shouldn't promise what they are.

Also a per-builder id is perfectly fine. My idea is that once you are done with this.
Let's keep in mind that the main goal of this builder is convenience: so my idea is that the natural next step is to attach a Mesh builder automatically, that can be used to send messages to all other executors that were created together.

The per-builder id in this case is an advantage, not a liability: it allows for that id to be used to refer to message destinations without any translation.

glommer:
I took a look now. At first I was going to comment that we don't really want to pass a cpu set: we want a specific CPU, drawn from a set.
But it is not unreasonable to have a policy that will allow for cpu movement (no pin), but within a restrained set. So that's fine.

I think the direction is very good. It is essentially up to you if you want to work more on it before merging to add at least one pin policy or if you'd like to do it in pieces. I'l okay with both.

trtsl:
Yeah, I made this a cpu set because one of the examples in #32 is "Give me one executor in each socket in this box", which I assumed meant to allow the executor to execute anywhere in the socket, but I guess that example was more like "one particular cpu in each socket in this box".

glommer:
Right

glommer:
But now that you added it, a set policy is not useless: it is essentiall the same as `Unbounded` (cpus can move), but restricted to a particular subset of all cpus.

glommio/src/executor.rs
trtsl: One other thing I was wondering was whether the `usize` id should be passed to `fut_gen` at all.  If we don't promise anything about what the `usize` is then the user might not have much use for it in their `fut_gen` (they're just random numbers to the user).  And we already have e.g. `Local::id()` that the user can use to for identification.  We could also provide an analogous `Local::mesh_id()`.

glommer:
Hello @trtsl 

Just to make sure I'm not missing anything, what's your short term goal here: are you planning to add more code before merging, or merge a checkpoint version for now ?

trtsl:
Hi Glauber, thanks for checking in.  The only question I think worth addressing before merging is the one linked below since it affects the public api.  After that I can rebase and squash and we can merge what we have so far.

I should have something for you to review on the remainder of the pool builder fairly soon, but I had to rework some things to allow numa nodes to reside above or below packages / sockets in the hardware topology (I didn't realize packages can contain multiple numa nodes e.g.: [AMD EPYC](https://developer.amd.com/wp-content/resources/56308-NUMA%20Topology%20for%20AMD%20EPYC%E2%84%A2%20Naples%20Family%20Processors.PDF)).

Open question:
https://github.com/DataDog/glommio/pull/302#discussion_r605125581

glommio/src/executor.rs
glommer: right.

It makes sense that we don't pass it. We can already grab the current executor ID as a thread local, and can do the same with the mesh id when we get there

trtsl:
Let me know if you're ok to merge with this last change, and I'll rebase and squash.

glommer (approved):
great stuff! I'll merge the rebased version

PR #304, opened 2021-03-23
Mesh improvements

glommer:
### What does this PR do?

Improves the mesh channel:
* better docs site visualization
* better error description
* add send and send_to endpoints so we don't need to always create a stream

### Motivation

It felt a bit unergonomic in cases where a stream doesn't exist.


glommer:
@thirstycrow might want to take a quick look

duarten (commented):
LGTM!

thirstycrow:
> @thirstycrow might want to take a quick look

LGTM

PR #306, opened 2021-03-26
Add rust/glommio version info to bug template

jeffwidman:
When triaging bug reports, it's extremely helpful to know both the Rust
version and the Glommio version.

This isn't a big deal now, but will be as the project matures and grows
in popularity.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,2 +1,7 @@
 name: Bug Report
 about: Something is not working as expected
+
+Environment:
+
+- Rust version:
+- Glommio version:


jeffwidman:
Failing test is completely unrelated given that this only modifies a markdown file.

glommer:
You're right.

That's the error:

```

  nightly-x86_64-unknown-linux-gnu installed - rustc 1.53.0-nightly (52e3dffa5 2021-03-25)

info: checking for self-updates
error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly. If you don't need the component, you can remove it with:

    rustup component remove --toolchain nightly --target x86_64-unknown-linux-gnu rustfmt
```

We've been doing an experiment recently where we are trying to format glommio with nightly's fmt, since it is so much more capable than stable's.

@bryandmc do you have any idea about the above ? Why wouldn't nightly have a fmt now ?

PR #307, opened 2021-03-26
Switch to `cimg`-based images for CI

jeffwidman:
CircleCI deprecated the `circleci/rust` image in favor of `cimg/rust`:
* https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/
* https://circleci.com/developer/images/image/cimg/rust

According to their docs this results in faster, more deterministic
builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 5
 jobs:
   cargobuild:
     docker:
-      - image: circleci/rust:1.51.0
+      - image: cimg/rust:1.51.0
     steps:
       - checkout
       - run:
@@ -30,7 +30,7 @@ jobs:
 
   cargofmt:
     docker:
-      - image: circleci/rust:1.51.0
+      - image: cimg/rust:1.51.0
     steps:
       - checkout
       - run:
@@ -62,7 +62,7 @@ jobs:
 
   cargodoc:
     docker:
-      - image: circleci/rust:1.51.0
+      - image: cimg/rust:1.51.0
     steps:
       - checkout
       - run:
@@ -94,7 +94,7 @@ jobs:
 
   cargoclippy:
     docker:
-      - image: circleci/rust:1.51.0
+      - image: cimg/rust:1.51.0
     steps:
       - checkout
       - run:


jeffwidman:
Looks like a untarring is failing due to `usr/local/cargo` doesn't exist. I'm both a circleci and rust n00b, so is this supposed to exist? 

Or is this something that we need to manually create as part of the circleci workflow before executing the jobs?

Googling isn't showing me a whole lot so far...

glommer:
To be honest I am not the most knowledgeable person about circleCI either.
Most likely from that description there is no permission to create a directory in whichever current directory we are at the moment. `tar` would create the full directory structure.

jeffwidman:
Once I figure out what's wrong with the Tar stuff, then need to also update to `1.51.0` as there's now a release out: https://github.com/CircleCI-Public/cimg-rust/pull/53

glommer:
Fantastic. I think I know what's wrong with the tar stuff. You have to bump the cache version here:

```
          key: v5-cargo-cache-{{ arch }}-{{ checksum "Cargo.lock" }}
```

(and other similar ones)

jeffwidman:
Thx! Just rebased/force-pushed. And we have green! 🚀 

For `key: v5-cargo-cache-{{ arch }}-{{ checksum "Cargo.lock" }}`, I have googled and keep coming up empty... any idea if this is a cache key protocol built directly into cargo, or is it something more just generally for circleCI, or even specific to just this repo's CI implementation and has to be bumped whenever there's a major version bump of rust? I'm just curious to know what things do/don't result in needing to bump the cache key/invalidate the cache... even if you don't know all the answers, if you have any idea of where I should start that'd be helpful as my google foo is seriously lacking.

glommer:
I believe this is not because of rust versions but because of the transition from one image type to another.
The cache is generated one way in one image type (for instance, owned by a specific user), but that user may not exist in the new image type.

Anyhow it's not like I know a lot about CI, that is just a summary of what I found around earlier.

glommer (approved):

PR #308, opened 2021-03-26
Bump to `rust` `1.50.0`

jeffwidman:
While `rust` `1.51.0` just landed, it's so new that [it's not yet
available in the `cimg/rust` registry](https://github.com/CircleCI-Public/cimg-rust/issues/52).
So for now I bumped to `1.50.0`.

Depends on https://github.com/DataDog/glommio/pull/307, that should be merged first and then this rebased.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2
 jobs:
   cargobuild:
     docker:
-      - image: circleci/rust:1.49.0
+      - image: cimg/rust:1.49.0
     steps:
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


jeffwidman:
I did wonder if it might be simpler to only pin to minor releases (`1.50`) rather than the full version including patch. It'd make for slightly less-deterministic builds, but also slightly less maintenance, and for most open source projects that I've been involved in we've found that to be a worthwhile trade off. Especially because if something starts failing, we can always dig deeper by pinning the specific versions at that point in time.

jeffwidman:
Closing due to superceded by https://github.com/DataDog/glommio/pull/316

PR #311, opened 2021-03-26
Try re-enabling deadlinks checking

jeffwidman:
This was originally disabled back in https://github.com/DataDog/glommio/commit/7daa35a7a7d379c7a5259aa446d8385241bd453b.

That was four months ago, so perhaps the upstream issues have been fixed
and we can re-enable this now. Worth a shot... 😄

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,9 +44,8 @@ jobs:
           rustup install nightly
           cargo +nightly doc --all
 
-    # temporarily disabled.
-    # - name: Validate links
-    #   run: cargo deadlinks --dir target/doc/glommio
+      - name: Validate links
+        run: cargo deadlinks --dir target/doc/glommio
 
   build:
     runs-on: ubuntu-latest


jeffwidman:
Also if this doesn't pass, then probably the `install cargo-deadlinks` step should be commented out as well, since it's pointless to install if it's not actually used. Puppies die.

jeffwidman:
https://app.circleci.com/pipelines/github/DataDog/glommio/810/workflows/5be9f666-b1b5-41cc-b5a2-cdfa61c57fc9/jobs/2886 is an issue flagged by `deadlinks`, although I'm too much of a rust n00b to know if that's a legit issue or something spurious (ie, why this was commented out in the first place).

The other failure is completely unrelated to this PR, and is simply that `rustfmt` broke on `nightly`... I filed https://github.com/DataDog/glommio/issues/312 to track fixing it so that it doesn't cause spurious test failures like this.

glommer:
See if you can run it locally first, and get `echo $?` to return 0, which I guess is how CI would test for this.
The problem I had last time is that after an update, it started to complain about completely unrelated stuff, from other crates.

This one at least seems to point at something in the glommio code, but I am not really sure what that means

jeffwidman:
This is pretty old and I'm abandoning it as I'm not currently working much in Rust, anyone else feel free to pick this up if you want.

Issue #317, opened 2021-04-01
"kernel witnessed Romulus killing Remus" on AWS Ubuntu 20.04 kernel 5.4.0


qiujiangkun:
It's weird, since linux 5.4.0 isn't that old at all. Any idea how to debug?
```
ubuntu:~/glommio$ cargo run --example hello_world
thread 'main' panicked at 'Failed to register a probe. The most likely reason is that your kernel witnessed Romulus killing Remus (too old!!)', glommio/src/sys/uring.rs:209:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
ubuntu:~/glommio$ sudo dpkg --get-selections |grep linux-image
linux-image-5.4.0-1038-aws                      install
linux-image-5.4.0-1039-aws                      install
linux-image-5.4.0-1041-aws                      install
linux-image-aws                                 install
ubuntu:~/glommio$ grep io_uring_setup /proc/kallsyms
0000000000000000 t io_uring_setup
0000000000000000 T __x64_sys_io_uring_setup
0000000000000000 T __ia32_sys_io_uring_setup
0000000000000000 d event_exit__io_uring_setup
0000000000000000 d event_enter__io_uring_setup
0000000000000000 d __syscall_meta__io_uring_setup
0000000000000000 d args__io_uring_setup
0000000000000000 d types__io_uring_setup
0000000000000000 t __event_exit__io_uring_setup
0000000000000000 t __event_enter__io_uring_setup
0000000000000000 t __p_syscall_meta__io_uring_setup
0000000000000000 t _eil_addr___ia32_sys_io_uring_setup
0000000000000000 t _eil_addr___x64_sys_io_uring_setup
ubuntu:~/glommio$ grep -i uring /boot/config-$(uname -r)
CONFIG_IO_URING=y
```

glommer:
5.4 is old.
The minimum kernel required to run glommio is 5.8
io_uring presence is not the only requirement: we also test for individual io_uring operations that are used by glommio.

qiujiangkun:
This should be mentioned somewhere more explicitly

glommer:
Fair point.

Where would you suggest, that would have helped you the most? In the README? In the error message? Both ?

qiujiangkun:
I would first see the error message and then README, but if I can read it in README in the first place, I couldn't have seen the error message. So I prefer both.

glommer:
Thank you!
[...]

PR #318, opened 2021-04-01
bring lib.rs document to README & add minimal linux kernel version

qiujiangkun:
### What does this PR do?

bring lib.rs document to README & add minimal linux kernel version

### Motivation

Currently there is no minimal linux kernal version in documentation. 

### Related issues

#317


glommer:
Hello! Cargo fmt is failing.

I think this is because we are using nightly's cargo, which also format comments and examples.
Please re-run with `cargo +nightly fmt --all` and I'll merge.

Thanks a lot

PR #320, opened 2021-04-05
allow for remote wakers: enable use side by side with tokio

glommer:
### What does this PR do?

Allows users to use glommio executors as well as tokio executors.
This was already technically possible, by connecting both executors with a channel. However because our wakers want to be woken in the same executor they were originated, this would crash on asserts.

This PR allows wakers to be manipulated in foreign executors, by reference counting them atomically (with the state machine still being local), and passing them back to their owners if a remote wake is triggered.

### Motivation

My dear friend @thirstycrow 

### DIFF
--- a/glommio/src/task/raw.rs
+++ b/glommio/src/task/raw.rs
@@ -257,13 +257,6 @@ where
             return;
         }
 
-        assert_eq!(
-            Self::thread_id(),
-            Some(raw.my_id()),
-            "Waker::wake_by_ref is called outside of working thread. Waker instances can not be \
-             moved to or work with multiple threads"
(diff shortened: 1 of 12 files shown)


glommio/src/task/raw.rs
```diff
@@ -204,36 +204,27 @@ where
```
jeffwidman, line 204: Just curious on this commit message... is `Jiang` a tool of some kind like Valgrind? Or a human being reporting an issue?

glommer: Jiang is one of our contributors, that contributed our mesh implementation

Issue #319, opened 2021-04-05
call wake directly instead of accumulating it in a vector


glommer:
In `parking.rs`, we accumulate wakers in a vector and wake them all at the end. This is usually the right thing to do, because wakers can be reentrant. But in our case we don't need to do this, because our wakers won't actually run anything, just schedule a task for running (see the code in `raw/task.rs`) so there is no danger of reentrancy.

The complex thing here, is that the code in `uring.rs` uses the fact that there are entries in the vector to learn whether or not it can sleep. This keeps things simple because new wakers can be generated at any time (before we even call `wait`, from cancellations if the ring is full, etc).

However we can probably replace that with a counter.

bryandmc:
I think this issue can be closed due to #327

glommer:
it does

PR #327, opened 2021-04-15
Wake directly without accumulating waker

waynexia:
Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

### What does this PR do?

Wake directly without accumulating waker

### Related issues

#319 

### Additional Notes

The root change is removing `Reactor::wakers`. And call `Waker::wake()` instead of accumulating waker.

In the case of needing to count how many `wake()` are called, I accumulating the result in a mutable parameter rather than querying the vector's size.

But I think there might have some problems. In my environment, some test cases will hang *sometimes*. I pick one `producer_sleeps_before_consumer_consumes` to debug and find `submit_sqes_and_wait()` from `SleepableRing::sleep()` will not be wake up. So I mark this PR as draft for now. I'll continue on this and appreciate it if someone could help.

### Checklist

### DIFF
--- a/glommio/src/lib.rs
+++ b/glommio/src/lib.rs
@@ -301,13 +301,23 @@ extern crate scopeguard;
 use crate::parking::Reactor;
 use std::{fmt::Debug, time::Duration};
 
+/// Call `Waker::wake()` and log to `error` if panicked.
+macro_rules! wake {
+    ($waker:expr $(,)?) => {
+        use log::error;
+
+        if let Err(x) = panic::catch_unwind(|| $waker.wake()) {
+            error!("Panic while calling waker! {:?}", x);
+        }
+    };
+}
+
 mod free_list;
 
 #[allow(clippy::redundant_slicing)]
 #[allow(dead_code)]
 #[allow(clippy::upper_case_acronyms)]
 mod iou;
-#[macro_use]
 mod parking;
 mod sys;
 pub mod task;
--- a/glommio/src/lib.rs
+++ b/glommio/src/lib.rs
@@ -307,6 +307,7 @@ mod free_list;
 #[allow(dead_code)]
 #[allow(clippy::upper_case_acronyms)]
 mod iou;
+#[macro_use]
 mod parking;
 mod sys;
 pub mod task;
--- a/glommio/src/parking.rs
+++ b/glommio/src/parking.rs
@@ -641,21 +641,21 @@ impl Reactor {
         Ok(woke > 0)
     }
 
-    fn process_external_events(&self) -> Option<Duration> {
-        let next_timer = self.process_timers().0;
-        self.process_shared_channels();
-        next_timer
+    fn process_external_events(&self) -> (Option<Duration>, usize) {
+        let (next_timer, mut woke) = self.process_timers();
+        woke += self.process_shared_channels();
+        (next_timer, woke)
     }
 
     /// Processes new events, blocking until the first event or the timeout.
     fn react(&self, timeout: Option<Duration>) -> io::Result<()> {
         // Process ready timers.
-        let next_timer = self.process_external_events();
+        let (next_timer, woke) = self.process_external_events();
 
         // Block on I/O events.
         match self
             .sys
-            .wait(timeout, next_timer, || self.process_shared_channels())
+            .wait(timeout, next_timer, woke, || self.process_shared_channels())
         {
             // Don't wait for the next loop to process timers or shared channels
             Ok(true) => {
--- a/glommio/src/parking.rs
+++ b/glommio/src/parking.rs
@@ -26,7 +26,6 @@
 
 use crate::iou::sqe::SockAddrStorage;
 use ahash::AHashMap;
-use log::error;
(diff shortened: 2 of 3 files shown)


glommio/src/parking.rs
```diff
@@ -592,78 +604,62 @@ impl Reactor {
...
+        woke += self.process_timers().1;
+        woke += self.process_shared_channels();
+
         Ok(woke > 0)
     }
 
-    fn process_external_events(&self, wakers: &mut Vec<Waker>) -> Option<Duration> {
-        let next_timer = self.process_timers(wakers);
-        self.process_shared_channels(wakers);
+    fn process_external_events(&self) -> Option<Duration> {
```
glommer, line 644: This looks wrong to me, and likely the reason why tests aren't passing.
Nowhere you return a count of wakers from process_external events, and nowhere the fact that we have wakes here is fed to `wait`.

Keep in mind this is used in the decision of whether or not to sleep: if we had any wake, any wake at all, at any point, we must not sleep. If we woke, and sleep, we may never get to run the future and that will present itself as a hang in the test.

* if you have a waker *before* you called `wait`, you must not sleep
* if you have a waker *during* the call to `wait` (for example if you had to flush the cancellation queue, or because a remote shared channel notification atomically appeared) you must not sleep.

glommer (commented):
Thanks for tackling this!

There is at least one problem, which is the fact that you don't keep waking information from before `wait` (the return of `process_external_events`) and `wait`.

We must keep wait information for the entire call to `react`: if any wake happened at any point we must give up sleeping.

waynexia:
Thanks for your time!

I have rechecked that requirement. I think it is ready for review now.

glommio/src/lib.rs
```diff
@@ -307,6 +307,7 @@ mod free_list;
 #[allow(dead_code)]
 #[allow(clippy::upper_case_acronyms)]
 mod iou;
+#[macro_use]
```
glommer, line 310: I find it a bit better to put the macros in lib.rs, and avoid the `macro_use`. this is because cargo fmt can reorder imports, and unfortunately macros are sensitive to the order of imports.

glommio/src/sys/uring.rs
```diff
@@ -454,32 +455,30 @@ fn transmute_error(res: io::Result<u32>) -> io::Result<usize> {
...
-            return Some(());
+            return Some(false);
         }
 
         let src = consume_source(from_user_data(value.user_data()));
 
         let result = value.result();
 
+        let mut woke = false;
         if try_process(&*src).is_none() {
```
glommer, line 473: I think this needs work as well.
`try_process()` essentially gives the opportunity for the ring to process events that are internal to the ring architecture and therefore may be handled differently (one general comment about that: as uring.rs evolved, there is probably a way now to remove this special case!)

However if those events fire, I suspect we may also have to forego sleeping.

This code is a bit complex, so let's be careful: due to cancellations, it *may* be that `LinkRings` actually doesn't have to prevent waking. I have to inspect that code a bit more carefully again.

However, even if that's the case, I'd recommend that `try_process` returns a boolean in this case, and `LinkRings` return false.

glommer (commented):
Overall very good!

I left one minor comment about the macro placement, and another a bit more serious about `try_process`. It's a good exercise for you to try to follow the code that relies on it, and judge if the individual processors should return true or false. But in any case `try_process` should return a boolean and let that decision to the ring

Thanks!

glommer:
Also note that we're likely deleting the special handling of `Timeout`: https://github.com/DataDog/glommio/pull/330

glommio/src/sys/uring.rs
waynexia: From my understanding, `try_process()` is for the ring to decide whether an event should be waking or not. And in the current code, both events `LinkRings` and `Timeout` have no waker inside, which means removing the `try_process()` parameter will not change the ring's behavior. I tried it and the test cases pass (several times).

glommio/src/lib.rs
waynexia: I moved it lib.rs. But apart from other macro rules to make it is visible to mod `parking` and `sys`.

glommio/src/sys/uring.rs
waynexia: > However if those events fire, I suspect we may also have to forego sleeping.

We always sleep on the main ring, therefore I think it is OK to not consider this situation as we won't see the `LinkRings` event of this time "going to sleep"?

glommer: I think we should not indeed increment the waker on `LinkRings`. My main concern is with the interface.

However, let's leave this as is: I'll actually try to get rid of this special case altogether.
With the timer gone in the PR I linked, it's only one more to go...

glommer (approved):

glommer:
I am merging this. Thanks!

waynexia:
Thanks.

And one more thing, do we need to remove the entire `try_process()` parameter?

glommer:
let's leave `try_process` as is, and hopefully soon we can remove the entire thing

waynexia:
okay🤗

PR #328, opened 2021-04-16
Pool builder placement

trtsl:
### What does this PR do?

Implements `MaxSpread` and `MaxPack` placement strategies for `LocalExecutorPoolBuilder`.

### Motivation

Address issue #32.

### Related issues

None.

### Additional Notes

A bit large, but separated into files which should be fairly modular.

### Checklist

[~] If applicable, I have discussed my architecture


glommio/src/executor/mod.rs
```diff
@@ -577,58 +579,24 @@ pub enum Placement {
     /// For the `Unbound` variant, the [`LocalExecutor`]s created by a
     /// [`LocalExecutorPoolBuilder`] are not bound to any CPU.
     Unbound,
-    /* MaxSpread,
-     * MaxPack,
+    /// Iterates over `CpuLocation`s in the machine topology with a high degree
+    /// of separation from previous `CpuLocation`s returned by the iterator.
+    /// The order in which items are returned by `MaxSpread` is
+    /// non-deterministic.  If the number of CPUs requested is larger than the
```
glommer, line 585: why not fail in this case?
If we're asking for a hard bind, I think it is fine to fail in the case where asked_cpus > avail_cpus

glommio/src/executor/mod.rs
```diff
@@ -818,6 +799,16 @@ impl PoolThreadHandles {
         self.handles.push(handle)
     }
 
+    /// Returns the number of threads that were successfully spawned.
+    pub fn nr_spawn_ok(&self) -> usize {
+        self.handles.iter().filter(|r| r.is_ok()).count()
+    }
+
+    /// Returns the number of threads that failed to spawn.
+    pub fn nr_spawn_err(&self) -> usize {
```
glommer, line 809: I am not sure I am too hot on this, but I see the point.
Ideally, if one of the threads failed to spawn, we should fail the whole thing.

It doesn't seem that easy to do, as we can't just panic, but maybe we can `cancel()` the successful handles.

The challenge is separate failed execution - which is fine - from failed spawn.

glommio/src/executor/mod.rs
```diff
@@ -577,58 +579,24 @@ pub enum Placement {
...
-    /* MaxSpread,
-     * MaxPack,
+    /// Iterates over `CpuLocation`s in the machine topology with a high degree
+    /// of separation from previous `CpuLocation`s returned by the iterator.
+    /// The order in which items are returned by `MaxSpread` is
+    /// non-deterministic.  If the number of CPUs requested is larger than the
+    /// number of CPUs online, this `Placement` will cycle through
+    /// `CpuLocation`. Each cycle is non-deterministic meaning that
+    /// the order may differ from prior cycles.
+    MaxSpread,
```
glommer, line 589: Should those take an `Optional` parameter with a cpu list? The idea is that you may want to spread maximally (or minimally), but only within a subset of the CPUs.

I think this is particularly important in a scenario where you are sharing the node.

glommer (commented):
I am really happy with the direction, and in particular the lack of dependency on hwloc is a big plus.

A couple of things that I commented on the public API, which is important to get right.

glommio/src/executor/mod.rs
trtsl: sounds good -- will revise

glommio/src/executor/mod.rs
trtsl: I'll revise this as well -- I plan on having all threads wait until they have all been successfully spawned; if any fail to spawn, those that have spawned will exit without ever creating their local executor.  Hopefully, this way the user won't have to wonder about what has and has not been executed so far.  Let me know if you prefer a different approach.

glommio/src/executor/mod.rs
trtsl: Yeah, this sounds like a good idea -- will update as well.

trtsl:
Great, thanks for the feedback!  I'll make the changes and we can go from there.

trtsl:
Hello, I've made several changes.  In particular, there is a breaking change with the previous implementation of `on_all_shards`:
* `on_all_shards` now returns a `Result` which is an `Err` if there are insufficient CPUs for the desired placement
* i'm allowing the user to retrieve the values returned by the provided `fut_gen`, which requires the return values to be `Send + 'static` (new bound); i don't see a good reason not to do this, but didn't modify the existing `LocalExecutorBuilder` since that's been around a while (wanted to get your thoughts)
* the `PoolThreadHandles` was modified to accommodate the changes above

glommer:
looking, thanks!

glommio/src/executor/latch.rs
```diff
@@ -0,0 +1,227 @@
...
+
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc,
+    Condvar,
+    Mutex,
+};
+
+#[derive(Clone, Debug)]
+pub(crate) struct Latch {
```
glommer, line 23: I am still reading through, so take this with a grain of salt: but I suspect this would be a useful primitive to have in `glommio::sync`! It's essentially the async version of a barrier.

I don't anticipate widespread use (the same way barriers aren't used that widely either), but I can definitely see them being useful in some circumstances.

Won't affect the decision of what we do with this PR, but if you agree consider moving it later as a `pub` interface.

glommer: Ok, I already see that it has mutexes internally which make it harder to use in generic code.
We could have a version of that over shared channels, but that's for later.

It's still a good idea, and I'll consider opening a new issue for that.

Thanks!

glommio/src/executor/mod.rs
```diff
@@ -572,75 +579,145 @@ impl Default for LocalExecutorBuilder {
...
+    /// `SharedOnCpus` variant allows the number of shards specified in
+    /// [`LocalExecutorPoolBuilder::new`] to be greater than the number of CPUs
+    /// as long as at least one CPU is included in `CpuSet`.
+    ///
+    /// #### Errors
+    ///
+    /// If the provided [`CpuSet`] contains no CPUs, a call to
+    /// [`LocalExecutorPoolBuilder::on_all_shards`] will return `Result::
+    /// Err`.
+    SharedOnCpus(CpuSet),
```
glommer, line 635: this is a bit inconsistent with the rest

While you have `MaxPack` and `MaxPackOnCpus`, the `OnCpus` version of `Unbound` is... `Shared`?

glommio/src/executor/mod.rs
```diff
@@ -29,6 +29,12 @@
 
 #![warn(missing_docs, missing_debug_implementations)]
 
+mod latch;
+mod placement;
```
glommer, line 33: I don't mind moving executor.rs to a directory, but two things:

1. it's a conflict magnet, so I'd recommend doing that separately in a PR that can go in quickly. There are already conflicts due to recent changes in executor.rs, so be careful to carry them over.
2. please move multitask.rs as well, since it's only purpose is to isolate some of the functions of the executor. There may even be a case for moving parking.rs as well, although I am a bit more on the fence with that one, as there are functions from parking.rs used all over the place.

glommio/src/executor/mod.rs
```diff
@@ -572,75 +579,145 @@ impl Default for LocalExecutorBuilder {
...
+    /// is less than the number of CPUs in the provided `CpuSet` then some
+    /// CPUs will not be used.
+    ///
+    /// #### Errors
+    ///
+    /// If the number of shards specified in [`LocalExecutorPoolBuilder::new`]
+    /// is greater than the number of CPUs available, then a call to
+    /// [`LocalExecutorPoolBuilder::on_all_shards`] will return `Result::
+    /// Err`.
+    MaxSpreadOnCpus(CpuSet),
```
glommer, line 661: What is the justification for the `OnCpus` in your opinion?
My first impression for this is that we could just have `MaxSpread` take an `Option`:
if the `Option` is `None`, behave like your current `MaxSpread`, and if the `Option` contains `Some(CpuSet)`, then abide by that

Your placement code would then be:

```
            Placement::MaxSpread(cpus) => {
                let cpus = cpus.unwrap_or(CpuSet::online()?);
                Self::check_nr_cpus(nr_shards, &cpus)?;
                Self::MaxSpread(MaxSpreader::from_cpu_set(cpus))
            }
```

glommio/src/executor/mod.rs
```diff
@@ -659,7 +736,7 @@ pub struct LocalExecutorPoolBuilder {
     /// How often to yield to other task queues
     preempt_timer_duration: Duration,
     /// Indicates a policy by which [`LocalExecutor`]s are bound to CPUs.
-    placement: Placement,
+    placement: Option<Placement>,
```
glommer, line 739: Why is this an option? Is there any situation in which we expect this to be `None` ?
It seems to me that especially with the fact that `Placement` has a default, we could just use it straight away

Am I missing something ?

glommio/src/executor/mod.rs
```diff
@@ -731,109 +808,128 @@ impl LocalExecutorPoolBuilder {
...
+        let handle = Builder::new().name(name).spawn({
+            let io_memory = self.io_memory;
+            let preempt_timer_duration = self.preempt_timer_duration;
+            let spin_before_park = self.spin_before_park;
+            let latch = Latch::clone(&latch);
+
+            move || {
+                // only allow the thread to create the `LocalExecutor` if all other threads that
+                // are supposed to be created by the pool builder were successfully spawned
+                if latch.arrive_and_wait() == LatchState::Ready {
```
glommer, line 861: this is very good.

glommio/src/executor/placement/linux_sysfs.rs
```diff
@@ -0,0 +1,837 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
glommer, line 1: there's already `glommio/src/sys/sysfs.rs`.
We should keep those things in a single location, and `src/sys` is better for that.

Sorry I haven't noticed this before, I think this code was already here in a previous version

glommer (commented):
My comments are very minor at this point, but potentially impactful since some are user-visible (like the `OnCpus` thing). If it wasn't for that I'd even be willing to merge now and address later.

We're getting there, but if it's not too much work I'd kindly ask you to merge the executor-in-a-directory change first so you don't risk more conflicts.

Also, as a heads up, take a look at https://github.com/DataDog/glommio/pull/337

I plan to cut a release today, and seeing how this API is still a bit in-flight, I'm considering adding an experimental feature flag to it.

It shouldn't affect your work much, except you may have an extra source of conflicts in executor.rs

glommio/src/executor/mod.rs
trtsl: Caution -- somewhat lengthy:

I initially set up `Placement` as:

```rust
pub enum Placement {
    Unbound(Option<CpuSet>),
    MaxSpread(Option<CpuSet>),
    MaxPack(Option<CpuSet>),
}
```

But doing the implementation and docs it felt like `Unbound(Option<CpuSet>)` never worked particularly well.  Specifically, the implementation of `Unbound(None)` was unclear:  should this parse `sysfs` and then bind to all the CPUs that were online (implying that the builder can fail if we have some problem parsing `sysfs`), or should we simply ignore `sysfs` altogether.  Ignoring `sysfs` has two benefits:  1) if we fail to parse `sysfs` for some reason, at least `Unbound` would still work and 2) if a user brings additional CPUs online after starting the executors, they would be able to run on the newly online CPUs.  However, documenting this behavior for `Unbound(None)` felt a little clumsy, in particular in explaining that `Unbound(Some(all_online_cpus))` is different from `Unbound(None)`.  This compares to having the same behavior for e.g. `MaxSpread(None)` and `MaxSpread(Some(all_online_cpus))`.

Perhaps we should have:

```rust
pub enum Placement {
    Unbound,
    Shared(Option<CpuSet>),
    MaxSpread(Option<CpuSet>),
    MaxPack(Option<CpuSet>),
}
```

I went with the current implementation because writing `MaxSpread` just felt a little nicer than `MaxSpread(None)` and I didn't mind `MaxSpreadOnCpus(cpu_set)` relative to `MaxSpread(Some(cpu_set))`.  I suppose it makes it slightly less clear that the underlying implementation is the same, i.e. `MaxSpread` and `MaxSpreadOnCpus(all_online_cpus)` are identical.  I didn't think `Shared` a.k.a. `Shared(None)` would get much use when we already have `Unbound`, so I didn't include it as companion to `SharedOnCpus`, but maybe should have just for completeness.

Long story short: I'm happy to go with anything that doesn't involve `Unbound(Option<CpuSet>)`.  Let me know what you prefer.

glommio/src/executor/mod.rs
trtsl: It does look a bit confusing at first -- I made this an `Option<Placement>` because of ownership rules.  Specifically, inside `on_all_shards` I call a new `spawn_thread` method, which I created to be able to write unit tests for the `Latch`.  The `spawn_thread` method is called in a loop with a `&self`, but I also need to take ownership of the `placement` before calling `spawn_thread` (if all this is done is one method as before, Rust is smart enough to figure out that I'm only taking the `placement` field once).  My options were:

* use an `Option<Placement>` and call `Option::take` (current approach)
* use `std::mem::replace(&mut self.placement, Placement::Unbounded)` but this felt a little misleading inside the `spawn_thread` method (`placement` is not used there, but who knows what might happen in the future)
* destructure `self` and pass all the required fields into `spawn_thread` as separate argument (feels ugly)
* clone `placement`; this would be ok today, but might not work in the future for e.g. `Placement::Custom(f)` where `f` is some closure supplied by the user

glommio/src/executor/placement/linux_sysfs.rs
trtsl: No, problem -- I'll have a look and move.

glommio/src/executor/mod.rs
glommer: First, for me the behavior that `Unbound(None)` should have is very clear: it means you really don't care, so you don't parse sysfs at all, and never even call any syscall. It is the same as today: it's whatever the OS wants.

I see semantic value in having that as a separate option, though.

I still don't see what you mean by "Shared". Why "Shared"  ? (naming things: the hardest problem in computer science!!)

but regardless of the name, I like what you proposed last with Unbound being separate, and I agree its `None` version makes little sense, so I'd refine this to:

```rust
pub enum Placement {
    Unbound, // don't do anything
    Shared(CpuSet), // restrict to the following cpuset. Because you're always specifying it, there's no point in an option
    MaxSpread(Option<CpuSet>), // None => MaxSpread on all avail CPU. Set: MaxSpread in the given cpuset
    MaxPack(Option<CpuSet>), // ditto
}
```

trtsl:
> We're getting there, but if it's not too much work I'd kindly ask you to merge the executor-in-a-directory change first so you don't risk more conflicts.

Will do.

> Also, as a heads up, take a look at #337
> 
> I plan to cut a release today, and seeing how this API is still a bit in-flight, I'm considering adding an experimental feature flag to it.
> 
> It shouldn't affect your work much, except you may have an extra source of conflicts in executor.rs

No problem, sounds good

glommio/src/executor/mod.rs
trtsl: That works for me.  There might be a better name than `Shared` -- maybe it'll come to us.  I went with `Shared` because the same `CpuSet` is shared amongst all the executors and each of the CPUs has a shared responsibility for all executors ... `AllForOneAndOneForAll(CpuSet)` / `UnusProOmnibusOmnesProUno(CpuSet)`?

glommer: I don't know, after the `ScopedTask`, @bryandmc became my go-to person for naming things lol.
Maybe he has an opinion here.

However I think the name should say something about the CPU placement. The fact that it is shared across executors is not that relevant.

What we are looking for is something that communicates the fact that it is restricted to a subset of all cpus, but free to move within that set.

Maybe `Restricted`, `Subset`, or something like that ?

glommer (approved):

trtsl:
I rebased everything to the latest master.  The only thing "new" here is that I'm adding a couple of types to the global namespace that were already public variants of `GlommioError` but not accessible via the docs: https://github.com/DataDog/glommio/pull/328/commits/2dc8d44951fe7edbc7be51b2e052054ad6b497bd (doing so in this pull request because one of those variants is created by this pull request).

glommer:
All good about the docs too.

Merged! Let's sync up about follow ups to this if you are interested. It's been a pleasure working with you so far!

PR #329, opened 2021-04-16
reactor io stats collection

HippoBaro:
### What does this PR do?

The reactor now records certain statistics from completed sources. We
currently records file basic counters:
* IO ops (reads/writes)
* bytes read/written;
* file open/close.

Applications may access these with `Local::io_stats()`. The plumbings
should make it easy for future additions.

This PR also records the same io stats we record in the reactor, but per
task queues, via `Local::task_queue_io_stats(handle)`. 
Since task queues are internally kept track of with a monotonic counter,
we don't delete stats when a task queue is removed. This means that an
application may retrieve io stats for queue even after they have
been removed from the executor.

### Checklist

- [x] My unit tests cover both failure and success scenarios
- [ ] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/src/executor.rs
+++ b/glommio/src/executor.rs
@@ -52,6 +52,7 @@ use crate::{
     task::{self, waker_fn::waker_fn},
     GlommioError,
     IoRequirements,
+    IoStats,
     Latency,
     Reactor,
     Shares,
@@ -1599,6 +1600,28 @@ impl<T> Task<T> {
     /// [`ExecutorStats`]: struct.ExecutorStats.html
     pub fn executor_stats() -> ExecutorStats {
         LOCAL_EX.with(|local_ex| local_ex.queues.borrow().stats)
+    }
+
+    /// Returns an [`IoStats`] struct with information about IO performed by
+    /// this executor's reactor
+    ///
+    /// # Examples:
+    ///
+    /// ```
+    /// use glommio::{Local, LocalExecutorBuilder};
+    ///
+    /// let ex = LocalExecutorBuilder::new()
+    ///     .spawn(|| async move {
+    ///         println!("Stats for executor: {:?}", Local::io_stats());
+    ///     })
+    ///     .unwrap();
+    ///
+    /// ex.join().unwrap();
+    /// ```
+    ///
+    /// [`IoStats`]: crate::IoStats
+    pub fn io_stats() -> IoStats {
+        LOCAL_EX.with(|local_ex| local_ex.get_reactor().io_stats())
     }
 
     /// Cancels the task and waits for it to stop running.
--- a/glommio/src/executor.rs
+++ b/glommio/src/executor.rs
@@ -1624,6 +1624,37 @@ impl<T> Task<T> {
         LOCAL_EX.with(|local_ex| local_ex.get_reactor().io_stats())
     }
 
+    /// Returns an [`IoStats`] struct with information about IO performed from
+    /// the provided TaskQueue by this executor's reactor
+    ///
+    /// # Examples:
+    ///
+    /// ```
+    /// use glommio::{Latency, Local, LocalExecutorBuilder, Shares};
+    ///
+    /// let ex = LocalExecutorBuilder::new()
+    ///     .spawn(|| async move {
+    ///         let new_tq = Local::create_task_queue(Shares::default(), Latency::NotImportant, "test");
+    ///         println!(
+    ///             "Stats for executor: {:?}",
+    ///             Local::task_queue_io_stats(new_tq)
+    ///         );
+    ///     })
+    ///     .unwrap();
+    ///
+    /// ex.join().unwrap();
+    /// ```
+    ///
+    /// [`IoStats`]: crate::IoStats
+    pub fn task_queue_io_stats(handle: TaskQueueHandle) -> Result<IoStats> {
+        LOCAL_EX.with(
+            |local_ex| match local_ex.get_reactor().task_queue_io_stats(&handle) {
+                Some(x) => Ok(x),
+                None => Err(GlommioError::queue_not_found(handle.index)),
+            },
+        )
+    }
+
     /// Cancels the task and waits for it to stop running.
     ///
     /// Returns the task's output if it was completed just before it got
--- a/glommio/src/parking.rs
+++ b/glommio/src/parking.rs
@@ -64,6 +64,7 @@ use crate::{
     IoStats,
     Latency,
     Local,
+    TaskQueueHandle,
 };
 
 /// Waits for a notification.
@@ -293,6 +294,10 @@ impl Reactor {
 
     pub(crate) fn io_stats(&self) -> IoStats {
         self.sys.io_stats()
+    }
+
+    pub(crate) fn task_queue_io_stats(&self, handle: &TaskQueueHandle) -> Option<IoStats> {
+        self.sys.task_queue_io_stats(handle)
     }
 
     #[inline(always)]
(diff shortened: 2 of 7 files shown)


glommio/src/parking.rs
```diff
@@ -229,6 +230,52 @@ impl SharedChannels {
...
+
+impl IoStats {
+    fn new() -> Self {
+        Default::default()
+    }
+
+    /// The total amount of files opened in this executor so far.
+    ///
+    /// [file_opened] - [file_closed] gives the current open files count
+    pub fn file_opened(&self) -> u64 {
```
glommer, line 253: why are the getter names singular ?

glommio/src/parking.rs
```diff
@@ -229,6 +230,52 @@ impl SharedChannels {
     }
 }
 
+/// Allows information about the current state of this reactor to be consumed
+/// by applications.
+#[derive(Default, Debug, Copy, Clone)]
+pub struct IoStats {
+    files_opened: u64,
+    files_closed: u64,
+    files_reads: u64,
```
glommer, line 239: would be nice to account Direct I/O vs Buffered I/O separately.
It seems to me this could be useful to discover performance bugs where you think you opened the file in direct I/O mode but ends up buffered because of some fallback.

glommio/src/parking.rs
```diff
@@ -692,14 +763,51 @@ impl Reactor {
...
+                }
+
+                self.add_waiter(cx.waker().clone());
+                Poll::Pending
+            })
+            .await,
+        )
+    }
+
+    fn collect_io_stats(&self, io_result: io::Result<usize>) -> io::Result<usize> {
```
glommer, line 779: I am not sure this is the right place to collect these.

For buffered file streams, for instance, we not always call `collect_rw`.  (see the `poll_xxx` functions for BufferedFile)

What bothers me the most, though, is that now we have a jump table at the end of a potentially very fast I/O operation. That's God's way of telling you the abstraction level is wrong! At the right abstraction layer we know which operation type we are accounting so we don't need to jump

Sockets will be a bit trickier because there is a `Source` path and a `yolo` path, but we're also interested in knowing how many times we're succeeding in the yolo path, so that can be accounted separately (plus sockets are not covered in this PR)

glommio/src/sys/uring.rs
```diff
@@ -1144,6 +1147,8 @@ impl Reactor {
             IoRequirements::default(),
             notifier.eventfd_fd(),
             SourceType::Read(PollableStatus::NonPollable(DirectIo::Disabled), None),
+            Weak::new(),
```
glommer, line 1150: This looks very odd to me. We do we need a reference to the reactor at this point ?

My best guess is that it's because the stats struct is in the parking reactor. But should it be?
If it lives inside the ring reactor, not only you don't need this weak, but you have access to a lot more information like which ring saw the dispatch, what kind of operation we're dispatching, etc.

glommer (changes requested):

glommer:
@HippoBaro please add code to your own repository fork.
When people clone this repository, they clone all existing branches, including the one you just created

glommio/src/parking.rs
HippoBaro: The new version now tracks buffered and unbuffered IO separately

glommio/src/parking.rs
HippoBaro: Thanks for the feedback! I've reworked the patch and changed my approach such that:
- We no longer have a jump table on the hot path (we know exactly what counter to increment)
- We record stats for each individual rings (we also provide a convenient way to get the stats for all rings)

HippoBaro: Note that this work doesn't tackle sockets at all but we can add them later. The way things are set up, it should be easy (even for the yolo path that issue syscalls directly).

glommio/src/sys/uring.rs
HippoBaro: See new changes, we no longer have weak references :+1:

HippoBaro:
> @HippoBaro please add code to your own repository fork.
> When people clone this repository, they clone all existing branches, including the one you just created

My bad, I'll work in my fork going forward

glommio/src/sys/mod.rs
```diff
@@ -558,3 +573,163 @@ pub(crate) fn shutdown(raw: RawFd, how: Shutdown) -> io::Result<()> {
         _ => Ok(()),
     }
 }
+
+/// Stores information about IO performed in a specific ring
+#[derive(Default, Debug, Copy, Clone)]
+pub struct RingIoStats {
```
glommer, line 579: `sys` is not a public module, so those structs that are maent to be public need to be reexported somewhere.

Note that in the docs, where you link this, you are referring to `crate::sys::IoStats`, which I believe is broken

glommio/src/executor.rs
```diff
@@ -1601,6 +1602,59 @@ impl<T> Task<T> {
...
+    /// let ex = LocalExecutorBuilder::new()
+    ///     .spawn(|| async move {
+    ///         println!("Stats for executor: {:?}", Local::io_stats());
+    ///     })
+    ///     .unwrap();
+    ///
+    /// ex.join().unwrap();
+    /// ```
+    ///
+    /// [`IoStats`]: crate::sys::IoStats
```
glommer, line 1622: this link is likely broken.

See my explanation on the sys module itself.

glommio/src/sys/mod.rs
```diff
@@ -487,6 +488,8 @@ impl Wakers {
     }
 }
 
+pub(crate) type StatsCollectionFn = fn(&io::Result<usize>, &mut RingIoStats) -> ();
```
glommer, line 491: this is a bit un-rusty, but should be okay for now as I am interested in merging this soon.

The reason is, `fn` cannot capture state, so it is rarely used. More often you will see `Fn` and `FnMut`, but since those are traits, they require a Box to be stored.

glommio/src/sys/mod.rs
```diff
@@ -558,3 +573,163 @@ pub(crate) fn shutdown(raw: RawFd, how: Shutdown) -> io::Result<()> {
...
+    }
+}
+
+/// Stores information about IO
+#[derive(Debug)]
+pub struct IoStats {
+    pub main_ring: RingIoStats,
+    pub latency_ring: RingIoStats,
+    pub poll_ring: RingIoStats,
+    pub all_rings: RingIoStats,
```
glommer, line 665: why do we need an extra field for all rings?

glommer (changes requested):
Only blocker is the fact that IoStats lives in `sys` and is not reexported.
This should be trivial to fix:
- In lib.rs, add a `pub use`. We already do that with other types, like `error::GlommioError`, just find the place where it is used.
- when you mention `IoStats` in the documents, please change to `crate::IoStats`.

glommio/src/sys/mod.rs
HippoBaro: Yes this is done on purpose. Using a `Fn` (or variant) would add an direction (via `dyn`) and an extra allocation. Because I don't need to capture state, this is actually very simple and efficient.

glommio/src/sys/mod.rs
HippoBaro: it's provided for convenience (and also so that we map-reduce only once). Do you want me to replace it with a method that compute that on-demand?

glommer: yes, that's better

glommer: and then maybe we don't need the other methods on `Iostats`

glommer (approved):

Issue #326, opened 2021-04-15
Sleep hangs when the sleep duration is very short


thirstycrow:
The following test works on my laptop, but hangs while running on our virtual box for gitlab ci.

```rust
#[cfg(test)]
mod tests {
    use std::time::Duration;

    use glommio::prelude::*;
    use glommio::timer::sleep;

    #[test]
    fn test() {
        LocalExecutor::default().run(async {
            println!("begin");
            sleep(Duration::from_secs(1)).await;
            println!("here");
            sleep(Duration::from_millis(1)).await;  // It hangs here
            println!("end");
        });
    }
}
```

glommer:
Thanks Jiang

can you reproduce locally with microsecond-level sleeps ?

glommer:
I tried a couple of things locally, including running the tests concurrently with a `yes` call burning CPU and could not reproduce.

Could this be a kernel issue, maybe? 

In any case, it would be helpful if you could add some debug statements so we can figure out what's wrong.

A couple of interesting places:
* do we even register this timer into the reactor ? (that's done in `timer_impl.rs` and eventually calls `insert_timer`)
* if it gets registered, does the timer gets awaken by process_timers(), or is it registered into the `io_uring` ?
* if it goes through io_uring, does the cqe ever complete ?

thirstycrow:
I cannot reproduce with the pasted code today, but it still happen on a test in our application. I added some print statements in parking.rs and uring.rs. According to the output, the executor is sleeping in the main ring and never wake up again.

```
running 1 test
process external events before wait
processing timers: {}, now: Instant { tv_sec: 178989, tv_nsec: 375240497 }
   pending count: 0
   ready count: 0
waiting for the next timer: None
wait returns: false
timer registered with id: 1
timer (id: 1) inserted
processing timers: {(Instant { tv_sec: 178989, tv_nsec: 376478530 }, 1): Waker { data: 0x14f810026970, vtable: 0x561ea8a8bf58 }}, now: Instant { tv_sec: 178989, tv_nsec: 375566825 }
   pending count: 1
   ready count: 0
process external events before wait
processing timers: {(Instant { tv_sec: 178989, tv_nsec: 376478530 }, 1): Waker { data: 0x14f810026970, vtable: 0x561ea8a8bf58 }}, now: Instant { tv_sec: 178989, tv_nsec: 375630124 }
   pending count: 1
   ready count: 0
waiting for the next timer: Some(848.406µs)
[...]

PR #330, opened 2021-04-16
don't treat timers separately

glommer:
This is ancient code, from a time in which we had to handle the preempt
timer in a special way. Our uring code is now more capable and more
generic, and the cancellation queue allow us to handle the preempt timer
like any other event.

I don't know for sure if this is the cause of Issue 326, but I suspect
it is: if a timer fires when we are processing the cancellations, and it
happens to be the only event that fires, it is theoretically possible
that we go to sleep, in which case we'll never wake up.

Although I am not sure this is the cause of the bug, it is one fewer
special case.

Ref #326

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -989,7 +989,6 @@ impl UringCommon for SleepableRing {
             self.ring.peek_for_cqe(),
             |source| match &mut *source.source_type.borrow_mut() {
                 SourceType::LinkRings => Some(()),
-                SourceType::Timeout(_) => Some(()),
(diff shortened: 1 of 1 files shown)


glommer:
@thirstycrow please try this one.

thirstycrow:
@glommer This one does not make any difference

glommer:
Thanks for verifying.

I am merging this anyway because it simplifies the code. Hopefully we'll find your issue soon

PR #335, opened 2021-04-23
scoped task: a way to pass a future that is not 'static

glommer:
Many of you have asked in the past for a way to spawn a task without
the 'static bound. Not only the 'static bound is hard to understand for
new users, it also forces us to resort to shared pointers in situations
where simple references would do.

As I have said before, the 'static bound comes from the fact that the
task may be detached. Even if it is not, the mere possibility of doing
so requires us to be strict abound that bound.

The solution for that is to create a new type of task: a ScopedTask.
The ScopedTask can never be detached, so we know it will live for as
long the variables it references. We can then just let the borrow
checker do its job.

### DIFF
--- a/glommio/src/executor.rs
+++ b/glommio/src/executor.rs
@@ -34,6 +34,7 @@ use std::{
     collections::{hash_map::Entry, BinaryHeap},
     future::Future,
     io,
+    marker::PhantomData,
     pin::Pin,
     rc::Rc,
     sync::Arc,
@@ -991,7 +992,7 @@ impl LocalExecutor {
         me.yielded = true;
     }
 
-    fn spawn<T>(&self, future: impl Future<Output = T>) -> Task<T> {
+    fn spawn<T>(&self, future: impl Future<Output = T>) -> multitask::Task<T> {
         let tq = self
             .queues
             .borrow()
@@ -1002,10 +1003,10 @@ impl LocalExecutor {
 
         let id = self.id;
         let ex = tq.borrow().ex.clone();
-        Task(ex.spawn(id, tq, future))
+        ex.spawn(id, tq, future)
     }
 
-    fn spawn_into<T, F>(&self, future: F, handle: TaskQueueHandle) -> Result<Task<T>>
+    fn spawn_into<T, F>(&self, future: F, handle: TaskQueueHandle) -> Result<multitask::Task<T>>
     where
         F: Future<Output = T>,
     {
@@ -1015,7 +1016,7 @@ impl LocalExecutor {
         let ex = tq.borrow().ex.clone();
         let id = self.id;
 
-        Ok(Task(ex.spawn(id, tq, future)))
+        Ok(ex.spawn(id, tq, future))
     }
 
     fn preempt_timer_duration(&self) -> Duration {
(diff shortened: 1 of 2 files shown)


glommer:
@bryandmc @thirstycrow @laa , you may like this one

duarten (approved):
Nice!

bryandmc (approved):
Looks good. Very useful and implementation is straight forward. I say ship it.

glommer:
Heads up:

Our contributor @trtsl just pointed out that this is Unsound, along with a reproducer that causes this to leak memory.
I will see if there's anything we can do to avoid that, but if not possible, this will unfortunately have to be either marked unsafe, or reverted.

duarten:
> Heads up:
> 
> Our contributor @trtsl just pointed out that this is Unsound, along with a reproducer that causes this to leak memory.
> I will see if there's anything we can do to avoid that, but if not possible, this will unfortunately have to be either marked unsafe, or reverted.

Ouch. Where is the reproducer?

glommer:
```rust

**Comment by bryandmc** [comment]
I feel like I should have remembered this considering it's the reason scoped threads was removed from the standard library .. mem::forget being safe always complicates things.. I wish you could just put a bound that says you can't forget something. May want to see how crossbeam solved this with their scoped threads. Because this is the exact same issue as the standard libraries scoped threads had. 

It's also possible that this is fundamentally unsafe and has to be marked as such. I don't see any issue with making it unsafe with a good doc comment explaining this issue because frankly I'd still use it.. it's easy to just not "forget" things.. 

I'll go investigate the crossbeam impl and see if there's any magic solution to this..

**Comment by glommer** [comment]
Hoping this at least makes for an interesting read: https://glaubercosta-11125.medium.com/async-rust-history-strikes-back-d69aced6760

**Comment by jon-chuang** [comment]
> I feel like I should have remembered this considering it's the reason scoped threads was removed from the standard library .. mem::forget being safe always complicates things.. I wish you could just put a bound that says you can't forget something. May want to see how crossbeam solved this with their scoped threads. Because this is the exact same issue as the standard libraries scoped threads had.
> 
> It's also possible that this is fundamentally unsafe and has to be marked as such. I don't see any issue with making it unsafe with a good doc comment explaining this issue because frankly I'd still use it.. it's easy to just not "forget" things..
> 
> I'll go investigate the crossbeam impl and see if there's any magic solution to this..

@bryandmc So the way crossbeam does this is: `transmute(a: 'a T) -> 'static T`. 

Is this safe? Not at all. The purpose of `forget` is to handle the case where you are asking an FFI to destruct an object rather than the rust code. So it looks as follows:
```
// in ffi.c
a = malloc(..);
// in ffi.rs
let a = unsafe { from_ptr(ffi_c::a()); }

// do something with a
scope(|| { f(&a));

// since the lifetime of the scope has ended, this is completely safe, and no memory leak or thread leak occurs
std::mem::forget(&a);
```

bryandmc:
The async_scoped you posted seems to have pretty similar limitations: https://docs.rs/async-scoped/0.6.1/async_scoped/struct.Scope.html#safety  (spoiler: you can't drop this either..) So we may not be able to use _exactly_ that method. But I am glad to hear you think it can be done, as we'd definitely love to have the feature.

jon-chuang:
> The async_scoped you posted seems to have pretty similar limitations: https://docs.rs/async-scoped/0.6.1/async_scoped/struct.Scope.html#safety (spoiler: you can't drop this either..) So we may not be able to use _exactly_ that method. But I am glad to hear you think it can be done, as we'd definitely love to have the feature.

The safe API we would want is probably `scope_and_block`: https://docs.rs/async-scoped/0.6.1/async_scoped/struct.Scope.html#method.scope_and_block
The docs mention some issues with deadlocks for high recursion, and needing a multi-threaded env for tokio. I'll see if I can find out why those limitations would occur.

PR #337, opened 2021-04-26
prepare for release 0.4.0

glommer:
releases a new version.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 futures = "0.3.5"
 futures-lite = "1.11.1"
 ansi_term = "0.12.1"
-glommio = { version = "0.3.0", path = "../glommio" }
+glommio = { version = "0.4.0", path = "../glommio" }
 pretty-bytes = "0.2.2"
 clap = "2.33"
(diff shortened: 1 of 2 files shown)


bryandmc:
Are there any outstanding, known bugs? And does that mean you'll cut the release from whatever SHA this produces? if so.. may want to get that joinHandle fix in first.

glommer:
I'd like to release something at a 2 month max cadence, if we have something bigish to show for it.
(I'll do a write up).

In this case, we have the `ScopedTask`, and the initial work on `LocalExecutorPoolBuilder` is good too (although it is very early).

You'll notice that I haven't merged this yet so yeah, might want to bring the join handle fix in.

I am not aware of any other bugs except maybe https://github.com/DataDog/glommio/issues/326

bryandmc (approved):
Yeah the sleep one is the only one I can find in the issue tracker. So sounds good to me. It's probably wise to have a regular-ish release cadence so might as well start that.

bryandmc (approved):
I say ship it

glommer:
Thanks for taking a look Bryan!
this is likely landing on crates.io tomorrow.

PR #339, opened 2021-04-28
fix join handle drop condition

glommer:
This is wrong: the comment says we should close if this is the last
reference *and* it is not closed. But we're testing with an or
statement.

It could obviously be that the comment is wrong, but:
1. logically it is not: why would we unconditionally close here if
   not closed?
2. while everything surprisingly works even with this condition, a
   couple of tests break if I change the FIFO order of execution, and
   this fixes it.

### What does this PR do?

### DIFF
--- a/glommio/src/task/join_handle.rs
+++ b/glommio/src/task/join_handle.rs
@@ -112,7 +112,7 @@ impl<R> Drop for JoinHandle<R> {
                 // If this is the last reference to the task and it's not closed, then
                 // close it and schedule one more time so that its future gets dropped by
                 // the executor.
-                let new = if (refs == 0) | (state & CLOSED == 0) {
+                let new = if (refs == 0) & (state & CLOSED == 0) {
                     SCHEDULED | CLOSED
                 } else {
                     state & !HANDLE


bryandmc:
Which tests/tasks were affected by their order?

glommer:
semaphore_overflow was one such example. I only debugged that one.

But conceptually anything that had a detached task, since the change in order meant the join handle was dropped first, and was at that point closed.

Full disclosure, what I am trying to do is to execute futures that are ready immediately. This is close in spirit to what Seastar did and is definitely an optimization, but this is plaguing my application a bit (right now I am working around it app-side)

Consider the following function:

```rust
async fn maybe_ready(&self, x: ObjId) -> SomeObject {
    match self.value_in_cache(x) {
       Some(x) => x,
       None => self.read_from_file(x),
    }
}
```

And then you need to read 1,000 of those objects, and you want to do so in parallel so you don't serialize on I/O:

```rust
let mut tasks = vec![];
for id in all_ids.iter() {
    tasks.push(Task::local(async move { 
       reader.maybe_ready(id);
    });
 });

stream::iter(tasks).then(|x| x).count().await;
```

This will generate 1,000 tasks that are now in the end of the queue. If everything is in cache, we could have resolved this in the current scheduler run. But now we need to wait for many other things (potentially a lot) to finish.

Unfortunately I don't see a way to avoid the allocation of a task because reusing wakers among tasks spells trouble.
But we could, as soon as we allocate the waker, execute the task. If it is ready, there is no need to queue.

The following test broke, though.
And once I fixed it, something else broke =)

So we'll see. I think it is a good change, but afraid of breaking too much stuff. I'll work on this a bit more.

bryandmc (approved):
Took me a little to trace through the code but I agree that this makes sense. The motivation sounds promising and definitely a valuable optimization for the future.

PR #341, opened 2021-04-29
bulk (vectorized) IO dispatch and coalescing

HippoBaro:
### What does this PR do?

This PR adds a new method `DmaFile::read_many` that takes an
iterator of iovec and dispatch them in bulk. The function returns a
stream-like struct that yields `ReadResult`s that users may
asynchronously consume.

The benefit of this approach is that it makes it possible to submit an
arbitrary number of concurrent IO and wait for them from a single task,
reducing the CPU overhead. This effectively means that a Task is no
longer needed as a unit of parallelization when doing IO in glommio.

Using the glommio benchmark utility, I measure a 19% IOPS
improvement for random 4k reads. However, I believe my local disk
is the limiting factor at this point (the result I get matches the
manufacturer IOPS rate for random read at queue saturation).

In addition to reducing the scheduler overhead, this PR also adds
a IO request coalescing/merging utility. The iovec stream is
preprocessed such that many `ReadResult` may feed from the same
buffer. For instance:

```rust
let iovecs: Vec<(u64, usize)> = vec![(64, 256), (32, 4064), (4095, 10), (8192, 4096)];
let results = new_file.read_many(iovecs.into_iter(), 4096, None);
assert_eq!(results.count().await, 4);
assert_eq!(Local::io_stats().all_rings().file_reads().0, 2);
```

The above code requests 4 buffers and gets them back from the stream,
yet the reactor only performs two reads internally.

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/io/bulk_io.rs
+++ b/glommio/src/io/bulk_io.rs
@@ -4,7 +4,14 @@ use crate::{
 };
 use core::task::{Context, Poll};
 use futures_lite::Stream;
-use std::{collections::VecDeque, os::unix::io::AsRawFd, pin::Pin, rc::Rc};
+use itertools::{Itertools, MultiPeek};
+use std::{
+    cmp::{max, min},
+    collections::VecDeque,
+    os::unix::io::AsRawFd,
+    pin::Pin,
+    rc::Rc,
+};
 
 #[derive(Debug)]
 pub(crate) struct OrderedBulkIo<U> {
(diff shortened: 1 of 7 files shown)


glommer:
Pretty good overall.
No major comments, I guess likely because we already discussed this a lot.

I am not as hot on the aligned version. Once you start coalescing, that means very little.
Let's see how the coalescing code ends up, and maybe we'll keep it, maybe we'll ditch it.

glommio/src/io/bulk_io.rs
```diff
@@ -49,3 +49,52 @@ impl<U: Copy + Unpin> Stream for OrderedBulkIo<U> {
...
+pub struct ReadManyResult {
+    pub(crate) inner: OrderedBulkIo<ReadManyArgs>,
+    pub(crate) current_result: ReadResult,
+}
+
+impl Stream for ReadManyResult {
+    type Item = super::Result<(u64, ReadResult)>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let next = match Pin::new(&mut self.inner).poll_next(cx) {
```
glommer, line 76: you can use the `ready!` macro here.

glommio/src/io/bulk_io.rs
```diff
@@ -49,3 +49,52 @@ impl<U: Copy + Unpin> Stream for OrderedBulkIo<U> {
...
+        };
+
+        match next {
+            None => Poll::Ready(None),
+            Some((source, args)) => {
+                if let Some(mut source) = source {
+                    let read_size =
+                        enhanced_try!(source.take_result().unwrap(), "Reading", self.inner.file)?;
+                    let mut buffer = source.extract_dma_buffer();
+                    buffer.trim_to_size(std::cmp::min(read_size, args.system_read.1));
```
glommer, line 88: why do we need to trim, if this buffer won't ever be visible to the user anyway ?

glommio/src/io/bulk_io.rs
```diff
@@ -98,3 +105,195 @@ impl Stream for ReadManyResult {
...
+        if taken > 0 {
+            self.next_merged = Some((taken, current.unwrap()));
+            self.next()
+        } else {
+            None
+        }
+    }
+}
+
+#[cfg(test)]
```
glommer, line 186: very good that this is tested separately.

glommio/src/io/bulk_io.rs
```diff
@@ -98,3 +105,195 @@ impl Stream for ReadManyResult {
...
+                ((0, 256), (0, 4096)),
+                ((32, 4064), (0, 4096)),
+                ((4095, 10), (4095, 10)),
+                ((8192, 4096), (8192, 4096))
+            ]
+        );
+    }
+
+    #[test]
+    fn large_input_passsthrough() {
```
glommer, line 210: passssss ??

glommio/src/io/bulk_io.rs
```diff
@@ -49,3 +49,52 @@ impl<U: Copy + Unpin> Stream for OrderedBulkIo<U> {
...
+            None => Poll::Ready(None),
+            Some((source, args)) => {
+                if let Some(mut source) = source {
+                    let read_size =
+                        enhanced_try!(source.take_result().unwrap(), "Reading", self.inner.file)?;
+                    let mut buffer = source.extract_dma_buffer();
+                    buffer.trim_to_size(std::cmp::min(read_size, args.system_read.1));
+                    self.current_result = ReadResult::from_whole_buffer(buffer);
+                }
+                Poll::Ready(Some(Ok((
```
glommer, line 91: I am having a really hard time understanding this:
every time you poll, if there is a Source, you process that `Source`. And then you return a single slice from it.

How can you return multiple slices from a single `Source` if you never test for `current_result` in your poll function?

glommio/src/io/bulk_io.rs
HippoBaro: lol :shrug:

glommio/src/io/bulk_io.rs
HippoBaro: If the inner stream yields `(_, Some(Source))` then I create a `ReadResult` from it and return that (trimmed according to the current iovec). If the inner stream yields `(_, None)`, then I reuse the last `ReadResult`. This way I let go of the source as soon as possible but I keep the underlying buffer for as long as necessary.

glommio/src/io/bulk_io.rs
HippoBaro: Correct; I was trimming twice, will remove

glommio/src/io/bulk_io.rs
glommer: ok. Since this is properly unit tested, I'm essentially ok with whatever is fast.
(And what determines what is fast is data, not my personal taste anyway)

glommer (approved):
I'm ready to merge this.

@HippoBaro let me know if you want to rework any commits or not.
You could potentially stash your commit n-1 into some of the previous ones. It'll be up to you.

HippoBaro:
> @HippoBaro let me know if you want to rework any commits or not.
You could potentially stash your commit n-1 into some of the previous ones. It'll be up to you.

I just rebased and the commits are clean now :+1:

PR #343, opened 2021-04-30
Deprecate scoped

glommer:
### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist


bryandmc:
Maybe we can mark them also as 'unsafe' where appropriate..? Or is there no safe way to use this api at all, even if the user controls for certain invariants..? Either way, making it unsafe makes it fail to compile until they put an unsafe block so that would really force the user to investigate, instead of just missing the deprecated warning in their build log..

glommer:
Yes, I am now leaning towards marking this as unsafe.

I want to find a safe way to do this (async makes things a bit harder than threads, but I only played with the crossbeam implementation a bit), but this is not it, unfortunately. There is no guarantee that the destructors will run, and if they don't all bets are off.

glommer:
closing in favor of marking this unsafe.

PR #350, opened 2021-05-11
simpler, faster reactor

HippoBaro:
### What does this PR do?

I've been doing some refactoring and performance improvements in the
reactor:
* `Source` wraps an `InnerSource` that contains fields wrapped in `RefCell`s
for interior mutability. This PR simplifies this by using a single `RefCell` for
the `InnerSource` altogether. This way all code inside the
`InnerSource` can use the compile-time borrow checker.
* Up to now, the source map existed as a static thread-local variable.
Thread-local entities are rather expensive in rust and so this commit
relocates the source map as a member of the `Reactor`. It is wrapped in
an `Rc` and shared in the different rings.
* We now avoid `Clone`ing sources when unnecessary in the reactor.

Some relevant IO benchmarks:

Before:
```
Buffered I/O: Wrote 1.07 GB in 352.827446ms, 3.04 GB/s
Buffered I/O: Closed in 55.866µs, Amortized total 3.04 GB/s
Direct I/O: Wrote 1.07 GB in 293.262222ms, 3.66 GB/s
Direct I/O: Closed in 121.008µs, Amortized total 3.66 GB/s
Direct I/O, write-behind: Wrote 1.07 GB in 222.827561ms, 4.82 GB/s
Direct I/O, write-behind: Closed in 1.030332ms, Amortized total 4.8 GB/s
Buffered I/O: Scanned 1.07 GB in 641.45869ms, 1.67 GB/s, 408670 IOPS
Direct I/O: Scanned 1.07 GB in 716.152568ms, 1.5 GB/s, 366046 IOPS
Direct I/O, read ahead: Scanned 1.07 GB in 270.486906ms, 3.97 GB/s, 969159 IOPS
Direct I/O, glommio API: Scanned 1.07 GB in 255.506772ms, 4.2 GB/s, 1025980 IOPS
Direct I/O, glommio API, large buffer: Scanned 1.07 GB in 110.936917ms, 9.68 GB/s, 18469 IOPS
```

After:
```
Buffered I/O: Wrote 1.07 GB in 244.095229ms, 4.4 GB/s
Buffered I/O: Closed in 58.921µs, Amortized total 4.4 GB/s
Direct I/O: Wrote 1.07 GB in 227.122739ms, 4.73 GB/s
Direct I/O: Closed in 110.318µs, Amortized total 4.73 GB/s
Direct I/O, write-behind: Wrote 1.07 GB in 171.369786ms, 6.27 GB/s
Direct I/O, write-behind: Closed in 782.615µs, Amortized total 6.24 GB/s
Buffered I/O: Scanned 1.07 GB in 623.257661ms, 1.72 GB/s, 420604 IOPS
Direct I/O: Scanned 1.07 GB in 704.036703ms, 1.53 GB/s, 372345 IOPS
Direct I/O, read ahead: Scanned 1.07 GB in 263.53527ms, 4.07 GB/s, 994724 IOPS
Direct I/O, glommio API: Scanned 1.07 GB in 249.565138ms, 4.3 GB/s, 1050407 IOPS
Direct I/O, glommio API, large buffer: Scanned 1.07 GB in 101.966746ms, 10.53 GB/s, 20094 IOPS
```

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -296,12 +296,13 @@ where
                 sqe.prep_read(op.fd, buf.as_bytes_mut(), pos);
 
                 let src = peek_source(from_user_data(op.user_data));
+                let mut srcb = src.borrow_mut();
                 if let SourceType::Read(PollableStatus::NonPollable(DirectIo::Disabled), slot) =
-                    &mut *src.source_type.borrow_mut()
+                    &mut srcb.source_type
                 {
                     *slot = Some(IoBuffer::Dma(buf));
                 } else {
-                    panic!("Expected Read source type");
+                    unreachable!("Expected Read source type");
                 };
             }
             UringOpDescriptor::Open(path, flags, mode) => {
@@ -351,8 +352,9 @@ where
             UringOpDescriptor::ReadFixed(pos, len) => {
                 let mut buf = buffer_allocation(len).expect("Buffer allocation failed");
                 let src = peek_source(from_user_data(op.user_data));
+                let mut srcb = src.borrow_mut();
 
-                match &mut *src.source_type.borrow_mut() {
+                match &mut srcb.source_type {
                     SourceType::Read(PollableStatus::NonPollable(DirectIo::Disabled), slot) => {
                         sqe.prep_read(op.fd, buf.as_bytes_mut(), pos);
                         *slot = Some(IoBuffer::Dma(buf));
@@ -403,7 +405,8 @@ where
                 );
 
                 let src = peek_source(from_user_data(op.user_data));
-                match &mut *src.source_type.borrow_mut() {
+                let mut srcb = src.borrow_mut();
+                match &mut srcb.source_type {
                     SourceType::SockRecv(slot) => {
                         *slot = Some(buf);
                     }
@@ -414,7 +417,8 @@ where
             UringOpDescriptor::SockRecvMsg(len, flags) => {
                 let mut buf = DmaBuffer::new(len).expect("failed to allocate buffer");
                 let src = peek_source(from_user_data(op.user_data));
-                match &mut *src.source_type.borrow_mut() {
+                let mut srcb = src.borrow_mut();
+                match &mut srcb.source_type {
                     SourceType::SockRecvMsg(slot, iov, hdr, msg_name) => {
                         iov.iov_base = buf.as_mut_ptr() as *mut libc::c_void;
                         iov.iov_len = len;
@@ -461,8 +465,8 @@ fn transmute_error(res: io::Result<u32>) -> io::Result<usize> {
 
 fn process_one_event<F, R>(cqe: Option<iou::CQE>, try_process: F, post_process: R) -> Option<bool>
 where
-    F: FnOnce(&InnerSource) -> Option<()>,
-    R: FnOnce(&InnerSource, io::Result<usize>) -> io::Result<usize>,
+    F: FnOnce(Ref<'_, InnerSource>) -> Option<()>,
+    R: FnOnce(Ref<'_, InnerSource>, io::Result<usize>) -> io::Result<usize>,
 {
     if let Some(value) = cqe {
         // No user data is POLL_REMOVE or CANCEL, we won't process.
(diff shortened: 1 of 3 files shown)


glommio/src/sys/uring.rs
```diff
@@ -297,11 +297,11 @@ where
 
                 let src = peek_source(from_user_data(op.user_data));
                 if let SourceType::Read(PollableStatus::NonPollable(DirectIo::Disabled), slot) =
-                    &mut *src.source_type.borrow_mut()
+                    &mut src.borrow_mut().source_type
```
glommer: I am surprised this works.
`borrow_mut()` should not survive its scope.
Maybe it works because the scope of this is the whole conditional.

But I guess a safer way would be to write `let mut src_mut = src.borrow_mut()` before the condition so that the reference safely is alive the whole time

glommio/src/sys/uring.rs
```diff
@@ -352,7 +352,7 @@ where
                 let mut buf = buffer_allocation(len).expect("Buffer allocation failed");
                 let src = peek_source(from_user_data(op.user_data));
 
-                match &mut *src.source_type.borrow_mut() {
+                match &mut src.borrow_mut().source_type {
```
glommer: Note that if this is broken and clippy doesn't complain because of the `unsafe` block, it would be very hard to catch. WE had issues before with things dying in the io_uring loop and because most of the time the requests complete right away, we don't notice. Until they don't...

glommio/src/sys/uring.rs
```diff
@@ -794,37 +796,39 @@ impl UringCommon for PollRing {
...
+    pub(crate) fn set_timeout(&mut self, d: Duration) -> Option<Duration> {
+        let t = &mut self.inner.borrow_mut().timeout;
         let old = *t;
         *t = Some(TimeSpec64::from(d));
         old.map(Duration::from)
     }
 
     fn timeout_ref(&self) -> Ref<'_, Option<TimeSpec64>> {
-        self.inner.timeout.borrow()
+        Ref::map(self.inner.borrow(), |x| &x.timeout)
```
glommer, line 843: neat

glommio/src/sys/uring.rs
```diff
@@ -269,8 +269,12 @@ lazy_static! {
     static ref IO_URING_RECENT_ENOUGH: bool = check_supported_operations(GLOMMIO_URING_OPS);
 }
 
-fn fill_sqe<F>(sqe: &mut iou::SQE<'_>, op: &UringDescriptor, buffer_allocation: F)
-where
+fn fill_sqe<F>(
+    sqe: &mut iou::SQE<'_>,
+    op: &UringDescriptor,
+    buffer_allocation: F,
+    source_map: &mut SourceMap,
```
glommer, line 276: yeah, good one

glommio/src/sys/uring.rs
HippoBaro: Can you explain why this isn't safe? The borrow is src and it lives long enough. The compiler allows the borrow to be in the `match` statement because it can guarantee none of the inner data escapes the scope.

I also tried to remove the unsafe bit and replaced it by many minimal ones  and the compiler doesn't take offense there either.

glommio/src/sys/uring.rs
HippoBaro: I believe this is because the lifetime of the member `source_type` is that of the `Ref`/`RefMut` itself. The compiler probably delays the `drop`ping of the borrow based on that. See how it works in a safe context:

```
    pub(crate) fn set_timeout(&mut self, d: Duration) -> Option<Duration> {
        let t = &mut self.inner.borrow_mut().timeout;
        let old = *t;
        *t = Some(TimeSpec64::from(d));
        old.map(Duration::from)
    }
```

glommio/src/sys/uring.rs
HippoBaro: Took me while to find the API!

glommer (approved):

glommio/src/sys/uring.rs
HippoBaro: I have rebased my branch and I no longer use this pattern. I now give the borrow its own, separate lifetime.

glommio/src/sys/uring.rs
HippoBaro: I have rebased my branch and I no longer use this pattern. I now give the borrow its own, separate lifetime.

PR #355, opened 2021-05-21
Remove `fasthash` from examples dependencies

fmckeogh:
### What does this PR do?

Removes an unused dependency from the examples manifest to allow running said examples on ARM64 (https://github.com/flier/rust-fasthash/issues/13).


glommer (approved):

glommer:
CI is not running because #reasons, so I am just overriding this.

Seems safe enough.
@fmckeogh thanks!

PR #357, opened 2021-05-24
docs.rs badge should point to glommio.

daschl:
Previously, it would point to mio which has a different
version


bryandmc (approved):
Looks good to me, nice catch

Issue #73, opened 2020-09-28
Merge read requests for random file access


glommer:
Although a fully fledged I/O Scheduler is probably way into the future, there is a functionality that I would like to see from an early implementation that can serve as a seed for that: random read request merging.

Imagine a read workload for a datastore. It is possible, if not likely, that many independent readers may be touching the same file, in the same - or close enough - position.

If that is the case, we don't have to go to storage and every time.
In preparation for such a change, in PR https://github.com/DataDog/scipio/pull/72 I am changing the return of `read_dma` and `read_dma_aligned` to be a `ReadResult` instead of a `DmaBuffer`. The `ReadResult` encapsulates a buffer, and two or more reads that end up reading from the same buffer may reuse them.

If reads are queued from positions similar to the one we are reading from (either adjacent or within a tolerance gap), we can extend the reads. For example: 

Let's say we fire read for position 0 for file A, reading 512 bytes.
Then another reader tries to read, 1024 bytes from position 1024.

We shouldn't ever delay firing the reads waiting for possible merges: but let's say when the time comes to fire the I/O, we see that those two requests are in the queue: it is better to read 2048 bytes and just give each one a slice of the buffer.

zserik:
As a first step, we should align read requests, and if then two requests are exactly the same, we should merge them into one, which returns the result to two requests with different wakers... This requires the functionality to distribute request responses.

laa:
Hi @glommer , what do you think is close enough in this case? The first assumption could be that close enough means one-page size. But still, file storage is slower than memory, I would say that distance of four pages (by default with the ability to change by the user) between requests could be close enough. WDYT?

glommer:
This will vary from device to device. Usually if you consider a 4kB page, if something is in the next page it might still be worth grabbing. So I'd say 8kB is a decent initial distance.

laa:
@glommer I suppose would be better to measure in pages, not in Kb. WDYT? In such terminology near enough is 2 pages. I suppose thinking in pages is better because 1-page distance always will be close enough.

glommer:
pages is an OS / memory concept. It doesn't necessarily translate to device properties.
There are devices for which larger transfers will be better, and there are devices which are okay with smaller transfers
(you can estimate that by dividing the peak bandwidth of the device by the peak iops)

laa:
[...]

PR #358, opened 2021-05-25
cross tasks io dedup

HippoBaro:
### What does this PR do?

**This is WIP.**

This PR adds a rudimentary IO scheduler on the read path such that tasks many consume IO performed by other tasks within an executor. 

Here is an example with 10 readers scanning the same file concurrently:
```
Direct I/O, glommio API, many readers:               Scanned 10.74 GB in 3.201660505s,  3.35 GB/s,  6399 IOPS
Direct I/O, glommio API, many readers, deduplicated: Scanned 10.74 GB in 517.044657ms, 20.77 GB/s, 39629 IOPS
```

Example code:
```
let first = read_some(new_file.clone(), 0..16384).await;
// we expect one IO to have been performed at this point
assert_eq!(Local::io_stats().all_rings().file_reads().0, 1);

// `first` holds a buffer and any further reads in this
// executor that falls inside 0..16384 will not trigger any IO
// but will instead feed from it, extending its lifetime.

let second = read_some(new_file.clone(), 67..578).await;
// should feed from the first buffer
assert_eq!(Local::io_stats().all_rings().file_reads().0, 1);
new_file.close_rc().await.expect("failed to close file");
```

This deduplication logic is guarded behind a file-level flag: `OpenOptions::immutable(bool)`. 

### Motivation

Imagine a read workload for a database. Depending on the exact read pattern, many independent readers may touch the same file. With this scheduler, they will benefit from each other and avoid performing duplicate IO, especially when read amplification is high (imagine a tree-like index structure).

### Related issues

#73 and #341 

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -25,7 +25,7 @@ type Result<T> = crate::Result<T, ()>;
 /// examples.
 #[derive(Debug)]
 pub struct BufferedFile {
-    file: GlommioFile,
+    pub(super) file: GlommioFile,
 }
 
 impl AsRawFd for BufferedFile {
@@ -151,12 +151,12 @@ impl BufferedFile {
     /// [`DmaFile`]: struct.DmaFile.html
     /// Reads from a specific position in the file and returns the buffer.
     pub async fn read_at(&self, pos: u64, size: usize) -> Result<ReadResult> {
-        let source =
-            self.file
-                .reactor
-                .upgrade()
-                .unwrap()
-                .read_buffered(self.as_raw_fd(), pos, size);
+        let source = self.file.reactor.upgrade().unwrap().read_buffered(
+            self.as_raw_fd(),
+            pos,
+            size,
+            self.file.scheduler.borrow().as_ref(),
+        );
         let read_size = source.collect_rw().await.map_err(|source| {
             GlommioError::create_enhanced(
                 source,
@@ -165,11 +165,7 @@ impl BufferedFile {
                 Some(self.as_raw_fd()),
             )
         })?;
-        Ok(ReadResult::from_sliced_buffer(
-            source.extract_buffer(),
-            0,
-            read_size,
-        ))
+        Ok(ReadResult::from_sliced_buffer(source, 0, read_size))
     }
 
     /// Issues `fdatasync` for the underlying file, instructing the OS to flush
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -166,7 +166,7 @@ impl BufferedFile {
             )
         })?;
         Ok(ReadResult::from_sliced_buffer(
-            source.extract_dma_buffer(),
+            source.extract_buffer(),
             0,
             read_size,
         ))
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -165,9 +165,11 @@ impl BufferedFile {
                 Some(self.as_raw_fd()),
             )
         })?;
-        let mut buffer = source.extract_dma_buffer();
-        buffer.trim_to_size(read_size);
-        Ok(ReadResult::from_whole_buffer(buffer))
+        Ok(ReadResult::from_sliced_buffer(
+            source.extract_dma_buffer(),
+            0,
+            read_size,
+        ))
     }
 
     /// Issues `fdatasync` for the underlying file, instructing the OS to flush
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -5,7 +5,7 @@
 //
 
 use crate::{
-    io::{glommio_file::GlommioFile, read_result::ReadResult},
+    io::{glommio_file::GlommioFile, read_result::ReadResult, OpenOptions},
     GlommioError,
 };
 use std::{
@@ -82,11 +82,12 @@ impl BufferedFile {
     ///
     /// [`create`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.create
     pub async fn create<P: AsRef<Path>>(path: P) -> Result<BufferedFile> {
-        let flags = libc::O_CLOEXEC | libc::O_CREAT | libc::O_TRUNC | libc::O_WRONLY;
-        GlommioFile::open_at(-1_i32, path.as_ref(), flags, 0o644)
+        OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .buffered_open(path.as_ref())
             .await
-            .map_err(|source| GlommioError::create_enhanced(source, "Creating", Some(path), None))
-            .map(|file| BufferedFile { file })
     }
 
     /// Similar to [`open`] in the standard library, but returns a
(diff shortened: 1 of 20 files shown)


glommio/src/io/buffered_file.rs
```diff
@@ -165,7 +165,7 @@ impl BufferedFile {
             )
         })?;
         Ok(ReadResult::from_sliced_buffer(
-            source.extract_dma_buffer(),
+            source.extract_buffer(),
```
glommer: the commit message doesn't make sense (commenting here because gh doesn't allow commenting on the commit message).

We already have an `IoBuffer` enum. So what exactly are you introducing? Will read on, but that reads strange

glommio/src/sys/mod.rs
```diff
@@ -11,6 +11,7 @@ use std::{
     ffi::CString,
     fmt,
     io,
+    io::Error,
```
glommer, line 14: again, commenting on the description of the source.
I don't know how big of a change this would be, but the part where you mention references, that would be good for any `Source`.

For consistency, we could add it everywhere. You have this fresh on your mind, how hard would that be ?

The real value of the shared source is that it is not disposed / cancelled when one of them goes away

glommio/src/sys/source.rs
```diff
@@ -225,3 +239,112 @@ impl Drop for Source {
         }
     }
 }
+
+pub struct SharedSource {
+    source: Source,
```
glommer, line 244: I am surprised by this.
The `Source` already has a list of results and wakers, so we are duplicating all of this here.

Keep in mind that those structs are allocated, so we'd like them to be as small as possible.
I thought you were going to create this from its building blocks.

glommio/src/sys/source.rs
```diff
@@ -225,3 +239,112 @@ impl Drop for Source {
...
+            }
+            x => panic!("Could not extract buffer. Source: {:?}", x),
+        }
+    }
+
+    pub(crate) fn add_waiter(&self, waker: Waker) {
+        // add the waker to the underlying source and insert the previous one (if any)
+        // into the backlog.
+        if let Some(previous) = self.source.add_waiter(waker) {
+            // before inserting the waker make we replace an old one if any
```
glommer, line 298: if any what?

I don't understand what are you trying to achieve with this code.

glommio/src/sys/source.rs
```diff
@@ -225,3 +239,112 @@ impl Drop for Source {
...
+        } else if let Some(result) = self.source.take_result() {
+            // the underlying source has a result, extract it
+            self.result.replace(Some(result.into()));
+            self.result.borrow().map(|x| x.clone().into())
+        } else {
+            // no result yet
+            None
+        };
+
+        if res.is_some() {
```
glommer, line 328: it's strange that this happens at the time we capture the result.

What is your rationale here ?

glommio/src/sys/source.rs
```diff
@@ -225,3 +239,112 @@ impl Drop for Source {
...
+            // be a good citizen and cooperate to wake the other tasks
+            self.wake_waiters();
+        }
+        res
+    }
+
+    pub(crate) fn has_result(&self) -> bool {
+        self.result.borrow().is_some() || self.source.has_result()
```
glommer, line 336: Yeah, this is confusing and an unfortunate side effect of reusing the entire `Source` as opposed to re-building it from building blocks. We should not need the second check

glommio/src/sys/source.rs
```diff
@@ -225,3 +239,112 @@ impl Drop for Source {
...
+    }
+
+    pub(crate) async fn collect_rw(&self) -> io::Result<usize> {
```
glommer, line 339: I am going to go on a limb here and take a guess that you added the entire `Source` to the `SharedSource` so you could have a consistent `collect_rw`. Am I warm?

This is better done through the trait system, I think

glommio/src/sys/source.rs
HippoBaro: `Source` can only have one waiter. Calling `add_waiter` replaces the current waiter and returns the previous one (or `None` if this is the first time it's called).

With `SharedSource` I want to create a queue of waiters. So when `add_waiter` is called, I first replace the waiter in the underlying `Source` and I add the previous one in a local vec so that I can't wake it up later on. This is what this code is doing.

glommio/src/sys/source.rs
HippoBaro: When a `Source` completes the reactor will wake the waker registered for it. For `SharedSource` I need this task to cooperate and wake the others. This is what this is doing.

I could explicitly call `wake_waiters` in consuming code but I felt it could remain an implementation detail hidden to consumers. This way, they can't get it wrong since this is the only way they can get the result out of the `SharedSource`.

glommio/src/sys/source.rs
HippoBaro: This is unfortunate indeed. At some point this was necessary-ish. I could rework this to directly poke into the underlying `Source`. That would be better since this essentially makes the `SharedSource` needlessly heavy.

glommio/src/sys/source.rs
HippoBaro: Yes, that's a good idea.

glommio/src/sys/source.rs
HippoBaro: > The Source already has a list of results and wakers

Not quite. `Source` keeps track of the result and **one** waker. We need a list of those. This is the primary reason that justifies the existence of `SharedSource`

glommio/src/sys/mod.rs
HippoBaro: Sorry I got confused (I wrote the commit message after the fact). I'll fix it up. Essentially what I meant to say is that we no longer return a `DmaBuffer` or a `Vec` to consumers. Instead we return the `IoBuffer` directly and it now implements `Deref` so 99% of the time consumers don't even need to `match`, they have easy access to `&[u8]`.

glommio/src/io/buffered_file.rs
HippoBaro: Sorry I got confused (I wrote the commit message after the fact). I'll fix it up. Essentially what I meant to say is that we no longer return a `DmaBuffer` or a `Vec` to consumers. Instead we return the IoBuffer directly and it now implements `Deref` so 99% of the time consumers don't even need to match, they have easy access to `&[u8]`.

glommio/src/sys/mod.rs
HippoBaro: The main difference between the two is that a `Source` has one consumer (and therefore, one waker) while `SharedSource` may have many. 

I think it would be pretty simple to unify them at the cost of maybe making `Source` a tad more expensive (even if there's only one waker I'd have to store it in a vec and that's an indirection vs what we currently have).

glommer: We don't want to make the `Source` more expensive. But couldn't we use traits to represent common behavior ?

HippoBaro:
@glommer The new version does away with `SharedSource`, the `Source` struct now fully supports having multiple consumers. It does so with minimal overhead using a `SmallVec<[Waker; 1]>`. Since there are no longer multiple variants of `Source` I did not feel it was necessary to introduce traits.

HippoBaro:
I have rebased on `master` and deduplication is now enabled exclusively for `ImmutableFiles`

glommio/src/io/buffered_file.rs
```diff
@@ -5,7 +5,7 @@
 //
 
 use crate::{
-    io::{glommio_file::GlommioFile, read_result::ReadResult},
+    io::{glommio_file::GlommioFile, read_result::ReadResult, OpenOptions},
```
glommer, line 8: if we're now using your work only through immutable files, why do we need this patch ?

HippoBaro: Strictly speaking, we don't need this anymore, yes. However I'm keen on keeping the change as it makes the API cleaner and more consistent. 

I'm okay with taking it out into a standalone PR if you think it would make this one cleaner (but since it will be rebased on master, I'm not sure what would be the point in doing so).

PR #359, opened 2021-05-28
Immutable file

glommer:
### What does this PR do?

Creates an ImmutableFile, that encodes users intention that a file will not change.

### Motivation

Hippo did.

### DIFF
--- a/glommio/src/io/immutable_file.rs
+++ b/glommio/src/io/immutable_file.rs
@@ -1,0 +1,394 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
+//
+use crate::io::{
+    dma_open_options::DmaOpenOptions,
+    DmaStreamReaderBuilder,
+    DmaStreamWriter,
+    DmaStreamWriterBuilder,
+    IoVec,
+    ReadManyResult,
+    ReadResult,
+};
+
+use std::path::Path;
+
+use futures_lite::{future::poll_fn, io::AsyncWrite};
+use std::{
+    io,
+    pin::Pin,
+    rc::Rc,
+    task::{Context, Poll},
+};
+type Result<T> = crate::Result<T, ()>;
+
+#[derive(Debug)]
+/// Builds a new [`ImmutableFile`], allowing linear and random access to a
+/// Direct I/O [`DmaFile`].
+///
+/// Working with an [`ImmutableFile`] happens in two steps:
+///
+/// * First, all the writes are done, through the [`ImmutableFilePreSealSink`],
+///   which has no read methods.
+/// * Then, after the sink is [`seal`]ed, it is no longer possible to write to
+///   it
+///
+/// [`DmaFile`]: struct.DmaFile.html
+/// [`ImmutableFilePreSealSink`]: ImmutableFilePreSealSink
+/// [`seal`]: ImmutableFilePreSealSink::seal
+pub struct ImmutableFileBuilder<P>
+where
+    P: AsRef<Path>,
+{
+    max_merged_buffer_size: usize,
+    max_read_amp: Option<usize>,
+    concurrency: usize,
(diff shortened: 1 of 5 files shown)


glommer:
seems like I fat fingered this and created this branch on the glommio branch directly and not on my own.
My apologies, I will delete this branch as soon as the PR is merged.

glommio/src/sys/mod.rs
```diff
@@ -130,6 +130,11 @@ pub(crate) fn accept_syscall(fd: RawFd) -> io::Result<RawFd> {
     syscall!(accept(fd, addr.as_mut_ptr() as *mut _, &mut length))
 }
 
+pub(crate) fn direct_io_ify(fd: RawFd, flags: libc::c_int) -> io::Result<()> {
+    syscall!(fcntl(fd, libc::F_SETFL, flags | libc::O_DIRECT))?;
+    Ok(())
+}
+
```
HippoBaro, line 137: Maybe we should expose the raw syscall instead? 

```
pub(crate) fn fcntl(fd: RawFd, flags: libc::c_int) -> io::Result<()> {
    syscall!(fcntl(fd, libc::F_SETFL, flags))?
}
```

glommio/src/io/immutable_file.rs
```diff
@@ -0,0 +1,395 @@
...
+    /// Builds an [`ImmutableFileSink`] with the properties defined by this
+    /// builder.
+    ///
+    /// The resulting sink can be written to until [`seal`] is called.
+    ///
+    /// [`ImmutableFile`]: ImmutableFileSink
+    /// [`new`]: ImmutableFileBuilder::new
+    /// [`seal`]: ImmutableFileSink::seal
+    pub async fn build_sink(self) -> Result<ImmutableFileSink> {
+        println!("siink builder");
```
HippoBaro, line 170: nit

HippoBaro: Or "niit", I should say :)

glommio/src/io/immutable_file.rs
```diff
@@ -0,0 +1,395 @@
...
+
+        Ok(ImmutableFile {
+            stream_builder,
+            max_merged_buffer_size: self.max_merged_buffer_size,
+            max_read_amp: self.max_read_amp,
+        })
+    }
+}
+
+impl AsyncWrite for ImmutableFileSink {
```
HippoBaro, line 235: Not a fan of the `ImmutableFileSink` construct and this line captures why: implementing `Write` on `Immutable` feels really wrong. Or, at least, the naming is unfortunate.

Could we instead build an immutable file from a regular `DmaFile`/`BufferedFile`? We could have a consuming function like so:

```
pub fn make_immutable(self) -> ImmutableFile {
  // convert and return
}
```

And we could also have a new `open_immutable` variant on the `OpeOptions` struct that would return an `ImmutableFile` directly.

I feel that this would simplify this work a bit as the write portion of your PR largely overlaps with the stream facilities we already have. Correct me if I'm wrong!

glommio/src/io/immutable_file.rs
glommer: yiikes! tnx

glommer: this is a big disadvantage of the tests capturing output. But overall still best this way.

Thanks for catching

glommio/src/io/immutable_file.rs
glommer: Not sure what your concern is. We are not implementing `Write` on the ImmutableFile. We are implementing it in the sink.

Suppose you now construct an ImmutableFile from a `DmaFile`. You now have to manually open the file, meaning the flags are whatever you passed it.

That means that if, by an accident, you happen to rewrite a file that was already in existence, the original `ImmutableFile` is changed and the contract of immutability is void.

That doesn't happen with the sink construct, because we force the file to be opened with `O_EXCL`: it will fail if it already exists.

So all in all, the sink is a necessary construct to make sure that the files are immutable as long as a certain path is always accessed through the Immutable Api.

Much harder to mess up.

Conceptually, it also matches the lifetime of phases of a file -> Create -> write, write, write -> seal -> read read read.

PR #362, opened 2021-05-31
use Error::from to slightly improve performance

qiujiangkun:
### What does this PR do?

This PR slightly improves performance (though not on hotpath)
`std::io::Error::from(ErrorKind)` is faster than `std::io::Error::new()`


glommer:
Thank you!

PR #370, opened 2021-06-09
Fix doctest, broken by 5dc9048bfffe3bac6d5997e3525df2e688008c6c

shikhar:
`cargo test` currently fails because it is treating these blocks as doctests


glommer:
thanks!

PR #371, opened 2021-06-11
support partial flushes from DmaStreamWriter

shikhar:
### What does this PR do?

**High-level**

Updated behavior for `DmaStreamWriter` `flush()` and `sync()` is to write and await the current buffer, even if it is not full. This means the file can contain some padding for the last write.

New APIs `flush_aligned()` and `sync_aligned()` are provided which maintain the previous behavior, when a full flush/sync is not required.

**Details**

In `flush_padded()`:
- No-op if there is nothing to flush, or a flush is already pending for the `current_pos`.
- The current buffer gets copied. The `aligned_pos` and `buffer_pos` are not updated. The `aligned_pos` in the file will subsequently get overwritten, if more data is written to the stream.

During `close()` or if close is not performed, in `drop()`, file truncation is done if `must_truncate()`.

Also added an internal `synced_pos` which is used to track the `flushed_pos` at the time the last sync was done, and syncing is elided based on that.

### Motivation

Make `DmaStreamWriter` more generally useful, since writes are not likely to be aligned. Getting confidence in data having made it to disk should not require a close and reopen. Currently, close is the only time the current buffer is written even if not full, and then the file is truncated.

### Related issues

https://github.com/DataDog/glommio/pull/375 laid some of the groundwork

### Checklist

- [x] My unit tests cover both failure and success scenarios
- [x] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/src/io/dma_file_stream.rs
+++ b/glommio/src/io/dma_file_stream.rs
@@ -854,10 +854,10 @@ impl DmaStreamWriterState {
         file: Rc<DmaFile>,
         do_close: bool,
     ) {
-        let final_pos = self.current_pos();
-        self.flush_partial(state.clone(), file.clone());
-        let mut pending = self.current_pending();
         self.file_status = FileStatus::Closing;
+        let final_pos = self.current_pos();
+        self.flush_padded(state.clone(), file.clone());
+        let mut pending = self.current_pending();
         Local::local(async move {
             defer! {
                 waker.wake();
@@ -936,7 +936,7 @@ impl DmaStreamWriterState {
 
     fn current_pending(&mut self) -> Vec<task::JoinHandle<()>> {
         let mut handles = Vec::with_capacity(self.pending_flush_count);
-        for (_pos, status) in &mut self.flushes {
+        for (_, status) in &mut self.flushes {
             if let FlushStatus::Pending(handle) = status {
                 if let Some(h) = handle.take() {
                     handles.push(h);
@@ -946,32 +946,36 @@ impl DmaStreamWriterState {
         handles
     }
 
-    fn flush_partial(&mut self, state: Rc<RefCell<Self>>, file: Rc<DmaFile>) -> bool {
+    fn flush_padded(&mut self, state: Rc<RefCell<Self>>, file: Rc<DmaFile>) -> bool {
         if self.buffer_pos == 0 {
             return false;
         }
-        let padding = self.buffer_size - self.buffer_pos;
-        assert!(padding > 0, "full buffer should have already been flushed");
-        let current_pos = self.current_pos();
-        if self.flushed_pos >= current_pos {
-            return false;
-        }
-        if self
+        assert!(
+            self.buffer_pos < self.buffer_size,
+            "full buffer should have already been flushed"
+        );
+
+        let flush_pos = self
             .flushes
             .back()
-            .filter(|(pos, _)| *pos >= current_pos)
-            .is_some()
-        {
+            .map_or(self.flushed_pos, |(pos, _)| *pos);
+
+        let current_pos = self.current_pos();
+
+        if flush_pos == current_pos {
             return false;
+        } else {
+            assert!(flush_pos < current_pos);
         }
 
         let buffer = self.current_buffer.take().unwrap();
 
-        // Restore a copy
-        let mut buffer_copy = file.alloc_dma_buffer(self.buffer_size);
-        buffer_copy.as_bytes_mut()[..self.buffer_pos]
-            .copy_from_slice(&buffer.as_bytes()[..self.buffer_pos]);
-        self.current_buffer = Some(buffer_copy);
+        if let FileStatus::Open = self.file_status {
+            let mut buffer_copy = file.alloc_dma_buffer(self.buffer_size);
+            buffer_copy.as_bytes_mut()[..self.buffer_pos]
+                .copy_from_slice(&buffer.as_bytes()[..self.buffer_pos]);
+            self.current_buffer = Some(buffer_copy);
+        }
 
         self.flush_one_buffer(buffer, state, file);
         true
@@ -1001,9 +1005,8 @@ impl DmaStreamWriterState {
 
 impl Drop for DmaStreamWriterState {
     fn drop(&mut self) {
-        assert_eq!(
-            0,
-            self.current_pending().len(),
+        assert!(
+            self.current_pending().is_empty(),
             "DmaStreamerWriter::drop() should have cancelled pending flushes"
         );
     }
@@ -1145,7 +1148,7 @@ impl DmaStreamWriter {
         let mut pending = {
             let mut state = self.state.borrow_mut();
             if partial && state.buffer_pos > 0 {
-                state.flush_partial(self.state.clone(), self.file.clone().unwrap());
+                state.flush_padded(self.state.clone(), self.file.clone().unwrap());
             }
             state.current_pending()
         };
@@ -1315,7 +1318,7 @@ impl AsyncWrite for DmaStreamWriter {
         if state.flushed_pos() == state.current_pos() {
             return Poll::Ready(Ok(()));
         }
-        state.flush_partial(self.state.clone(), self.file.clone().unwrap());
+        state.flush_padded(self.state.clone(), self.file.clone().unwrap());
         state.add_waker(cx.waker().clone());
         Poll::Pending
     }
--- a/glommio/src/io/dma_file_stream.rs
+++ b/glommio/src/io/dma_file_stream.rs
@@ -796,6 +796,7 @@ struct DmaStreamWriterState {
     file_status: FileStatus,
     error: Option<io::Error>,
     flushes: BTreeMap<u64, FlushStatus>,
+    pending_flush_count: usize,
     current_buffer: Option<DmaBuffer>,
     aligned_pos: u64,
     flushed_pos: u64,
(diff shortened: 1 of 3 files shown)


glommio/src/io/dma_file_stream.rs
```diff
@@ -789,18 +789,14 @@ struct DmaStreamWriterState {
...
-    // this is so we track the last flushed pos. Buffers may return
-    // out of order, so we can't report them as flushed_pos yet. Store for
-    // later.
-    out_of_order_write_returns: Vec<u64>,
+    aligned_pos: u64,
     flushed_pos: u64,
     buffer_pos: usize,
     write_behind: usize,
-    flush_on_close: bool,
+    sync_on_close: bool,
```
shikhar, line 799: already what it is called in public api, `with_sync_on_close_disabled()`

glommio/src/io/dma_file_stream.rs
```diff
@@ -698,7 +698,7 @@ impl AsyncRead for DmaStreamReader {
 pub struct DmaStreamWriterBuilder {
     buffer_size: usize,
     write_behind: usize,
-    flush_on_close: bool,
+    sync_on_close: bool,
```
shikhar, line 701: already what it is called in public api `with_sync_on_close_disabled()`

glommio/src/io/dma_file_stream.rs
```diff
@@ -1796,9 +1836,10 @@ mod test {
         assert_eq!(writer.current_pos(), 0);
         let buffer = [0u8; 5000];
         writer.write_all(&buffer).await.unwrap();
-        assert_eq!(writer.sync().await.unwrap(), 4096);
+        assert_eq!(writer.sync().await.unwrap(), 5000);
```
shikhar, line 1839: this seemed like the more intuitive behavior to me, even if there is padding

glommio/src/io/dma_file_stream.rs
```diff
@@ -789,18 +789,14 @@ struct DmaStreamWriterState {
     waker: Option<Waker>,
     file_status: FileStatus,
     error: Option<io::Error>,
-    pending: AHashMap<u64, task::JoinHandle<()>>,
+    // flush_pos -> (completed, handle)
+    pending: BTreeMap<u64, (bool, Option<task::JoinHandle<()>>)>,
```
shikhar, line 793: Just realized, the value would be better as an enum.

glommio/src/io/dma_file_stream.rs
```diff
@@ -1069,13 +1119,41 @@ impl DmaStreamWriter {
...
+            flushed_pos,
+            pos
+        );
+        Ok(flushed_pos)
+    }
+
+    /// TODO document
+    pub async fn sync_through(&self, pos: u64) -> Result<u64> {
+        let flushed_pos = self.flush_through(pos).await?;
+        self.file.clone().unwrap().fdatasync().await?;
```
shikhar, line 1151: `sync_file_range` may be worth considering?

glommer:
I am looking at this now

glommio/src/io/dma_file_stream.rs
glommer: yeah, those changes are good.

glommer:
I just did a cursory reading of this, without paying too much attention to the nitty gritty details. Some lines in particular weren't making a lot of sense, so I thought it was better to understand your whole approach first, then we can nosedive into the details.

A general comment first:
- The changes that change the flush name to sync internally are good. I would suggest you send a separate PR just with that first, and we'll merge that. The reason for that is that until we get this whole PR merged, you're inviting conflicts. The immutable file in particular is in-flux, so conflicts are likely. That's why I always suggest extracting the non-controversial things first.

Now to the actual work:
It seems to me that the core of what you are doing is to pad a buffer to its natural alignment, then flush that, and copy the extra bytes into a buffer. That's indeed the only way that has any chance of working, so that's good. Note that the fact that you are no longer appending, but writing to an existing position may confuse filesystems and lead to performance problems. So it is still best not to have to rely on this and flush on the end only. But at least it should work!

But:
- In that case I am less hot on the "upto" APIs. Do you have a particular use case of them, or were you just trying to conserve existing behavior? What would be best is to just be abundantly clear in the documentation that flush will overwrite buffers and may be inefficient if called outside the buffer boundary, and just be done with it. I am actually okay with something that waits for all existing in-flight buffers to finish, but "upto", where the users themselves have to keep track and specify a position, there's just too much room for error.
- The nastiest thing in this kind of flush that you are proposing is the following scenario:
   * you have 5000 bytes written. Then you flush. Because of padding you now wrote 4096 + 512 bytes, so the file is bigger now than the 5000 bytes. If there is a failure (or cancellation) at this point, we'd expect the resulting file to be 5000 bytes. So we need to truncate. But we want to be careful with that because truncate is expensive, and by the time this buffer finished flushing, the next one may have been written already.

I noticed that you do patch the code around truncate a bit, but I didn't understand if you are handling the case above. I would suggest, however, that you only do that on `Drop`: keep track of whether the file is in a state that needs truncating and truncate on `Drop`. I'd also like to see unit tests covering what happens if you drop after a `sync()`.

shikhar:
Thanks @glommer for the initial review!

I will create separate PR(s) with hopefully-uncontroversial changes (maybe one for the flush state tracking changes too, e.g. `VecDeque` instead of `HashMap`+`Vec`), that should make the diff in this PR smaller.

I implemented truncation on `Drop` in 1030c3329f4a491b881ebbc39322c528ec11a0ae. I couldn't get the call to async truncate to work unless I did a `block_on` it - I believe it was getting cancelled when the resulting future was dropped? So I went with explicitly calling the synchronous `sys::truncate`. Penalizing with a blocking truncation when close hygeine has not been maintained does not seem too bad to me :) Let me know if the approach seems ok or I should take another path.

Re: 'upto' APIs. I will write up why these seemed useful to me. Maybe in the process I will discover they are not.

shikhar:
Please see above 2 PRs that should take care of some of the incidental refactoring in here.

---

On `upto` APIs: imagine a WAL with discrete variable-length records. Could have 3 levels of safety before acknowledging a write:

1) Buffered - YOLO, this PR is useless, `write()` and ack.

2) Record handed off to disk (it's nice that with Direct IO we can know this without an fsync)

    Here awaiting a full `flush()` would imply waiting for more disk writes than necessary, than if we could just await `flush_upto(pos)` with the pos after the record was written. For a busy log to avoid partial writes, it could do a short async sleep first.

3) Record is fsynced to disk.

    Similar motiviation for `sync_upto()` which internally calls `flush_upto()` first. Using [`sync_file_range`](https://lwn.net/Articles/785690/) instead of `fdatasync` could be another enhancement. I added some further patches to also keep track of `synced_pos` and skip a sync if it's the same as `flushed_pos`.

glommio/src/io/dma_file_stream.rs
shikhar: Nevermind: https://man7.org/linux/man-pages/man2/sync_file_range.2.html

> This system call does not flush disk write caches and thus does not provide any data integrity on systems with volatile disk write caches.

glommer:
hi @shikhar 
`block_on` is not acceptable, and while a synchronous truncate is better than nothing, it will be a cause of stalls.
(a secret: unfortunately because the support for truncate in io_uring is still nascent, currently all of our truncates are synchronous. But that will change soon, and I'd like to keep everything working with that in mind)

The way we do this for close is to have a special operation on the ring, called `async_close` that lives on uring.rs.
Indeed, operations are usually tied to a `Source`, and if the `Source` is dropped, the operation is cancelled. So what we need here is an operation that is detached for its `Source`.

Please do the same for truncate: add an `async_truncate` into the ring. You will have to end up calling `sys::truncate` anyway (add a comment), but at least when we implement truncate through io_uring everything will work automatically.

About the API, what I don't like about the `upto` API is that the user has to specify a position. They shouldn't have to do it.
It's best if you flush all buffers that are in-flight and return the safe position. This should reflect in the name of the API. So maybe `sync_aligned()` ?

glommio/src/io/dma_file_stream.rs
```diff
@@ -1103,28 +1208,15 @@ impl DmaStreamWriter {
...
-    ///     // away. It will not be reflected on current_flushed_pos(), but a
-    ///     // sync() will wait on it.
+    ///     // away.
     ///     assert_eq!(writer.current_flushed_pos(), 0);
-    ///     assert_eq!(writer.sync().await.unwrap(), 4096);
+    ///     assert_eq!(writer.sync().await.unwrap(), 5000);
     ///     writer.close().await.unwrap();
     /// });
     /// ```
     pub async fn sync(&self) -> Result<u64> {
```
shikhar, line 1217: `flush()` is implemented via `AsyncWriteExt` which invokes `poll_flush()`

shikhar:
Thanks @glommer, that's great feedback.
- Updated to call truncate via reactor from drop.
- I agree that providing `flush_aligned()` and `sync_aligned()` (the current behavior) is the best way to keep a clean API that allows users to choose a performant alternative where feasible, updated.

I think the diff in this PR will be neater if we are able to merge https://github.com/DataDog/glommio/pull/375 first, so I look forward to your comments there. Then I can pick back up here.

shikhar:
merged in latest master now that #375 is in, which makes the diff here neater :tada: 

WDYT @glommer? TODOs on my end are PR description + docs for new pub fns `flush_aligned()` / `sync_aligned()`.

glommer:
Sorry it took me so long to get to this, @shikhar . Looking at this now

glommer:
Thanks so much, @shikhar . I'm ready to merge this.

There are a couple of functions marked as public that have their documentations as `TODO`. CI is unfortunately not running for your contributions for some reason, but I am positive this would break us.

Please document those functions, and make sure that `cargo +nightly fmt --all`, `cargo clippy`, `cargo doc --all-features` and `cargo clippy --all-features` are passing

shikhar:
Thanks for reviewing @glommer! I will be able to get back to this PR later in the week.

Not sure why CI doesn't run for my PRs specifically :-| But I will certainly run those commands locally.

glommer (approved):

glommer:
I am merging this and hoping for the best. CI seems to be working once I merge things to master, so if there are any problems we'll find out and can act on it.

PR #372, opened 2021-06-14
Marginally faster SPSC queue

duarten:
### What does this PR do?

The PR applies the following changes to the SPSC queue:

- Align the sender and receiver fields to 128 bits, to account for prefetching.

- Synchronize enqueue/dequeueing via the data slot, instead of through the indexes, to decrease the amount of potential cache misses from 2 (index of other thread, plus slot) to just 1 (the slot). As a consequence, the `size()` operation has more relaxed semantics, so we clamp the resulting value.

- Optimize the sender to look N positions ahead, to avoid an `Ordering::Acquire` reads.

The queue was benchmarked with:

```rust
 fn bench() {
        const REPETITIONS: u128 = 10 * 1000 * 1000;
        let (sender, receiver) = make::<u64>(1024);
        let consumer = std::thread::spawn(move || {
            let t = Instant::now();
            for i in 0..REPETITIONS {
                let mut opt: Option<u64>;
                while {
                    opt = receiver.try_pop();
                    opt.is_none()
                } {}
            }
            println!(
                "cost of receiving shared channel {:#?}ns",
                t.elapsed().as_nanos() / REPETITIONS,
            );
        });
        let t = Instant::now();
        for i in 0..REPETITIONS {
            while sender.try_push(i as u64).is_some() {}
        }
        println!(
            "cost of sending shared channel {:#?}ns",
            t.elapsed().as_nanos() / REPETITIONS,
        );
        consumer.join().unwrap();
    }
```

The queue in this PR clocks on average at 3ns per operation, while the one in Glommio clocks at 5ns per operation. However, no difference is observed with the `shared_channel` bench. The benchmarks were run on a 2,3 GHz i7 Mac.

### Motivation

I had to write an SPSC queue and this is how I typically approach them.

### Checklist

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -89,3 +89,8 @@ required-features = ["bench"]
 [[bench]]
 name = "sharding"
 harness = false
+
+
+[[bench]]
+name = "spsc_queue"
+harness = false
--- a/glommio/benches/spsc_queue.rs
+++ b/glommio/benches/spsc_queue.rs
@@ -1,0 +1,36 @@
+use glommio::channels::spsc_queue;
+use std::time::Instant;
+
+fn test_spsc(capacity: usize) {
+    const RUNS: u32 = 10 * 1000 * 1000;
+    let (sender, receiver) = spsc_queue::make::<u64>(1024);
+    let consumer = std::thread::spawn(move || {
+        let t = Instant::now();
+        for _ in 0..RUNS {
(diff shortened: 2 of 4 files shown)


glommer:
it would be useful to merge this benchmark you used into the benches directory.

duarten:
> it would be useful to merge this benchmark you used into the benches directory.

Will do.

glommer (approved):

PR #373, opened 2021-06-14
remove Unpin restriction from IoVec trait

glommer:
An IoVec should just be a base+pos combination. Adding Unpin to it
makes it harder to consume.

The Unpin restriction should be attached to IoVec uses whenever needed,
instead of always bundled in the trait.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist


HippoBaro (approved):
Thanks!

PR #375, opened 2021-06-16
DmaStreamWriter flush state management improvements

shikhar:
### What does this PR do?

Public APIs are not affected.

* Use a ring buffer for `pending` flushes as well as `out_of_order_write_returns`.
* Instead of `flush_id`, the last position represented in the write (`flush_pos`, or what `state.flushed_pos` should be updated to on completion) is used to identify each write.
* Rely on the natural sorting as it is an append-only stream, no need to resort the out-of-order write returns.
* When there is a contiguous prefix of completed flushes, `state.flushed_pos` is updated.

Although intended as a simplification, the LOC unfortunately turns out to be a bit more.

### Motivation

Makes partial flushes (https://github.com/DataDog/glommio/pull/371/) easier. Current logic for updating `state.flushed_pos` is fairly tied to aligned positions.

### Related issues

Enabler for https://github.com/DataDog/glommio/pull/371

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/io/dma_file_stream.rs
+++ b/glommio/src/io/dma_file_stream.rs
@@ -822,12 +822,22 @@ impl DmaStreamWriterFlushState {
 
     fn on_start(&mut self, flush_pos: u64, handle: task::JoinHandle<()>) {
         self.pending_flush_count += 1;
+        if let Some((prev_flush_pos, _)) = self.flushes.back() {
+            assert!(*prev_flush_pos < flush_pos);
+        }
         self.flushes
             .push_back((flush_pos, FlushStatus::Pending(Some(handle))));
     }
 
     fn on_complete(&mut self, flush_pos: u64) -> u64 {
         self.pending_flush_count -= 1;
+        // note:
+        // - writes can complete out-of-order
+        // - `flushes` is maintained in sorted order of position
+        // we will update the status of `flush_pos` to complete.
+        // we also count contiguous completed flushes from the front,
+        // `flushed_pos` can be updated accordingly and we don't
+        // need to track them anymore, so drain that range.
         let mut drainable_len = 0usize;
         let mut found_pending = false;
         for (pos, status) in &mut self.flushes {
--- a/glommio/src/io/dma_file_stream.rs
+++ b/glommio/src/io/dma_file_stream.rs
@@ -793,16 +793,83 @@ enum FlushStatus {
 }
 
 #[derive(Debug)]
+struct DmaStreamWriterFlushState {
+    flushes: VecDeque<(u64, FlushStatus)>,
+    pending_flush_count: usize,
+    flushed_pos: u64,
+}
+
+impl DmaStreamWriterFlushState {
+    fn new(write_behind: usize) -> DmaStreamWriterFlushState {
+        DmaStreamWriterFlushState {
+            flushes: VecDeque::with_capacity(write_behind * 2),
+            pending_flush_count: 0,
+            flushed_pos: 0,
+        }
+    }
+
+    fn take_pending_handles(&mut self) -> Vec<task::JoinHandle<()>> {
+        let mut handles = Vec::with_capacity(self.pending_flush_count);
+        for (_, status) in &mut self.flushes {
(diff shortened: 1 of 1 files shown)


glommer:
hello @shikhar 

unfortunately with this patch, tests fail. The test seems unrelated, but it starts to fail nevertheless.
Note that because CI doesn't have a new enough kernel you should be running `cargo test` yourself for the time being (yes, that sucks ;( )

About the patch itself: I don't mind too much the increased LOC count if the logic is easier to follow.

glommio/src/io/dma_file_stream.rs
```diff
@@ -988,14 +1007,13 @@ impl DmaStreamWriter {
             sync_on_close: builder.sync_on_close,
             current_buffer: None,
             waker: None,
-            pending: AHashMap::new(),
+            flushes: VecDeque::with_capacity(builder.write_behind * 2),
```
glommer, line 1010: why would we need the queue to be 2 * write_behind ?
In which situations do we expect to have more flushes than the number of write behind ?

glommio/src/io/dma_file_stream.rs
```diff
@@ -904,56 +907,72 @@ impl DmaStreamWriterState {
...
-            if *x == self.flushed_pos {
-                self.flushed_pos += self.buffer_size as u64;
-                self.out_of_order_write_returns.pop();
-            } else {
-                return;
+    fn on_flush_success(&mut self, flush_pos: u64) {
+        let mut complete_prefix_len = 0usize;
+        let mut complete_prefix_finalized = false;
+        let mut found_index = None;
+        for (i, (pos, status)) in self.flushes.iter_mut().enumerate() {
```
glommer, line 914: how is this easier to follow than a map?

glommio/src/io/dma_file_stream.rs
shikhar: Out-of-order completions will still be present. In the worst case you have the oldest flush pending, then `wb - 1` completions, allowing for `wb - 1` flushes to further be added to the queue. To keep it simple went with `wb * 2` for capacity.

glommio/src/io/dma_file_stream.rs
shikhar: Fair point! It's not just the `pending` map though. There is also the `out_of_order_writes` vec. The `flushes` ringbuffer combines the two. I found it easier to reason about the prefix of contiguous completions in this ring buffer representing a safe update to `state.flushed_pos`. Now there is a single place assignment to `state.flushed_pos` happens.

shikhar:
I was able to reproduce the test failure (does not always happen), I had assumed it was an unrelated flake. I'll dig more

```
---- channels::shared_channel::test::shared_drop_cascade_drop_executor stdout ----
thread 'channels::shared_channel::test::shared_drop_cascade_drop_executor' panicked at 'assertion failed: `(left == right)`
  left: `50`,
 right: `0`', glommio/src/channels/shared_channel.rs:913:9


failures:
    channels::shared_channel::test::shared_drop_cascade_drop_executor
```

glommio/src/io/dma_file_stream.rs
shikhar: Also not sure how I can make non-aligned positions work with the current approach taken in `adjust_flushed_pos()`.

glommio/src/io/dma_file_stream.rs
glommer: that makes sense.

glommio/src/io/dma_file_stream.rs
```diff
@@ -890,8 +895,6 @@ impl DmaStreamWriterState {
                 let res = file.close_rc().await;
                 collect_error!(state, res);
             }
-
-            state.flushed_pos = final_pos;
```
shikhar, line 894: `on_flush_success()` takes care of this

shikhar:
I can reproduce that failure if I run that test `shared_drop_cascade_drop_executor` in a loop on the master branch.

glommer (approved):

Issue #179, opened 2020-11-19
provide a specialized logging macro set


glommer:
Rust has the very helpful macros from `log` : `info!`, `warn!`, etc.

It would be helpful to have our own, that would include the executor name or ID on it.
That's because often times executors are all performing the same tasks, only in different parts of the dataset. It is helpful to know where exactly the issues are coming from.

The challenge in this issue is not to implement the macros, but to do it in an ergonomic way that existing applications can adapt without too much trouble.

laa:
Hi @glommer . Are you aware about tracing crate https://lib.rs/crates/tracing . Probably would be good to use it and provided concepts alongside with macro wrappers with useful information about executors and so on.

laa:
I would be curious to take this issue and issue #179. But before I need to be more acknowledged with code and probably will provide more PRs during the investigations. Also, I am very curious about your opinion of using a tracing crate. IMHO structural approach which they use is useful for glommio too.

glommer:
Although obviously familiar with tracing systems in general, I was not familiar with the tracing crate.
I would have to spend more time on it. In particular from an admittedly very shallow first reading on their website their examples focus too much on how to generate the traces, but I would like to have a better idea of the output too and also the cost of generating and keeping the traces until we are done with them.

Traces are definitely useful at some point in the application (like request paths), but usually not so much for events.

But I wouldn't mind seeing a proof of concept of an event system based on `tracing` if you think it can fit the requirements

Ralith:
> In particular from an admittedly very shallow first reading on their website their examples focus too much on how to generate the traces, but I would like to have a better idea of the output too and also the cost of generating and keeping the traces until we are done with them.

FWIW, tracing has pluggable backends, and is intended to be a fairly lightweight facade to promote interop and support diverse applications. How data is recorded and emitted is up to you, though a number of options like the [fmt](https://docs.rs/tracing-subscriber/0.2.15/tracing_subscriber/fmt/index.html) subscriber are available off the shelf.

bryandmc:
[...]

PR #381, opened 2021-06-29
Begins fixing #179: Add 'tracing' dependency + example tracepoints

bryandmc:
Added the very popular 'tracing' library, as well as a basic subscriber for use
in tests, benchmarks, etc. Only includes a few tracepoints as an example and
doesn't begin to pretend it's instrumented the entire system (FAR from it). I
think with the dependency included people will be more inclined to help out
adding tracepoints. Also fixed compiler errors due to inability to infer types,
potentially caused by the addition of this new crate and it's traits, or were
part of a new compiler behavior. All it does is make them more verbose and
precise about the type so it's included in this commit as well. Only tests are
affected by this.

### What does this PR do?
The need for tracing tools, and a decision to utilize the most popular framework in this space.

### Motivation
Hopefully to 'get the ball rolling' and encourage people to use it to create new `tracepoints` all over the codebase.

### Related issues
It's the beginning to #179 -- maybe we will want to customize this later as needed.

### Additional Notes
This is just to get things rolling. Ideally I'd prefer to just use usdt tracepoints but `tracing` is currently the most popular tool for this kind of thing in the rust async community. Another note is I was unable to compile without changing some types to include more information about their type, and aren't related to the actual change but were necessary to change unless I wanted to break the build or submit a separate commit (seemed excessive..). 

### Checklist

[?] I have added unit tests to the code I am submitting (I modified some to launch tracing to test)
[-] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -35,6 +35,7 @@ mod placement;
 
 use latch::{Latch, LatchState};
 pub use placement::{CpuSet, Placement};
+use tracing::trace;
 
 use std::{
     cell::RefCell,
@@ -213,7 +214,7 @@ macro_rules! to_io_error {
 fn bind_to_cpu_set(cpus: impl IntoIterator<Item = usize>) -> Result<()> {
     let mut cpuset = nix::sched::CpuSet::new();
     for cpu in cpus {
-        tracing::trace!("Attempting to bind to cpu: {}", cpu);
+        trace!("Attempting to bind to cpu: {}", cpu);
         to_io_error!(&cpuset.set(cpu))?;
     }
     let pid = nix::unistd::Pid::from_raw(0);
@@ -939,14 +940,6 @@ impl LocalExecutor {
         };
 
         let io_requirements = IoRequirements::new(latency, index);
-        tracing::trace!(
-            shares = ?shares.clone(),
-            ?latency,
-            index,
-            ?io_requirements,
-            "Creating task queue"
-        );
-
         let tq = TaskQueue::new(TaskQueueHandle { index }, name, shares, io_requirements);
 
         self.queues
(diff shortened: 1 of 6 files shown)


glommio/src/executor/mod.rs
```diff
@@ -2021,7 +2062,13 @@ mod test {
     }
 
     #[test]
+    #[allow(unused_must_use)]
     fn bind_to_cpu_set_range() {
+        tracing_subscriber::fmt::fmt()
```
glommer, line 2067: shouldn't init be something for the application to do ?

This is fine in some of our tests, I guess, but this is a particularly interesting place for it. Why here?

glommio/src/executor/mod.rs
```diff
@@ -1130,6 +1139,22 @@ impl LocalExecutor {
...
+            .map(|x| {
+                (
+                    *x.0,
+                    x.1.borrow().name.clone(),
+                    x.1.borrow().shares.clone(),
+                    x.1.borrow().io_requirements,
+                )
+            })
+            .collect::<Vec<_>>();
+        tracing::trace!(id = self.id, ?queues, "LocalExecutor running");
```
glommer, line 1156: Why do you feel the queues are important here ?
It seems to me that when the executor starts running this will almost always have the same value, and a predictable one.

glommio/src/executor/mod.rs
```diff
@@ -938,6 +939,14 @@ impl LocalExecutor {
         };
 
         let io_requirements = IoRequirements::new(latency, index);
+        tracing::trace!(
```
glommer, line 942: wondering if it wouldn't be better to just add `trace` instead of `tracing::trace`? and `use trace` on top? Should make it easier for us to compile-time replace it with a usdt macro in the future.

glommer (changes requested):

glommio/src/executor/mod.rs
bryandmc: yeah, the usdt stuff is a little trickier so we might have to wrap them both in another macro to be able to do that.. Again, I am fine just removing the "example" tracepoints. They aren't the important part at all. I think I can address all these comments pretty easily.

glommio/src/executor/mod.rs
bryandmc: Yeah this is just testing that, but it's ok I moved it in my next patch I just uploaded.

glommio/src/executor/mod.rs
bryandmc: I don't necessarily.. I admit I hastily added the tracepoints so I've gone ahead and removed most of them in my most recent patch. I really just wanted the dependency to land so people start adding stuff themselves as needed.

glommer: Agree with the intention, but let's make sure the ones we do add make sense.

In this case it's fair to trace that the executor was created. But the queues are not really useful

bryandmc: K, so everything as is but don't log the queues there? I'll go ahead and make that patch.. Again my major goal is to add the infrastructure. Not to mention it allows adding (temporary) trace lines instead of printlns when debugging. Which is nice.

glommer: that's right.
That we created an executor is good tracing info (adds information), which queues it has at startup is not

glommer:
Thanks for the heads up Bryan.

I actually thought I had pressed the button for this one already...

PR #398, opened 2021-08-13
Execute some reminescent syscalls in a thread

glommer:
While io_uring has gained support for some of them already, that support
only exists in much newer kernels. On 5.8 they are absent and given how
pervasive 5.8 is, it is good to have a fallback.

Implementing the io_uring opcodes is welcome and should be easy. The
whole fallback mechanism is implemented on top of the `Source`
abstraction, so we just have to add a runtime switch as a result of
a probe for those opcodes.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist


duarten (approved):
nice!

PR #399, opened 2021-08-27
make sure cpu placement works in kernels with no numa support

glommer:
That seems to be the case for docker-on-mac.


davidblewett (approved):
+1

PR #400, opened 2021-08-28
Fix RwLock doesn't work with multiple writers

waynexia:
### What does this PR do?

Fix the `RwLock` that will panic when multiple writers want to get the write guard. 

It only checks the number of readers and will grant more than one write guard, which leads to `RefCell` panic due to two mutable borrows.

### Motivation
N/A

### Related issues
N/A

### Additional Notes

### Checklist


glommer (approved):

glommer:
thanks!

PR #403, opened 2021-08-31
Implement Clone for Placement.

davidblewett:
### What does this PR do?

This PR implements the Clone trait for `glommio::Placement`. 

### Motivation

The impetus for this was passing a `fut_gen` to `glommio::LocalExecutorPoolBuilder::method.on_all_shards` that included a `Placement`. Since `on_all_shards` requires that the `fut_gen` be clonable, I had to go through hoops.

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/executor/placement/mod.rs
+++ b/glommio/src/executor/placement/mod.rs
@@ -97,7 +97,7 @@ use super::{LocalExecutor, LocalExecutorPoolBuilder};
 ///
 /// handles.join_all();
 /// ```
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum Placement {
     /// For the `Unbound` variant, the [`LocalExecutor`]s created by a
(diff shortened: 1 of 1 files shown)


glommio/src/executor/placement/mod.rs
```diff
@@ -163,6 +163,18 @@ pub enum Placement {
     Custom(Vec<CpuSet>),
 }
 
+impl Clone for Placement {
```
glommer, line 166: The tests are great, but why do we need to implement this manually?
I figured just deriving Clone would work

davidblewett: I just spaced it and had "implement the trait" stuck in my head.

glommer: Np. Let's use derive though because it is cleaner, but happy to take the tests. I had sent a PR for this as well, will close mine in favor of yours

Issue #401, opened 2021-08-30
Memory leak: task not released after completion


thirstycrow:
Tasks are not deallocated after completion since 26bc054. A little program to show the problem can be found here: https://github.com/thirstycrow/glommio/commit/192f24701a55ea3c068bb0895c3816c882c90fbb

glommer:
Thanks. I will take a look tomorrow.

thirstycrow:
Spent some time trying to understand how the ref count works, now I'm thinking that ref count should be initialized to 0 instead of 1, because after 26bc054 it will always be incremented when get scheduled.

glommer:
That intuitively makes sense. I will have to dig deeper to validate. Does that fix the issue for you, though? It's an important datapoint

thirstycrow:
It seems fixed. The memory footprint is not growing unboundedly now.

glommer:
@thirstycrow you are right.

Would you like to send a PR to fix that ?
Unit tests would be welcome, we can probably have some very low level ones in raw.rs

thirstycrow:
There's still some kind of memory leaking in our application, but this time it only happens on one of all shards. Have to solve it first. I'll send a PR for this then, if it's not too late.

glommer:
It's never too late! And it indeed is best to figure out the entire issue first to be sure

thirstycrow:
Close this in favor of #404

PR #404, opened 2021-09-06
Fix task memory leak

thirstycrow:
Three issues that prevent completed tasks from destroy are fixed:
- The ref count was initialized to 1, incremented on schedule, and decremented after run, so it will never going to be 0
- ~~The ref count could be updated on a foreign executor, so the relax ordering is not enough~~
- There's a chance that a foreign waker is dropped when the task is already complete. An additional foreign wake is necessary in this case.

Related issue: #401

### DIFF
--- a/glommio/src/task/header.rs
+++ b/glommio/src/task/header.rs
@@ -4,12 +4,11 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020 Datadog, Inc.
 //
 use core::{fmt, task::Waker};
-use std::{
-    cell::Cell,
-    sync::{
-        atomic::{AtomicI16, Ordering},
-        Arc,
-    },
+#[cfg(feature = "debugging")]
+use std::cell::Cell;
+use std::sync::{
+    atomic::{AtomicI16, Ordering},
+    Arc,
 };
 
 use crate::{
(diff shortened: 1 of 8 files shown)


glommio/src/task/header.rs
```diff
@@ -82,12 +89,38 @@ impl Header {
         // Put the waker into the awaiter field.
         abort_on_panic(|| self.awaiter = Some(waker.clone()));
     }
+
+    #[allow(dead_code)]
```
glommer, line 93: isn't it better to make this conditional on `debugging`?

glommer:
Why isn't relaxed enough? Being updated on a foreign executor should not require ordering stronger than relaxed.
Usually ordering is needed when there are other memory accesses that need to be synchronized before or after this address.
If all we're doing is testing the atomic variable, relaxed should be enough.

thirstycrow:
You're right. I had some misunderstanding about the memory orderings.

glommio/src/task/header.rs
thirstycrow: Yes, that would be better.

glommer (approved):

glommer:
@thirstycrow how slow is the task debugging thing you added? Any idea?

I'd love to see these tests executed as part of the `cargo test` default suite.

For now I am merging so we can fix the leak, but let's consider separately adding that to the main execution flow.

Thanks!

thirstycrow:
> @thirstycrow how slow is the task debugging thing you added? Any idea?
It depends on how many tasks needs to be tracked. The default filter function let all labelled tasks pass. A custom one with lower sample rate could reduce the impact. In the benchmark for our app, requests latency raise from 2ms to 10s with the default filter function, and reduced to 10ms with a sample rate of 0.1

PR #407, opened 2021-09-10
remove alpha wording from Readme.

glommer:
We are getting very close to a point in which I am comfortable moving
Glommio out of alpha.

In particular, the limitations mentioned are either things that we fixed
already (NVMe only, lack of async truncate) or we are unlikely to anyway
(like a memory allocator, which should frankly be a separate crate) and
I/O scheduler (we have something basic now for I/O dedup, but it is
unclear we need more than that).

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ or really just chat with friends
 
 ## What is Glommio?
 
-
 Glommio (pronounced glo-mee-jow or |glomjəʊ|) is a Cooperative Thread-per-Core crate for
 Rust & Linux based on `io_uring`. Like other rust asynchronous crates it allows
 one to write asynchronous code that takes advantage of rust `async`/`await`, but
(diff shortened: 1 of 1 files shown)


bryandmc:
I have mixed feelings but I also don't want to replicate what some other crates do by never hitting 1.0 .. some crates make 1.0 too big of a step and because of that, never feel like they are "ready" .. Any reservations? I'll do a once-over on the API and see if there's anything obvious that would cause us pain down the road.. tokio had an extremely long road to 1.0 but I think they had a unique situation (they existed with old futures, etc..).

glommer:
Fair point, but I'd like to at least remove the alpha wording and warnings, since most of them are either fixed or proved to be not that important.

I am happy to leave it brewing for a bit longer.

bryandmc:
I can agree with that 100%. Most are either fixed it less relevant so I think that's fair.

Mart-Bogdan:
> In particular, the limitations mentioned are either things that we fixed
> already (NVMe only, lack of async truncate) or we are unlikely to anyway
> (like a memory allocator, which should frankly be a separate crate) and
> I/O scheduler (we have something basic now for I/O dedup, but it is
> unclear we need more than that).

That's great news. But unfortuantely only readme was updated, but not documentation:
https://docs.rs/glommio/latest/glommio/#current-limitations

(exact version link https://docs.rs/glommio/0.7.0/glommio/index.html#current-limitations )

glommer:
Thanks for catching this.

I sent the following PR : https://github.com/DataDog/glommio/pull/572

PR #408, opened 2021-09-13
do not reallocate memory when processing shared channels

HippoBaro:
### What does this PR do?

The implementation used to iterate over the tree of wakers and would
recreate a new tree with the remaining ones. This is not efficient as it
leads to a high number of allocations. Instead, iterate over the
existing tree and reuse the memory in place.

Additionally:
* use the same tree to store wakers and checker functions to reduce
the number of lookups and simplify the lifecycle of both;
* use a `SmallVec` to store wakers and optimize for the case
where a single waker is consuming the shared channel.

I added some additional benchmarks to exercise this code path. This
is a small but measurable improvement, especially in the many-to-many
benchmark (which is really meant to be worst-case). See below.

Before:
```
cost of mesh message shared channel: 21.48µs, peers 8, capacity 1024
cost of mesh message shared channel: 32.648µs, peers 8, capacity 1
```

After:
```
cost of mesh message shared channel: 20.744µs, peers 8, capacity 1024
cost of mesh message shared channel: 25.872µs, peers 8, capacity 1
```

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/task/raw.rs
+++ b/glommio/src/task/raw.rs
@@ -215,30 +215,35 @@ where
     }
 
     /// Wakes a waker.
-    unsafe fn wake(ptr: *const ()) {
-        dbg_context!(ptr, "wake", {
-            let raw = Self::from_ptr(ptr);
-            if Self::thread_id() != Some(raw.my_id()) {
-                dbg_context!(ptr, "foreign", {
-                    let notifier = raw.notifier();
-                    notifier.queue_waker(Waker::from_raw(Self::clone_waker(ptr)));
-                });
-            } else {
-                let state = (*raw.header).state;
+    unsafe fn do_wake(ptr: *const ()) {
+        let raw = Self::from_ptr(ptr);
+        if Self::thread_id() != Some(raw.my_id()) {
+            dbg_context!(ptr, "foreign", {
+                let notifier = raw.notifier();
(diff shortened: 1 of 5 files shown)


glommio/src/task/raw.rs
```diff
@@ -215,64 +215,43 @@ where
...
-        dbg_context!(ptr, "wake", {
-            let raw = Self::from_ptr(ptr);
-            if Self::thread_id() != Some(raw.my_id()) {
-                dbg_context!(ptr, "foreign", {
-                    let notifier = raw.notifier();
-                    notifier.queue_waker(Waker::from_raw(Self::clone_waker(ptr)));
-                });
-            } else {
-                let state = (*raw.header).state;
+    unsafe fn do_wake(ptr: *const ()) {
```
glommer, line 218: this is unrelated to the channels work.

Usually I'd be okay with that, but we have had too many issues in this area. So please send this in a separate PR in case we need to bisect.

HippoBaro: Considering that the log is clean I was planning on rebasing. Taking that commit out wouldn't make much of a difference once rebased. Thoughts?

glommer: different commits is good enough

PR #409, opened 2021-09-13
update to rust 1.55 and apply Clippy suggestions

HippoBaro:
### What does this PR do?

Time flies, we ended up skipping 1.54 this time.

On a side node, we need to be more careful because some code was not
passing Clippy even in 1.53. We seem to have an issue whereby
contributors need to have a CircleCI account for it to detect their PR
and run their code.

This is problematic, and we should find a solution, otherwise we can
except more broken/unformatted/undocumented code to end up being merged.

Code that doesn't pass clippy isn't the end of the world as I can fix it
later on. Undocumented code though, that's another story!

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ orbs:
 jobs:
   fmt:
     docker:
-      - image: cimg/rust:1.53
+      - image: cimg/rust:1.55
     steps:
       - checkout
       - run:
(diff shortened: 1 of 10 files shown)


glommer (approved):

glommer:
Is this expected to pass clippy right now? It's not passing, but because you had made changes to the .ci file I would expect it to

HippoBaro:
Ah good catch; I broke the clippy invocation. I just fixed it. CI should pass yeah.

PR #412, opened 2021-09-16
API discoverability; align ourselves with the wider ecosystem 

HippoBaro:
### What does this PR do?

* Spawning tasks via a member function of `Task<T>` is not great from an
API discovery point of view but also means we often need to use the
starfish syntax such as `Task::<()>::local(...)`. A free function makes
it simpler for the type system to infer the return type of the `Task` to
create. This makes our spawning API similar to that of Tokio for
instance.
* Functions such as `create_task_queue` intuitively relate to the executor
and not to a `Task`. To make things easier, we define a free function
`executor()` that is now included in the prelude. It returns a proxy
struct to the `LocalExecutor`. This is where all executor functions
accessible to the user will be located moving forward.

Once Rust gets better thread-local-storage we will remove this proxy
entirely and return a direct reference to the `LocalExecutor` itself. Alas,
we can't do that yet.

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist


glommer:
On your first point, `Local` exists exactly to avoid the starfish notation. At this point I think `Local` is widespread enough that I would like to try and ping some known users like @thirstycrow and @bryandmc to make sure they are okay with any changes there (and for those API changes in general).

I agree about the second point about some of those functions naturally belonging to the executor. However is probably best to do `with_local_executor`, which is the same API as TLS uses.

I'll take a deeper look soon

examples/cooperative_preempt.rs
```diff
@@ -18,40 +18,40 @@ fn main() {
     //
     // There are two ways of yielding control:
     //
-    //  * Local::yield_if_needed(), which will yield if the task has run for too
-    //    long. What "too long" means is an implementation detail, but it will be
-    //    always somehow related to the latency guarantees that the task queues want
-    //    to uphold in their `Latency::Matters` parameter (or
+    //  * glommio::executor().yield_if_needed(), which will yield if the task has
```
glommer, line 21: probably best if `yield_if_needed` is totally free-floating, and not pegged to the executor.
It's possible to optimize it further so that not even an indirection to the executor is necessary (seastar, for example, just saves the address of the yield variables in a TLS variable). While their yield is a bit different, we could in principle use a similar optimization

examples/cooperative_preempt.rs
```diff
@@ -18,40 +18,40 @@ fn main() {
...
+    //    run for too long. What "too long" means is an implementation detail, but
+    //    it will be always somehow related to the latency guarantees that the task
+    //    queues want to uphold in their `Latency::Matters` parameter (or
     //    Latency::NotImportant).
     //
-    //  * Local::later(), which will yield immediately (execute the rest of the
-    //    function later).
+    //  * glommio::executor().later(), which will yield immediately (execute the
```
glommer, line 27: BTW, the wording on later's doc has to be updated. It moves the entire task queue out of the executor (to move just the current task you can call directly `futures_lite::yield_now()`, which is standard Rust.

That also probably means it could use a name change. `later()` is what seastar does, but it doesn't mean much. (And seastar's later is closer to futures_lite::yield_now(), it doesn't yield the entire task queue). 

Suggestions ?

HippoBaro: Tokio uses `coop()` as in "cooperate" which I believe works here. wdyt?

examples/cooperative_preempt.rs
HippoBaro: `yield_if_needed` is one of those that fits everywhere. It fits being in the `Task` itself because we're yielding a task. But it also fits being in the executor as it relies on the internal state to make a decision. 

I could do as I did for `local` and `local_into`, i.e. make them part of the executor struct but also add a free function alias for convenience. Let me know what feels right to you!

glommer: I agree that `yield_if_needed` can be accessed through many places as a concept, I am just concerned about the optimization opportunities that may be lost, but maybe that's a different conversation.

examples/cooperative_preempt.rs
glommer: I don't like `coop`. That is a lot similar in principle to `yield_if_needed`. It will count however many times you executed and make a decision based on that. This doesn't: it is unconditional. You call `later()`, it will yield.

What we should communicate instead is that unlike `yield_now`, the entire task queue is being yielded. Something like `yield_now_but_the_entire_task_queue_not_just_the_task`. Except of course, with a nicer name

HippoBaro: What about `yield_task_queue()`?

glommer: I am between that an `yield_task_queue_now`. Let's give some time for other people to comment

bryandmc:
I just wanna mention that I am gonna make some more extensive comments later today after work, but that I do have some comments.

HippoBaro:
@bryandmc Are you still planning to comment on this PR?

glommio/src/executor/mod.rs
```diff
@@ -2314,7 +2516,7 @@ mod test {
         );
         let start = getrusage_utime();
         ex.run(async {
-            Local::local_into(
+            crate::local_into(
```
bryandmc, line 2519: why is this different than the spawn terminology?

examples/cooperative_preempt.rs
bryandmc: I think I like yield_task_queue might be descriptive enough.. what additional context are we getting from the "now" ? Maybe I'm just missing it..

glommio/src/executor/mod.rs
```diff
@@ -2546,22 +2749,22 @@ mod test {
             let tq2_count = Rc::new(RefCell::new(vec![0, 0]));
             let now = Instant::now();
 
-            let t1 = Local::local_into(
+            let t1 = crate::local_into(
```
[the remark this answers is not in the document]
HippoBaro: I named the free function `spawn_` to make it clearer what it does. But looking back at it it's just confusing. Would you prefer both to be called `spawn_local_*` or `local_*`?

glommio/src/executor/mod.rs
```diff
@@ -2394,26 +2596,27 @@ mod test {
             .make()
             .unwrap();
         ex.run(async {
-            Local::local(async move {
+            crate::local(async move {
                 assert!(
-                    Local::executor_stats().total_runtime() < Duration::from_nanos(10),
+                    crate::executor().executor_stats().total_runtime() < Duration::from_nanos(10),
```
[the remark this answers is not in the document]
HippoBaro: There are two aspects to this. The first is discoverability, `executor()` makes it obvious that this is an interface to execute things. `Local` not only is confusing ("local" means many things depending on the context) but in our case, `Local` is a specialization of `Task`. Knowing which one to use when is confusing!

The second is that this opens the door to write code like such:

```rust
let ref executor = glommio::executor();
for x in 0..100 {
  executor.spawn_local(async move { ... })
}
```

There is a single thread-local deref in the above, vs 100 using the `Local` API as it exists today. I know this is detail, but TLS in rust is really not great as of today, so making this cheaper is a win.

glommio/src/executor/mod.rs
HippoBaro: See my reply above, I had good intentions but I made this more complicated than it needs to be. Will rename to `spawn_*` here as well.

glommio/src/executor/mod.rs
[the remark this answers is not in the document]
bryandmc: I kinda like spawn_local.. I think there's strong precedence for this because of tokio and it's 'spawn' method. I know it's longer but if this is about discoverability, that's the solution I think makes the most sense. I guess we could debate that it could just be 'spawn' but it might be unclear that we are always single-threaded so I think spawn_local is the simplest.

glommio/src/executor/mod.rs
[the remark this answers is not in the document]
bryandmc: At first I didn't understand what you mean but now I get it. Yes this makes sense. I have warmed up to it. I think it's a good solution.

examples/cooperative_preempt.rs
HippoBaro: I added a free function `yield_if_needed`, I think it makes sense to make it easily accessible just like the spawn functions :+1:

examples/cooperative_preempt.rs
HippoBaro: This is what I ended up doing:
* I added a `yield_now` function that calls `futures_lite::future::yield_now()` internally. I wanted to call it `yield` but it's a reserved keyword.
* I renamed `later` to `yield_task_queue_now`. I kept the `_now` to be consistent with `yield_now`.
* I updated the documentation to make it very clear what yielding a task queue vs a task mean.

glommio/src/executor/mod.rs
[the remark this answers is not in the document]
HippoBaro: All the functions that spawn now begin with `spawn_*`. Thanks!

glommio/src/executor/mod.rs
HippoBaro: This is fixed now :bow:

HippoBaro:
@bryandmc for some reason the `Re-request review` github button won't work, so I'm writing this message to let you know I've made changes since your last comments. Let me know what you think!

examples/cooperative_preempt.rs
```diff
@@ -16,42 +16,46 @@ fn main() {
...
+    //    implementation detail, but it will be always somehow related to the
+    //    latency guarantees that the task queues want to uphold in their
+    //    `Latency::Matters` parameter (or Latency::NotImportant).
     //
-    //  * Local::later(), which will yield immediately (execute the rest of the
-    //    function later).
+    //  * glommio::executor().yield_task_queue_now(), works like yield_if_needed()
+    //    but yields unconditionally.
+    //
+    //  * glommio::executor().yield_now(), which unconditional yield the current
```
glommer, line 30: isn't this the same as the one in futures_lite ? Why go through the executor ?

glommer (approved):

examples/cooperative_preempt.rs
HippoBaro: Just to be consistent with the other ones really

glommer: A reader will automatically ask themselves "but what is the difference between that and futures::yield_now()`? So it's good to mention that it is the same, just under a different interface.

And while we are at it, let's add doc links

HippoBaro: Alright, I did just that and made lots of corrections to our documentation throughout the crate while at it.  We now have better, more uniform comments and documentation with more links as appropriate.

bryandmc:
Glad to see you merged because I agree that it looks good!

PR #416, opened 2021-09-23
refactor `read_many` to take a stream of iovecs as input

HippoBaro:
Instead of taking an iterator of (offsets, positions), make it a stream.

The reason behind doing this is that using an iterator forces us to
schedule all the IO synchronously at call-time (or keep them in memory
which isn't great either). While this is fine when dealing with a small
number of requests (and -marginally- more efficient) it becomes
problematic as the number of IO requests increases.

Specifically, doing this may starve other tasks performing IO in the same
task queue since the reactor submits IO requests to the device in FIFO
order. Therefore, if two concurrent tasks call `read_many` they will
race and all the IO submitted by the first task will be processed before
a single IO from the second reaches the IO ring. This is obviously wrong
as we want some amount of fairness even within a common task queue.

Another use case is when performing parallel calls to `read_many` with
the intention of zipping the resulting streams. For instance, if some
data is spread in many files we may want to fetch data in parallel and
combine the result asynchronously.

The `read_many` stream now eagerly consumes the input stream until
it yields or until we reach 128 in-flight sources. This allows a maximum
of 128 sources to run concurrently. After that, the stream will wait until
the first source is fulfilled.

128 is a constant for now and was chosen to match the depth of our
io_uring submission queues. Although this value is a constant right now,
a further improvement would be to make it dynamic to maximize
throughput and lower IO latency on the fly.

Baby steps!

### DIFF
--- a/glommio/src/io/bulk_io.rs
+++ b/glommio/src/io/bulk_io.rs
@@ -32,12 +32,27 @@ impl IoVec for (u64, usize) {
     }
 }
 
-pub(crate) struct MergedIOVecs<V: IoVec>(pub(crate) VecDeque<V>, pub(crate) (u64, usize));
+pub(crate) struct MergedIOVecs<V: IoVec> {
+    pub(crate) coalesced_user_iovecs: VecDeque<V>,
+    pub(crate) pos: u64,
+    pub(crate) size: usize,
+}
+
+impl<V: IoVec> IoVec for MergedIOVecs<V> {
+    fn pos(&self) -> u64 {
+        self.pos
+    }
+
+    fn size(&self) -> usize {
+        self.size
+    }
+}
 
 impl<V: IoVec> MergedIOVecs<V> {
     fn combine(&mut self, mut other: Self) -> Option<Self> {
-        if self.1 == other.1 {
-            self.0.append(&mut other.0);
+        if self.pos() == other.pos() && self.size() == other.size() {
+            self.coalesced_user_iovecs
+                .append(&mut other.coalesced_user_iovecs);
             None
         } else {
             Some(std::mem::replace(self, other))
@@ -49,15 +64,16 @@ impl<V: IoVec> Iterator for MergedIOVecs<V> {
     type Item = (V, (u64, usize));
 
     fn next(&mut self) -> Option<Self::Item> {
-        if let Some(io) = self.0.pop_front() {
-            Some((io, self.1))
-        } else {
-            None
-        }
+        self.coalesced_user_iovecs
+            .pop_front()
+            .map(|io| (io, (self.pos(), self.size())))
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        (self.0.len(), Some(self.0.len()))
+        (
+            self.coalesced_user_iovecs.len(),
+            Some(self.coalesced_user_iovecs.len()),
+        )
     }
 }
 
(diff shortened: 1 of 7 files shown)


glommio/src/io/bulk_io.rs
```diff
@@ -72,240 +32,529 @@ impl IoVec for (u64, usize) {
...
+pub(crate) struct MergedIOVecs<V: IoVec>(pub(crate) VecDeque<V>, pub(crate) (u64, usize));
 
-/// A stream of ReadResult produced asynchronously.
-///
-/// See [`DmaFile::read_many`] for more information
-#[derive(Debug)]
-pub struct ReadManyResult<V: IoVec + Unpin> {
-    pub(crate) inner: OrderedBulkIo<ReadManyArgs<V>>,
-    pub(crate) current: Option<ScheduledSource>,
+impl<V: IoVec> Unpin for MergedIOVecs<V> {}
```
duarten, line 37: Why is this needed? I thought it would be automatically implemented.

duarten (commented):
Looks very nice, some small comments.

glommio/src/io/immutable_file.rs
```diff
@@ -394,15 +395,15 @@ impl ImmutableFile {
...
         iovs: S,
         max_merged_buffer_size: usize,
         max_read_amp: Option<usize>,
-    ) -> ReadManyResult<V>
+    ) -> ReadManyResult<V, impl Stream<Item = (ScheduledSource, ReadManyArgs<V>)>>
     where
-        V: IoVec + Unpin,
-        S: Iterator<Item = V>,
+        V: IoVec,
+        S: Stream<Item = V> + Unpin,
```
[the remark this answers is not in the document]
HippoBaro: Ah right thanks! I wish there was a clippy pass to enforce consistency for that. I'll investigate. This is not the first time I do this.

glommio/src/io/immutable_file.rs
```diff
@@ -558,7 +559,7 @@ mod test {
 
         {
             let iovs = vec![(0, 1), (3, 1)];
-            let mut bufs = stream.read_many(iovs.into_iter(), 0, None);
+            let mut bufs = stream.read_many(stream::iter(iovs.into_iter()), 0, None);
```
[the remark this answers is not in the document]
HippoBaro: No, there is no equivalent to `IntoIter` for streams. We may be able to use `.into()` though.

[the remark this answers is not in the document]
duarten: Oh, I thought `stream.read_many` was the `DmaFile` that was changed to accept a stream.

glommer (approved):
Looks great to me.

Once you and Duarte are on the same page for the comments we can merge

glommio/src/io/bulk_io.rs
```diff
@@ -72,240 +32,529 @@ impl IoVec for (u64, usize) {
     }
 }
 
-#[derive(Debug, Copy, Clone)]
-pub(crate) struct ReadManyArgs<V: IoVec + Unpin> {
-    pub(crate) user_read: V,
-    pub(crate) system_read: (u64, usize),
-}
+pub(crate) struct MergedIOVecs<V: IoVec>(pub(crate) VecDeque<V>, pub(crate) (u64, usize));
```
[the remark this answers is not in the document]
HippoBaro: Updated, it's more readable now thanks :bow:

glommio/src/io/bulk_io.rs
HippoBaro: `Unpin` is hard so I may be doing something wrong here. If I remove that I won't be able to compile until I require the user to pass an `Unpin` `IoVec` which I think is unnecessary (stream never require that from what I can see).

Open to suggestion on this one.

duarten: `IoVec` is `impl`'d by glommio, not by users, right? It seems like adding the `Unpin` bound to the returned `IoVec` type in `read_may` should be fine.

I also think these `impl Unpin` are correct, I was just surprised because I hadn't come across the need for them yet.

No objections on either direction you want to go.

HippoBaro: No, `IoVec` is a trait. Making it one makes the user code a lot more flexible because users can give us structs with a lot more context that's useful when the IO comes back. 

That being said, maybe there is no way around requiring them to implement `Unpin` themselves? I was trying to avoid that.

HippoBaro: I dropped 7e73bf1f721659d70c3389058c188b849ab95a74 so the `IoVec` the user gives us need to implement `Unpin`.

duarten (approved):

PR #417, opened 2021-09-23
add a stall detector that logs stacktraces of unyielding tasks

HippoBaro:
This PR adds a rudimentary stall detection mechanism to Glommio.

It measures the runtime of task queues and logs if they don't yield when
they are due. 

Knowing if a queue is stalling the reactor is nice, but finding the 
exact code location where the stall occurs remains very hard. Especially
if a task queue hosts many concurrent fibers.

To help with that, this PR introduces a stall detection mechanism that
records stack traces of stalling tasks. It works as follows:
* When a task queue is scheduled for execution, we set up a timer that
triggers some time after the queue is expected to yield (we add a 10%
error margin to avoid false positives). A background thread collocated
with the local executor waits on the timer at all times.
* When the timer fires, the thread sends a signal (`SIGUSR1`) to the 
local executor thread. Upon receiving the signal, the local executor 
records a complete trace of the local stack. Here we take advantage of 
the fact that by default, the kernel invokes signal handlers on top of 
the existing stack. i.e. the frames we record are those of the 
problematic user code that was meant to yield. The recorded frames are 
pushed on a non-blocking communication channel that links the signal 
handler and the local executor.
* When a task queue yields, the local executor disarm the timer and
checks the communication channel for potential recorded frames, if there
are any then we can conclude that the queue stalled, so we log them.

This code works in practice but has two major drawbacks:
* The timer dance is expensive; expect a high number of syscalls. 
Because of this runtime overhead, the stall detector is disabled by 
default. To opt-in, the feature `stall-detection` must be enabled at 
compile-time.
* We log stalls only after the queue yield. Therefore, if there is a bug
in your code and your queue never yields, the stall detector will never
log the code location that's at fault (even though we probably have
recorded the trace by then). The reason for this is that logging from a
signal handler is illegal.

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -17,6 +17,7 @@ readme = "../README.md"
 
 [dependencies]
 ahash = "~0.7"
+backtrace = { version = "0.3.61", optional = true }
 bitflags = "1.3"
 bitmaps = "3.1.0"
 buddy-alloc = "~0.4"
@@ -35,6 +36,7 @@ pin-project-lite = "~0.2"
 rlimit = "~0.6"
 scoped-tls = "1.0.0"
 scopeguard = "1.1"
+signal-hook = { version = "0.3.10", optional = true }
 sketches-ddsketch = "0.1"
 smallvec = { version = "1.7", features = ["union"] }
 socket2 = { version = "~0.3", features = ["unix", "reuseport"] }
@@ -45,6 +47,7 @@ typenum = "1.14"
 fastrand = "1.5"
 futures = "~0.3"
 hdrhistogram = "7.3.0"
+logtest = "2.0.0"
 pretty_env_logger = "~0.4"
 rand = "~0.8"
 tokio = { version = "1.12", default-features = false, features = ["rt", "macros", "rt-multi-thread", "net", "io-util", "time", "sync"] }
@@ -56,7 +59,7 @@ cc = "1.0"
 [features]
 bench = []
 debugging = []
-stall-detection = []
+stall-detection = ["signal-hook", "backtrace"]
 
 [[bench]]
 name = "executor"
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -56,6 +56,7 @@ cc = "1.0"
 [features]
 bench = []
 debugging = []
+stall-detection = []
 
 [[bench]]
 name = "executor"
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -1167,10 +1167,17 @@ impl LocalExecutor {
 
                 let (need_repush, last_vruntime) = {
                     let mut state = queue.borrow_mut();
-                    if runtime > Duration::from_millis(5) {
+                    // we consider a queue to be stalling the system if it returned more than a
+                    // millisecond after it should have. i.e. a task queue has 1ms after
+                    // `need_preempt()`returns true to yield.
+                    let stall_threshold = self
+                        .preempt_timer_duration()
+                        .saturating_add(Duration::from_millis(1))
+                        .max(Duration::from_millis(5));
+                    if runtime > stall_threshold {
                         warn!(
-                            "Reactor stall! {} : {} {:?}",
-                            &self.id, &state.name, runtime
+                            "Reactor stall! {} : {} {:?} > {:?}",
+                            &self.id, &state.name, runtime, stall_threshold
                         );
                     }
 
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -35,6 +35,7 @@
 mod latch;
 mod multitask;
 mod placement;
+mod stall;
 
 use latch::{Latch, LatchState};
 pub use placement::{CpuSet, Placement, PoolPlacement};
@@ -59,6 +60,8 @@ use scoped_tls::scoped_thread_local;
 
 use log::warn;
 
+#[cfg(feature = "stall-detection")]
+use crate::executor::stall::StallDetector;
 use crate::{
     error::BuilderErrorKind,
     parking,
@@ -122,7 +125,7 @@ pub(crate) struct TaskQueue {
     shares: Shares,
     vruntime: u64,
     io_requirements: IoRequirements,
-    _name: String,
+    name: String,
     last_adjustment: Instant,
     // for dynamic shares classes
     yielded: bool,
@@ -167,7 +170,7 @@ impl TaskQueue {
             shares,
             vruntime: 0,
             io_requirements: ioreq,
-            _name: name.into(),
+            name: name.into(),
             last_adjustment: Instant::now(),
             yielded: false,
         }))
@@ -936,6 +939,9 @@ pub struct LocalExecutor {
     parker: parking::Parker,
     id: usize,
     reactor: Rc<reactor::Reactor>,
+
+    #[cfg(feature = "stall-detection")]
+    stall_detector: StallDetector,
 }
 
 impl LocalExecutor {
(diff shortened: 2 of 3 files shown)


HippoBaro:
Marking this as draft because the unit tests stall (how ironic) when they run with a high number of threads. So far I've been unable to figure out why.

glommer:
I will read the code later today. One comment on the description: the feature flag thing is not acceptable. Ideally you should be able  to turn this on and off in an existing reactor. At a minimum at creation time with a flag, ideally even dynamically.

The timer dance is then hidden behind a conditional.

glommio/Cargo.toml
```diff
@@ -46,13 +48,15 @@ tokio = { version = "0.3.5", default-features = false, features = ["rt", "macros
...
 pretty_env_logger = "0.4"
+logtest = "2.0.0"
 
 [build-dependencies]
 cc = "1.0.47"
 
 [features]
 bench = []
 debugging = []
+stall-detection = ["signal-hook", "backtrace"]
```
glommer, line 59: this shouldn't be a feature, as I said before

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,283 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
+//
+#![allow(dead_code)]
```
glommer, line 6: it's never good to shut up clippy for this. Best to find what triggers it and remove it

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,283 @@
...
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread::JoinHandle,
+    time::{Duration, Instant},
+};
+
+#[cfg(feature = "stall-detection")]
+thread_local! {
```
glommer, line 19: TLS is a bad choice. This should be a member of the reactor.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,283 @@
...
+    static STALL_STRACE_COLLECTOR: (
+        crossbeam::channel::Sender<backtrace::BacktraceFrame>,
+        crossbeam::channel::Receiver<backtrace::BacktraceFrame>
+    ) = crossbeam::channel::bounded(1 << 10);
+}
+
+#[cfg(feature = "stall-detection")]
+lazy_static! {
+    static ref STALL_SIG_HANDLER: signal_hook::SigId = unsafe {
+        signal_hook::low_level::register(signal_hook::consts::SIGUSR1, move || {
```
glommer, line 29: the documentation should be very explicit that once the stall detector is enabled, the signal is registered.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,283 @@
...
+    },
+    thread::JoinHandle,
+    time::{Duration, Instant},
+};
+
+#[cfg(feature = "stall-detection")]
+thread_local! {
+    static STALL_STRACE_COLLECTOR: (
+        crossbeam::channel::Sender<backtrace::BacktraceFrame>,
+        crossbeam::channel::Receiver<backtrace::BacktraceFrame>
```
glommer, line 22: why do you need two channels? This flows in a single direction

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,283 @@
...
+                sys::timerfd::ClockId::CLOCK_MONOTONIC,
+                sys::timerfd::TimerFlags::empty(),
+            )
+            .map_err(std::io::Error::from)?,
+        );
+        let tid = unsafe { nix::libc::pthread_self() };
+        let terminated = Arc::new(AtomicBool::new(false));
+        let timer_handler = std::thread::spawn(enclose::enclose! { (terminated, timer) move || {
+            while timer.wait().is_ok() {
+                if terminated.load(Ordering::Acquire) {
```
glommer, line 71: Why do you need Acquire here?

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,283 @@
...
+        })
+    }
+
+    pub(crate) fn enter_task_queue(
+        &self,
+        queue_name: String,
+        start: Instant,
+        max_expected_runtime: Duration,
+    ) -> nix::Result<StallDetectorGuard<'_>> {
+        // We consider a queue to be stalling the system if it failed to yield in due
```
glommer, line 92: no, this should be configurable by the user when they configure the stall detector in the specific reactor.
Let's say you do know of a stall that happens sometimes, but don't prioritize fixing it now: at least your logs shouldn't be spammed by it

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    fn disarm(&self) -> nix::Result<()> {
+        self.timer.unset()
+    }
+}
+
+impl Drop for StallDetector {
+    fn drop(&mut self) {
+        let timer_handler = self.handler.take().unwrap();
+        self.terminated.store(true, Ordering::Release);
```
glommer, line 128: there should not be the need to set this terminated variable at all. When the timerfd is disarmed, we should just unregister it. If an event does fire, that means it happened close enough to the threshold, so why not log it?

glommio/src/executor/mod.rs
```diff
@@ -1054,23 +1065,35 @@ impl LocalExecutor {
...
+                            r.run();
+                            tasks_executed_this_loop += 1;
+                        } else {
+                            break;
+                        }
                     }
-                }
 
-                let runtime = time.elapsed();
+                    (time.elapsed(), tasks_executed_this_loop)
```
glommer, line 1095: I don't mind doing RAII for the stall detector, but then it should be dropped manually as soon as we exit the loop, and not at the end of the function. You don't want to include accounting time in the backtrace

glommer (changes requested):

glommio/src/executor/stall.rs
HippoBaro: Good point. They should be members of the `StallDetector` struct that I can access through the `LocalExecutor`. Will update.

glommio/src/executor/stall.rs
HippoBaro: That's one channel; you're looking at the sender/receiver pair.

glommio/src/executor/stall.rs
HippoBaro: I will make sure the documentation is crystal clear.

glommio/src/executor/stall.rs
HippoBaro: If I disarm the timer then the `read` syscall in the background thread will never return. Likewise, I can't close the fd since the kernel does reference counting on its side (and closing fds used across threads is a bad idea anyway.).

I'm not sure I can't get away without it but let me know if I'm missing your point.

glommer: it's running on a separate thread. So what if it doesn't return?

When you arm it again on the next run, then it might return. And you have to rearm it at every run anyway

HippoBaro: The `StallDetector` lifecycle is the same as the `LocalExecutor`. This code runs only when you drop a glommio executor, typically at the end of the application. 

I think it's good etiquette to release all the ressources once we are done with them. 🙂 

Just to be clear, in between task queues the timer is disarmed and rearmed just like you describe it.

glommer: Ok, I see. That's fine, then, but Relaxed atomics should be enough here.
As a rule of thumb, you only need stricter ordering than relaxed when there are two variables that need to be seen in the same order.

davidblewett: Fixed in `9287cf8678118e4231bf8daddacc8afacf69da57`.

HippoBaro:
Closing in favor of #499

PR #422, opened 2021-09-25
Make CpuSet an actual set interface

davidblewett:
### What does this PR do?

This PR adds set API to CpuSet, by delegating to `collections::HashSet`.

### Motivation

I am building a service abstraction that needs to be able to validate that multiple Glommio tasks do not share `CpuLocation` instances.

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/executor/placement/mod.rs
+++ b/glommio/src/executor/placement/mod.rs
@@ -55,6 +55,8 @@ use pq_tree::{
     Node,
 };
 use std::{
+    collections::hash_map::RandomState,
+    collections::hash_set::{Difference, Intersection, Iter, SymmetricDifference, Union},
     collections::HashSet,
     convert::TryInto,
     iter::FromIterator,
@@ -268,6 +270,54 @@ impl CpuSet {
     /// Consumes the `CpuSet` and returns the [`CpuLocation`]s.
     fn take(mut self) -> Vec<CpuLocation> {
         self.0.drain().collect()
+    }
+
+    // Delegate Set implementation
+    /// Returns `true` if the `CpuSet` contains a `CpuLocation`.
+    pub fn contains(&self, value: &CpuLocation) -> bool {
+        self.0.contains(value)
+    }
+
+    /// Visits the [`CpuLocation`]s representing the difference, i.e., the values that are in self but not in
+    /// other.
+    pub fn difference<'a>(&'a self, other: &'a Self) -> Difference<'a, CpuLocation, RandomState> {
+        self.0.difference(&other.0)
+    }
+
+    /// Visits the [`CpuLocation`]s representing the intersection, i.e., the values that are both in self and
+    /// other.
+    pub fn intersection<'a>(&'a self, other: &'a Self) -> Intersection<'a, CpuLocation, RandomState> {
+        self.0.intersection(&other.0)
+    }
+
+    /// Returns true if self has no [`CpuLocation`]s in common with other. This is equivalent to checking
+    /// for an empty intersection.
+    pub fn is_disjoint(&self, other: &Self) -> bool {
+        self.0.is_disjoint(&other.0)
+    }
+
+    /// Returns true if this `CpuSet` is a subset of another, i.e., other contains at least all the
+    /// values in self.
+    pub fn is_subset(&self, other: &Self) -> bool {
+        self.0.is_subset(&other.0)
+    }
+
+    /// Returns true if this `CpuSet` is a superset of another, i.e., self contains at least all the
+    /// values in other.
+    pub fn is_superset(&self, other: &Self) -> bool {
+        self.0.is_superset(&other.0)
+    }
+
+    /// Visits the [`CpuLocation`]s representing the symmetric difference, i.e., the values that are in self
+    /// or in other but not in both.
+    pub fn symmetric_difference<'a>(&'a self, other: &'a Self) -> SymmetricDifference<'a, CpuLocation, RandomState> {
+        self.0.symmetric_difference(&other.0)
+    }
+
+    /// Visits the [`CpuLocation`]s representing the union, i.e., all the values in self or other, without
+    /// duplicates.
+    pub fn union<'a>(&'a self, other: &'a Self) -> Union<'a, CpuLocation, RandomState> {
+        self.0.union(&other.0)
     }
 }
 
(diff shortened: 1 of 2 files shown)


glommer (approved):
Looks great to me.

Let's give @trtsl some time to comment

duarten (approved):

glommio/src/executor/placement/mod.rs
```diff
@@ -214,23 +248,76 @@ impl CpuSet {
 
     /// Returns a reference to the [`CpuLocation`]s currently included in the
     /// `CpuSet`.
-    pub fn as_vec(&self) -> &Vec<CpuLocation> {
-        &self.0
+    pub fn as_vec(&self) -> Vec<&CpuLocation> {
+        self.0.iter().collect()
```
HippoBaro, line 252: I don't think returning a vector of references is the right thing to do here. Since we changed the internal representation from a Vec to something else I'm not even sure this function makes sense anymore + you added an iterator interface that achieves the same thing.

glommio/src/executor/placement/mod.rs
```diff
@@ -214,23 +248,76 @@ impl CpuSet {
...
-        self.as_vec().len()
+        self.0.len()
     }
 
     /// Consumes the `CpuSet` and returns the [`CpuLocation`]s.
-    fn take(self) -> Vec<CpuLocation> {
-        self.0
+    fn take(mut self) -> Vec<CpuLocation> {
+        self.0.drain().collect()
+    }
```
HippoBaro, line 273: A better approach would be to add an `into_iter` function that returns an owning iterator that users can `collect()` themselves into whatever they need, vec or otherwise. That's the more common and flexible approach IMO.

glommio/src/executor/placement/mod.rs
```diff
@@ -214,23 +248,76 @@ impl CpuSet {
...
+
+    // Delegate Set implementation
```
HippoBaro, line 275: nit; this should have three `/` to be used by cargo doc

glommio/src/executor/placement/mod.rs
```diff
@@ -347,7 +434,7 @@ impl CpuSetGenerator {
     pub fn next(&mut self) -> CpuIter {
         match self {
             Self::Unbound => CpuIter::Unbound,
-            Self::Fenced(cpus) => CpuIter::from_vec(cpus.as_vec().clone()),
+            Self::Fenced(cpus) => CpuIter::from_vec(cpus.as_vec().into_iter().cloned().collect()),
```
HippoBaro, line 437: See my point above. By implementing `into_iter` directly we'd save some work here.

glommio/src/executor/placement/mod.rs
```diff
@@ -178,14 +184,42 @@ impl Default for Placement {
...
+        let mut set = if let Some(size) = size {
+            HashSet::with_capacity(size)
+        } else {
+            HashSet::new()
+        };
+        for cpu in iter {
+            set.insert(cpu);
+        }
+        Self(set)
+    }
```
HippoBaro, line 203: Seems like it could be made a lot simpler by calling `from_iter()`.

not tested but something like this should work:

```suggestion
    fn from_iter<I: IntoIterator<Item = CpuLocation>>(cpus: I) -> Self {
        Self(HashSet<CpuLocation>::from_iter(cpu.into_iter()))
    }
```

HippoBaro (changes requested):
Cargo fmt is complaining, otherwise this look like a net improvement to me. I left a few remarks on things I think could be made simpler/more idiomatic.

glommio/src/executor/placement/mod.rs
davidblewett: Should I remove this function? I refactored the one calling site to not use it. If we want to remove it, we need to do it before `1.0.0`.

glommio/src/executor/placement/mod.rs
davidblewett: I removed the caller; should I remove this one too?

HippoBaro: Please do, `.iter()` can be used to achieve the same result.

glommio/src/executor/placement/mod.rs
HippoBaro: Either keep this or remove it and add an owning iterator variant. I have a preference for the letter option.

glommio/src/executor/placement/mod.rs
davidblewett: Can you elaborate? I added an owning `into_iter` impl, but it's still going to have to clone isn't it (because self is `&mut`)? 

```
Self::Fenced(cpus) => CpuIter::from_vec(cpus.clone().into_iter().collect())
```

glommio/src/executor/placement/mod.rs
davidblewett: I didn't really want this in the published docs; IMO it's an implementation detail that we're delegating and not implementing directly.

glommio/src/executor/placement/mod.rs
HippoBaro: I thought `self` was owned here so I should’ve said `iter()` and not `into_iter()`. My point was that `as_vec()` was allocating a vec and we could avoid it. You recent does the trick ✅

HippoBaro (approved):
LGTM; will wait some to give @trtsl a chance to tune in before merging.

trtsl:
The changes look fine to me, and I think @HippoBaro's clean-up suggestions are good ones.  One observatinon I have is that the new public set operations (`symmetric_difference`, `intersection`, `union`, etc.) pretty much lock us in to using a `HashSet` (which is fine), but I wonder if it's cleaner to just provide a refrence to the internal `HashSet` and let the user do with it what they please, particularly as I'm not sure how many users will use the set operations.  This was essentially the approach taken with `CpuSet:as_vec`.  Could potentially even just implement `Deref<Target = HashSet>` for `CpuSet` and then just remove all the set operations.

HippoBaro:
> The changes look fine to me, and I think @HippoBaro's clean-up suggestions are good ones. One observatinon I have is that the new public set operations (`symmetric_difference`, `intersection`, `union`, etc.) pretty much lock us in to using a `HashSet` (which is fine), but I wonder if it's cleaner to just provide a refrence to the internal `HashSet` and let the user do with it what they please, particularly as I'm not sure how many users will use the set operations. This was essentially the approach taken with `CpuSet:as_vec`. Could potentially even just implement `Deref<Target = HashSet>` for `CpuSet` and then just remove all the set operations.

I like the `Deref` option. That being said I'm not sure how implementing set-like functions will limit us in practice since a `CpuSet` will always support these operators regardless of the internal representation we choose. 
i.e. if we trade our current approach for the `Deref` one, then changing the `HashSet` to something else will be equally API breaking. 

Let me know your thoughts!

trtsl:
> That being said I'm not sure how implementing set-like functions will limit us in practice since a `CpuSet` will always support these operators regardless of the internal representation we choose

The return types of `difference`, `symmetric_difference`, `intersection`,  and `union` are `HashSet` specific, but since I don't think we're performance sensitive here, I can't think of a reason why we would change away from a `HashSet` in the future ... but then again, I thought the same thing about `Vec` since `CpuSet::as_vec` provided access to all the `CpuLocation`s.  I guess sometimes the convenience is nice.

trtsl:
Just to play devil's advocate: we could have also just provided a `CpuSet::as_set(&self) -> HashSet<...>` method and kept the `Vec` internally (saving the user from doing the conversion themselves using `CpuSet::as_vec`).  That is, since `TopolgoyIter` uses a `Vec`, we let the user request the conversion to `HashSet` rather than having the implementation convert `HashSet` to `Vec` every time a `Placement` is used.

HippoBaro:
> Just to play devil's advocate: we could have also just provided a `CpuSet::as_set(&self) -> HashSet<...>` method and kept the `Vec` internally (saving the user from doing the conversion themselves using `CpuSet::as_vec`). That is, since `TopolgoyIter` uses a `Vec`, we let the user request the conversion to `HashSet` rather than having the implementation convert `HashSet` to `Vec` every time a `Placement` is used.

Well in that case we don't need any `as_*` functions at all since we have iterators and `Vec` and `HashSet` implement `FromIterator` :smile:  

But IMO a hash set _feels_ like the right internal representation so I'm fine with what we have here.

PR #425, opened 2021-09-25
Don't run placement example

duarten:
### What does this PR do?

Marks an example with `no-run`.

### Motivation

The example requires the executing host to have 4 cpus, which may not be the
case. We could change it to 1 cpu, but then the example would look weird.

### Checklist


HippoBaro:
Ah sorry I fixed this already in c23e53e0fb8d4a3a0f69931d8b35868847b3eb4d 😬

duarten:
Closing :D

PR #426, opened 2021-09-27
skew the scheduler towards locality and throughput

HippoBaro:
This PR changes how the scheduler behaves within a task queue
to favor throughput and locality.

This is a trick from the Go scheduler to get most of the benefits of a
LIFO queue but preserve the fairness we get from a FIFO queue. This
approach eliminates the (potentially large) scheduling latency that
otherwise arises from adding scheduled runnable to the end of the
task queue.

The change works as such:
* scheduled tasks go in a "priority slot" instead of the queue. If
the slot is already taken, the previous task is replaced and the old one
goes at the back of the queue.
* when the scheduler retrieves the next task, it'll take the task from
the priority slot instead of the queue, if any.
* The `LocalExecutor` maintains an internal counter that is incremented
each time we run a task from the priority slot instead of the queue.
When this counter reaches a constant (256 as of this commit), the task
in the priority slot is pushed at the back of the queue, and we pop the
next task as normal.
* Each time we switch task queues, the priority slot is emptied such
that we pick from the FIFO queue the next time we visit it.

I added a new benchmark that simulates a worst-case scenario for the
scheduler. Those are the results for glommio master:

```
time to complete for 100 tasks: 40.715854ms (407.158µs per task)
time to complete for 1000 tasks: 397.54535ms (397.545µs per task)
time to complete for 10000 tasks: 3.81689191s (381.689µs per task)
```

And after I apply the changes in this PR:
```
time to complete for 100 tasks: 836.719µs (8.367µs per task)
time to complete for 1000 tasks: 1.014916ms (1.014µs per task)
time to complete for 10000 tasks: 2.421971ms (242ns per task)
```

I also get a 40% improvement when running the glommio SPSC
buffered shared channel benchmark!

Such wow!

### DIFF
--- a/glommio/src/executor/multitask.rs
+++ b/glommio/src/executor/multitask.rs
@@ -13,7 +13,7 @@ use crate::{
     task::{task_impl, JoinHandle},
 };
 use std::{
-    cell::RefCell,
+    cell::{Cell, RefCell},
     collections::VecDeque,
     future::Future,
     marker::PhantomData,
@@ -35,6 +35,10 @@ use std::{
 /// is woken. When it's woken up, its schedule function is called, which means
 /// the `Runnable` gets pushed into a task queue in an executor.
 pub(crate) type Runnable = task_impl::Task;
+
+/// An identity fingerprint of a runnable task; useful to compare two tasks for
+/// equality without keeping references
+pub(crate) type RunnableId = task_impl::TaskId;
 
 /// A spawned future.
 ///
@@ -110,10 +114,31 @@ impl LocalQueue {
     }
 }
 
+const MAX_UNFAIR_SCHEDULING: usize = 256;
+
 /// A single-threaded executor.
 #[derive(Debug)]
 pub(crate) struct LocalExecutor {
     local_queue: LocalQueue,
+
+    /// run_next, if present is a runnable that was scheduled by the previous
+    /// task and should be run next instead of what's in the local queue.
+    /// This approach eliminates the (potentially large) scheduling latency
+    /// that otherwise arises from adding the scheduled runnable to the end of
+    /// the run queue.
+    ///
+    /// Another way to look at this is as a 1-element LIFO buffer in front of
+    /// the fair (FIFO) task queue. The latter is an instrument for fairness
+    /// while this optimizes for throughput.
+    run_next: RefCell<Option<Runnable>>,
+    /// last_popped is the last runnable return by this executor
+    /// It is used to avoid placing the same task in run_next
+    last_popped: RefCell<Option<RunnableId>>,
+
+    /// run_next_count counts the number of throughput-oriented scheduling
+    /// decision we have made in a row. It is used to inject some fairness into
+    /// the mix.
+    run_next_count: Cell<usize>,
 
     /// Make sure the type is `!Send` and `!Sync`.
     _marker: PhantomData<Rc<()>>,
@@ -128,6 +153,9 @@ impl LocalExecutor {
     pub(crate) fn new() -> LocalExecutor {
         LocalExecutor {
             local_queue: LocalQueue::new(),
+            run_next: RefCell::new(None),
+            last_popped: RefCell::new(None),
+            run_next_count: Cell::new(0),
             _marker: PhantomData,
         }
     }
@@ -146,10 +174,7 @@ impl LocalExecutor {
             let tq = tq.upgrade();
 
             if let Some(tq) = tq {
-                {
-                    let queue = tq.borrow();
-                    queue.ex.local_queue.push(runnable);
-                }
+                tq.borrow().ex.push_task(runnable);
                 maybe_activate(tq);
             }
         };
@@ -159,6 +184,7 @@ impl LocalExecutor {
         task_impl::spawn_local(executor_id, future, schedule)
     }
 
+    /// Creates a [`Task`] for a given future and runs it immediately
     pub(crate) fn spawn_and_run<T>(
         &self,
         executor_id: usize,
@@ -170,6 +196,7 @@ impl LocalExecutor {
         Task(Some(handle))
     }
 
+    /// Creates a [`Task`] for a given future and schedules it to run eventually
     pub(crate) fn spawn_and_schedule<T>(
         &self,
         executor_id: usize,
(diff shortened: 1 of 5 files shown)


HippoBaro:
**Note that this PR breaks two Glommio unit tests**. I removed them. These unit tests weren't testing for correctness but were testing the order in which the reactor was running tasks within a common task queue. Since we don't make any guarantee as to that, I feel fine removing them.

@glommer although I think this is correct, I'd really like if you could take a good look at this. :bow:

glommio/src/timer/timer_impl.rs
```diff
@@ -881,23 +881,6 @@ mod test {
         });
     }
 
-    #[test]
-    fn basic_timer_action_instant_past_works() {
```
glommer, line 885: this test cannot be deleted.

It is testing something important: that a timer that expired in the past will still trigger.

glommio/src/executor/multitask.rs
```diff
@@ -110,11 +114,32 @@ impl LocalQueue {
...
 }
 
+const MAX_UNFAIR_SCHEDULING: usize = 256;
+
 /// A single-threaded executor.
 #[derive(Debug)]
 pub(crate) struct LocalExecutor {
     local_queue: LocalQueue,
 
+    /// run_next, if present is a runnable that was scheduled by the previous
```
glommer, line 124: aren't we injecting this latency into the task that was sitting in the queue, though?

If you have a task that was sitting in the queue ready to run (let's say the result of some I/O that finally came back), and now you have a chain of 256 tasks that will hit the mini-LIFO, all we've done was move latency around, no?

This is not to say the change is bad. A newly created task running immediately tends to be good because it will dispatch I/O earlier.

I'd like to think a bit more about adversarial cases and what kind of workloads triggers them.

You said this technique is used by the Go scheduler, so maybe they have some write up on this ?

HippoBaro: I'm reasonably comfortable with the go's runtime code so that was my guide, you can find it here: https://cs.opensource.google/go/go/+/master:src/runtime/runtime2.go;l=627-639;drc=master;bpv=1;bpt=1  It is very well documented.

Additionally, this is the commit that introduced the mechanism in Go: https://cs.opensource.google/go/go/+/e870f06c3f49ed63960a2575e330c2c75fc54a34

glommio/src/timer/timer_impl.rs
HippoBaro: Apologies, this test is passing just fine. I added it back!

glommio/src/executor/multitask.rs
HippoBaro: The idea is to run a task's continuations as deep as possible to take advantage of data locality. As you said, the usual case is that we stop doing so well before reaching the limit of 256 because the task will return `Poll::Pending` because of IO or some other data dependencies.

As for starving other tasks in the queue, I share your concern and the limit of 256 was added to limit this. Go doesn't need it because it implements per-goroutine runtime quotas (a goroutine is a `Task` in Glommio). We have a limit per task queue instead.

> If you have a task that was sitting in the queue ready to run (let's say the result of some I/O that finally came back), and now you have a chain of 256 tasks that will hit the mini-LIFO, all we've done was move latency around, no?

Yes, up to the latency limit that's attached to the task queue. This system doesn't prevent the scheduler from yielding if necessary (`need_preempt()` is still called in between each time). 

So, fundamentally, the before/after situation hasn't changed. If you define a task queue with a latency requirement of 1ms and you spawn 1000 tasks inside, the scheduler guarantees that the queue (and not the tasks inside) will be visited a minimum of once per ~millisecond. So if each of those tasks are CPU hogs, then they will be visited about once per second at most.

glommio/src/executor/multitask.rs
```diff
@@ -185,10 +212,71 @@ impl LocalExecutor {
...
+            (Some(runnable), true) => {
+                // run the next task immediately
+                self.run_next_count.replace(self.run_next_count.get() + 1);
+                Some(runnable)
+            }
+            (Some(runnable), false) => {
+                // pop from the queue instead
+                self.run_next_count.replace(0);
+                let ret = self.local_queue.pop();
+                self.local_queue.push(runnable);
```
duarten, line 230: If the queue is empty, then we'll run nothing. Maybe these two lines should be swapped?

Also, maybe we should only reset the counter if the queue is not empty (with a different task)?

glommio/src/executor/multitask.rs
```diff
@@ -185,10 +212,71 @@ impl LocalExecutor {
...
+                ret
+            }
+            (_, _) => {
+                self.run_next_count.replace(0);
+                self.local_queue.pop()
+            }
+        };
+
+        if let Some(ref runnable) = ret {
+            self.last_popped.replace(Some(RunnableId::from(runnable)));
```
duarten, line 240: Shouldn't we always set `last_popped`, even if just setting it to `None`?

glommio/src/executor/multitask.rs
```diff
@@ -147,8 +175,7 @@ impl LocalExecutor {
 
             if let Some(tq) = tq {
                 {
-                    let queue = tq.borrow();
-                    queue.ex.local_queue.push(runnable);
+                    tq.borrow().ex.push_task(runnable);
                 }
```
duarten, line 179: Do we still need this scope?

duarten:
I like the idea! Some small comments.

glommio/src/executor/multitask.rs
HippoBaro: Good point!

glommio/src/executor/multitask.rs
HippoBaro: Uh oh I think you found a bug :sweat_smile:  

The reason I was poping before pushing is that I want to avoid the case where I have to realloc the circular buffer only to pop after just afterwards. I'll make sure I cover the edge-case where I pop nothing!

HippoBaro: I still want to reset the counter no matter what though. It doesn't matter whether we ended up running the priority runnable or not, the counter should be reset because **we tried** to run something else.

glommio/src/executor/multitask.rs
```diff
@@ -185,10 +212,71 @@ impl LocalExecutor {
...
+        // empty the priority slot so that we pick a task from the queue next time the
+        // scheduled runs this queue
+        let _ = self.last_popped.borrow_mut().take();
+        if let Some(runnable) = self.run_next.borrow_mut().take() {
+            self.local_queue.push(runnable)
+        }
+        self.run_next_count.replace(0);
+
+        !(self.local_queue.queue.borrow().is_empty() && self.run_next.borrow().is_none())
+    }
```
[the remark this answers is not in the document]
HippoBaro: I added a new `make_fair` function that does the priority task resetting while `is_active` remains at it was.

duarten (approved):
LGTM

glommer:
I have a few problems with this strategy:

First, please validate my understanding: what you are proposing we do here is prioritize explicit wakes within the same task queue by essentially allowing temporary reordering of those tasks that are chained together by the explicit wake.

That's fine, but:
* Explicit wakes outside the reactor are rare, and I have no reason to believe they are more important than wakes triggered by io_uring, for example. Most explicit wakes are in channels. Why is a channel wake more important than I/O completion?
* On that point, this only covers local channels. In most interesting applications, the channels that matter a lot are cross-executor.

So I think this risks unfairness for questionable gains. This could be made more interesting by somehow handling foreign wakes in a similar manner, and/or identifying I/O patterns down in the reactor that could also benefit from similar treatment.

bryandmc:
@HippoBaro - Hey can you post any/all of your benchmarks all together in one block, so it's easy to spot when people come across this thread? I will make other comments but so far I like this as an idea if we can mitigate some of the downsides (which it seems like you are doing..). Good stuff!

PR #427, opened 2021-09-28
improve the `LocalExecutor` builders

HippoBaro:
This PR makes a series of changes to the builders and placement
strategies:
* The `Placement` struct is renamed to `PoolPlacement` as it maps
placements for a pool of executors.
* Embed the number of executors to create directly in the placement 
policy. Instead of having two orthogonal configurations for a 
`LocalExecutorPoolBuilder`(the number of executors to create and the CPU
placement policy) that can conflict with each other (for instance if we 
provide a custom placement and a different number of shards to create), 
merge the two together. A `PoolPlacement` now also takes (when 
appropriate), the number of executors to create in addition to the 
`CpuSet` to use. The `LocalExecutorPoolBuilder` constructor now takes a 
`PoolPlacement` directly instead of the number of executors to create.
This simplifies the API and makes it harder to misuse.
* Align the `LocalExecutorBuilder` API with the pooled version. The 
pooled version is presently strictly superior to the non-pooled one 
because it uses the `PoolPlacement` logic. It is therefore possible to
spawn a single executor that's fenced to a collection of CPUs using the
pooled API, but not with the non-pooled one. With the existing 
`LocalExecutorBuilder`, an executor is either pinned to a specific CPU, 
or to none at all. To remedy this, this commit creates a new `Placement`
enum that contains a subset of placement strategies that make sense for 
a single executor. In turn, `LocalExecutorBuilder::new` now takes a 
`Placement` strategy and the `pin_to_cpu` API is removed.


glommer:
@trtsl

glommer (approved):
I like the fact that `Placement` is now the way to specify cpus in the single executor case. The only reason why it wasn't done that way of course is that the single executor preceded `Placement` so the consistency is good.

I am not sure I like the name `Exact`. I think I would prefer `Fixed`. Maybe there is even a case to be made that this is really just a special case of `Fenced`. Happy to hear from others.

glommio/src/executor/placement/mod.rs
```diff
@@ -67,37 +67,35 @@ use std::{
...
 /// ## Example
 ///
-/// Some `Placement`s allow manually filtering available CPUs via a [`CpuSet`],
-/// such as `MaxSpread`.  The following would place shards on four CPUs
-/// (a.k.a.  hyper-threads) that are on NUMA node 0 and have an even numbered
-/// package ID according to their [`CpuLocation`]. The selection aims to achieve
-/// a high degree of separation between the CPUs in terms of machine topology.
-/// Each [`LocalExecutor`] would be bound to a single CPU.
+/// Some `PoolPlacement`s allow manually filtering available CPUs via a
+/// [`CpuSet`], such as `MaxSpread`.  The following would place shards on four
```
davidblewett, line 80: This is only a variant of `PoolPlacement` right? Should we reword this to mention `Fenced` or `Exact`?

glommio/src/executor/placement/mod.rs
```diff
@@ -128,61 +125,124 @@ pub enum Placement {
...
+    /// [`CpuLocation`] such that the set of all CPUs selected has a low
+    /// degree of separation. The selection proceeds from all CPUs that are
+    /// online in a non-deterministic manner.  The `Option<CpuSet>`
+    /// parameter may be used to restrict the [`CpuSet`] from which CPUs are
+    /// selected; specifying `None` is equivalent to using
+    /// `Some(CpuSet::online()?)`.
     ///
     /// #### Errors
     ///
     /// If the number of shards specified in [`LocalExecutorPoolBuilder::new`]
```
davidblewett, line 154: This needs to be reworded to reference the variant parameter instead.

glommio/src/executor/placement/mod.rs
```diff
@@ -128,61 +125,124 @@ pub enum Placement {
...
+    /// [`CpuLocation`] such that the set of all CPUs selected has a high
+    /// degree of separation. The selection proceeds from all CPUs that are
+    /// online in a non-deterministic manner.  The `Option<CpuSet>`
+    /// parameter may be used to restrict the [`CpuSet`] from which CPUs are
+    /// selected; specifying `None` is equivalent to using
+    /// `Some(CpuSet::online()?)`.
     ///
     /// #### Errors
     ///
     /// If the number of shards specified in [`LocalExecutorPoolBuilder::new`]
```
davidblewett, line 139: This needs to be reworded to reference the variant parameter instead.

glommio/src/executor/placement/mod.rs
```diff
@@ -128,61 +125,124 @@ pub enum Placement {
...
     /// number of shards requested from the pool builder.
     ///
     /// #### Errors
     ///
     /// [`LocalExecutorPoolBuilder::on_all_shards`] will return `Result::Err` if
-    /// either of the follow is true:
-    /// * The length of the vector does not match the number of shards requested
-    ///   in [`LocalExecutorPoolBuilder::new`].
-    /// * Any of the [`CpuSet`]s is empty.
+    /// either of the follow is any of the [`CpuSet`]s is empty.
```
davidblewett, line 166: This phrase is confusing.

davidblewett (changes requested):
This looks good. There are some doc changes that need to be adjusted to the new `PoolPlacement` variant format.

trtsl:
I think the changes make sense.  Though I do have a couple of thoughts related to `PoolPlacement`s / `CpuSet`s that may be worth revisiting but are not directly related to this pull request:

- Given that the api now has several set operations, it may we worth having a closer look at `PartialEq`.  It is possible for two placements to compare equal but for the pool builder to use different cpu sets (because it relies on iterating a hashset and the number of cpus requested may be less than the set size); it is also possible for cpus sets to compare not equal but run on the same cpus (when the cpu sets are different but the number of cpus selected is less than set size).  Finally, `Placement::Custom` still relies on a `Vec` which is dependent on ordering.  One option: it would be possible for the implementation of `PartialEq` to check what `CpuSets` will in fact be used and then return the result ... perhaps even allowing e.g. `Placement::Custom` to equal `Placement::MaxSpread` if all the `CpuSet`s that the pool builder will use match.
- The changes from this weekend now allow construction of arbitrary `CpuSets` by the user (i.e. with invalid `CpuLocation`s).  An invalid CpuSet in the pool builder would likely fail silently.

glommio/src/executor/placement/mod.rs
```diff
@@ -128,61 +125,124 @@ pub enum Placement {
...
+///
+/// ## Example
+///
+/// Some `Placement`s allow manually filtering available CPUs via a
+/// [`CpuSet`], such as `MaxSpread`.  The following would place shards on four
+/// CPUs (a.k.a.  hyper-threads) that are on NUMA node 0 and have an even
+/// numbered package ID according to their [`CpuLocation`]. The selection aims
+/// to achieve a high degree of separation between the CPUs in terms of machine
+/// topology. Each [`LocalExecutor`] would be bound to a single CPU.
+///
```
trtsl, line 197: The docs here are still related to what is now`PoolPlacement::MaxSpread`.

glommio/src/executor/placement/mod.rs
HippoBaro: This comment describes the behavior of the following code snippet, so I don't think it's necessary to mention the others (all the variants are just below and individually documented).

glommio/src/executor/placement/mod.rs
```diff
@@ -128,61 +125,124 @@ pub enum Placement {
...
+    /// If the provided [`CpuSet`] contains no CPUs, the builder will fail.
+    Fenced(CpuSet),
+    /// The [`LocalExecutor`] is bound to the CPU specified by
+    /// `Exact`.
+    ///
+    /// #### Errors
+    ///
+    /// [`LocalExecutorBuilder`] will return `Result::Err` if the CPU doesn't
+    /// exist.
+    Exact(usize),
```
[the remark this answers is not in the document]
HippoBaro: `CpuLocation` contains information hard for the user to come by like numa node, package id, etc. I think it makes the API simpler to require a core ID, from them we can fetch the associated `CpuLocation`.

HippoBaro:
> The changes from this weekend now allow construction of arbitrary CpuSets by the user (i.e. with invalid CpuLocations). An invalid CpuSet in the pool builder would likely fail silently.

Yeah, this would be pretty bad. I've added a runtime check that asserts any `CpuLocation` we deal with is part of the CPU set available to the current process. 

I haven't acted on the `PartialEq` feedback here because I feel this would deserve a dedicated PR.

PR #430, opened 2021-10-02
Add a benchmark for measuring foreign wake latencies.

thirstycrow:
### What does this PR do?

Add a benchmark for foreign wake latency.


glommio/benches/foreign_wake.rs
```diff
@@ -0,0 +1,134 @@
...
+            Histogram::new(3).unwrap(),
+            Histogram::new(3).unwrap(),
+        ])
+    }
+
+    fn print(&mut self) {
+        let h3 = self.0.pop().unwrap();
+        let h2 = self.0.pop().unwrap();
+        let h1 = self.0.pop().unwrap();
+        let h0 = self.0.pop().unwrap();
```
glommer, line 29: I am honestly having a little bit of trouble following this code.

The only calls to `record` are in the glommio executor, but you have measurements that are for tokio -> tokio as well.

I think it would benefit from more explicit named calls. Maybe instead of a vector just add a hash with an enum key, and some comments on the code to help us follow.

Right now it is very. convoluted.

thirstycrow: It should look better now.

glommer: This looks a lot better indeed, way more readable. Thanks!

glommer:
I am happy with this, but it is failing a test. That's the new test that @HippoBaro added to make sure that licenses to new crates we add are compliant.

@HippoBaro , please advise.

HippoBaro:
The new test fails because one of our dependencies needs an update. See https://github.com/DataDog/glommio/pull/433. Once this is merged this PR will need to be rebased on top :bow:

PR #434, opened 2021-10-05
Reduce foreign wake latency

thirstycrow:
### What does this PR do?

Reduce foreign wake latency by sending notifications via the event fd installed on the latency ring. Notifications will only be sent if the task being waken was spawned on a queue with latency requirement. A comparison between with and without notifications produced by the foreign_wake benchmark:

```
$ RUST_LOG=debug RUST_BACKTRACE=1 cargo bench --bench foreign_wake 
    Finished bench [optimized] target(s) in 0.04s
     Running unittests (target/release/deps/foreign_wake-6d33a328eb2cc763)

Latency requirement: NotImportant
Label              Min   Mean    P99  P99.9    Max
glommio->tokio:  0.014  0.285  0.614  0.822  0.902
tokio schedule:  0.000  0.008  0.055  0.123  0.178
tokio process :  0.515  1.962  2.965  3.067  3.161
tokio->glommio: 11.816 96.668 97.983 99.135 99.135

Latency requirement: Matters(100ms)
Label              Min   Mean    P99  P99.9    Max
glommio->tokio:  0.000  0.169  0.399  0.518  0.564
tokio schedule:  0.000  0.018  0.100  0.176  0.213
tokio process :  0.513  1.794  2.415  2.717  2.873
tokio->glommio:  0.004  0.374  0.633  0.677  0.730
```

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist


glommio/src/sys/uring.rs
```diff
@@ -1604,14 +1601,16 @@ impl Reactor {
             membarrier::heavy();
             let events = process_remote_channels() + self.flush_syscall_thread();
             if events == 0 {
-                self.link_rings_and_sleep(&mut main_ring, &self.eventfd_src)
-                    .expect("some error");
+                if self.eventfd_src.is_installed().unwrap() {
```
glommer, line 1604: I wonder if we shouldn't register two eventfds for this? 

The wakeup code is tricky, and you are effectively adding one more condition.

glommio/src/task/header.rs
```diff
@@ -27,6 +27,9 @@ pub(crate) struct Header {
     /// Current state of the task.
     pub(crate) state: u8,
 
+    /// Latency matters or not
+    pub(crate) latency_matters: bool,
```
glommer, line 31: may be better to represent it as `Option<RawFd>`, and add `Some(eventfd)`. `None` would signal that we don't need to write to the eventfd.

glommio/src/task/raw.rs
```diff
@@ -221,6 +227,9 @@ where
             dbg_context!(ptr, "foreign", {
                 let notifier = raw.notifier();
                 notifier.queue_waker(Waker::from_raw(Self::clone_waker(ptr)));
+                if (*raw.header).latency_matters {
```
glommer, line 230: If we use a separate eventfd, we can then write `if Some(fd) = latency_matters { write(fd) }`

glommer (commented):
Great work, and great numbers!

My main concern is that the eventfd handling code has broken a lot in the past. It is tricky and hard to test. Because of that, I am not convinced if using the same eventfd for both wake up notifications is the way to go. We need to add more special cases to a code that is already full of them.

Have you considered using a separate eventfd ?

thirstycrow:
@glommer There seems to be an issue with the existing code here (https://github.com/DataDog/glommio/blob/master/glommio/src/sys/uring.rs#L997-L1016) in that if it failed to install the eventfd because of running out of sqe, the next time when the executor runs here, it will going to sleep without installing the eventfd, because the result of the eventfd_src is already taken during the last run. I'm not quite sure about it, maybe I missed something, but if the reasoning is correct, it could be solved by using the same eventfd for both foreign wake and sleep awakening.

glommer:
Ok. 

It seems you are also moving the eventfd handler from the main ring to the latency ring in all cases. That makes it simpler to unify. 

This seems like an ok approach to me. The sleep code is tricky, but it is the kind of tricky that no amount of code reading will convince me that is ok... there are always hard to predict corner cases.

@HippoBaro Would you mind taking this for a production spin? Please note that you have to remove `spin_before_park` from the application, otherwise the sleep side of eventfd is not doing anything and we're not really testing it.

thirstycrow:
@glommer Two more commits were added to avoid unnecessary writes to the eventfd, which is reduced by 86% in the `foreign_wake` benchmark.

thirstycrow:
`is_sleeping` and `notified` are used together in `notify_if_sleeping()`. It would cause problem if `write_eventfd` should be called but didn't. That would happen if `is_sleeping` returned true and `notified` also returned true but should return false.
But that seems impossible because of the membarrier right after `prepare_to_sleep` where `is_sleeping` is set true, is this correct?

```rust
    pub(crate) fn notify_if_sleeping(&self) {
        if self.is_sleeping.load(Ordering::Relaxed) {
            self.notify_if_needed();
        }
    }

    pub(crate) fn notify_if_needed(&self) {
        if !self.notified.swap(true, Ordering::Relaxed) {
            write_eventfd(self.eventfd_fd());
        }
    }
```

thirstycrow:
@glommer You're right, one boolean should be enough, the code is updated to work that way.

glommer:
We are still, as a side-effect, moving the memory order to relaxed where it was SeqCst before. I think it should be fine, but if you have the chance to run the shared channel torture benchmark that would be a good test.

Other than that I'm ready to merge this once it passes all tests

@HippoBaro I strongly suspect that the dependency bot is broken, though. Maybe let's disable it for now if it keeps breaking.

thirstycrow:
> I think it should be fine, but if you have the chance to run the shared channel torture benchmark that would be a good test.

The test failed at the 84 round out of 100 with `Too many open files`. It seems the eventfd is not closed even if the executor thread finish.

HippoBaro:
> @HippoBaro I strongly suspect that the dependency bot is broken, though. Maybe let's disable it for now if it keeps breaking.

The bot is not broken, if you look at the stdout of the job, it fails because `chrono 0.4.19` has a known security issue. Unfortunately, this time around it hasn't been fixed yet. I made this CI job (`Third-party / check`) non-mandatory so it's not a blocker for merging this. 


```
error[A001]: Potential segfault in `localtime_r` invocations
   ┌─ /github/workspace/Cargo.lock:19:1
   │
19 │ chrono 0.4.19 registry+https://github.com/rust-lang/crates.io-index
   │ ------------------------------------------------------------------- security vulnerability detected
   │
   = ID: RUSTSEC-2020-0159
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2020-0159
   = ### Impact
     
     Unix-like operating systems may segfault due to dereferencing a dangling pointer in specific circumstances. This requires an environment variable to be set in a different thread than the affected functions. This may occur without the user's knowledge, notably in a third-party library.
     
     ### Workarounds
     
     No workarounds are known.
     
     ### References
     
     - [time-rs/time#293](https://github.com/time-rs/time/issues/293)
   = Announcement: https://github.com/chronotope/chrono/issues/499
   = Solution: No safe upgrade is available!
   = chrono v0.4.19
     └── tracing-subscriber v0.2.25
         └── (dev) glommio v0.6.0
             └── (dev) examples v0.0.0

advisories FAILED, bans ok, licenses ok, sources ok
```

waynexia:
@HippoBaro How about letting that bot checks security problems against master branch or only trigger it when Cargo.toml changes? Control only emerging problems and leave the inspection of existing problems elsewhere.

PR #442, opened 2021-10-08
shared channel: ensure timely wake

trtsl:
### What does this PR do?

Improves on pull request #441 which allowed sender to wait on receiver to disconnect before waking.  With this request, senders should wake as soon as they are closed in other tasks.  The unit test has been updated to check this case explicitly.  Some additional minor refactoring.

### Motivation

Improve on the prior pull request.

### Related issues

Pull request #441

### Additional Notes

na

### Checklist

[y] I have added unit tests to the code I am submitting
~[] My unit tests cover both failure and success scenarios~
~[] If applicable, I have discussed my architecture~

### DIFF
--- a/glommio/src/channels/shared_channel.rs
+++ b/glommio/src/channels/shared_channel.rs
@@ -971,13 +971,15 @@ mod test {
 
     #[test]
     fn close_sender_while_blocked_on_send() {
-        use std::time::Instant;
+        use std::sync::{Condvar, Mutex};
 
         let (sender, receiver) = new_bounded(10);
+        let cv_mtx_1 = Arc::new((Condvar::new(), Mutex::new(0)));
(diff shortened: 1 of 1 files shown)


trtsl:
Hmm ... tests failed on the timers I'm using in the test.  Was trying to keep it simple rather than fiddle with synchronization, but unless you think just making the timers longer is appropriate (i.e. sleep longer in each timer), I'll push a test with condition variables later.

glommer:
I don't mind the timed tests, and I am guilty of that myself a couple of times when sync gets really hard... but if they are failing already right away on github actions, best to invest some time in making them more robust.

trtsl:
I don't see an impact on performance from `benches/shared_channel.rs`, though I am getting pretty wide variances so it's hard to say definitively.

Issue #445, opened 2021-10-19
Proposal: Expose poll style API for DmaFile


waynexia:
This proposes to add some interfaces for `DmaFile` (and maybe other necessary structures later).

`DmaFile`'s read/write implementations use `Source::collect_rw()` which relies on [poll_fn](https://docs.rs/futures-lite/1.12.0/futures_lite/future/fn.poll_fn.html) to convert a poll-based workflow into async fn. IMHO It would also be natural to expose those poll-based APIs alone with async version. This brings a bit more flexibility by providing another way to use `DmaFile` under the async environment. Otherwise, we need to do a lot of (bad) workarounds to "poll" those read/write tasks.

Please let me know your opinion!

PR #446, opened 2021-10-19
Poll-based interface for DmaFile

waynexia:
Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

### What does this PR do?

Expose poll-based interface for DmaFile.

### Motivation

#445 

### Related issues

#445 

### Additional Notes

### Checklist


glommer:
Hi @waynexia 

How is this better than using
```
let x = file.read_dma_aligned(...);
pin!(x);
futures_lite::poll_fn(|cx|, x.poll(cx));
```

I understand the above is obviously more code, but we don't expose a poll interface to anything, and the above is a idiomatic way how those things are usually done in Rust to the best of my knowledge

Also @duarten , if you want to chime in, please do!

waynexia:
Thanks @glommer. 

To use those async methods in the "poll" context, we need to store the future somewhere (to poll the same future again). And it's hard to write down the generated async future's type for keeping it without `dyn` and one extra indirection (box, raw pointer or reference, etc.) to my knowledge.

But this is not a big overhead after all. However exposing the poll-based interface obviously adds a lot of methods and structures. So this is just a draft proposal to bring a discussion on it.

duarten:
> **Except that it's an async function**: I can't use it directly from `poll_next` (a poll function is obviously not `async`).
> 

Why can't you call an `async` function from `poll_next`?

> Now back to the topic at hand: I could box the opaque type returned, store it on my stream struct, and call it a day. But I would have to use `dyn`, and I don't want the overhead. If the function had returned a struct implementing `Future`, it would be much simpler. There would be no boxing, no `dyn`, no overhead.

Why would you need to use `dyn`? An `async` function is a function returning a struct implementing `Future`. You do need boxing in any to be able to call `poll` on the pinned future, no?

waynexia:
Greeting @duarten!

>Why can't you call an async function from poll_next?

You can, but you can't `.await` it and you need to do something like [this](https://github.com/DataDog/glommio/pull/446#issuecomment-947790091) as "*a poll function is obviously not async*". And because we want to poll on the same future every time, we need to store it rather than create one each time. That is why @glommer's suggestion is not suited for this scenario to my understanding.

>Why would you need to use dyn? An async function is a function returning a struct implementing Future. You do need boxing in any case to be able to call poll on the pinned future, no?

Using `dyn` because the generated opaque Future struct can't be written down. Then the box is needed to give the `dyn Future<>` a determined size as well as to put it on the heap.

For the proposed poll api, you can either use the `.await` combinator like those `async` things, or store the returned Future structure and poll it at your opinion.

duarten:
> @duarten your example works because you use the stack for storage. Try to make this work with a stream, and you'll realize you can't always do that.

Hum.. not sure I understand. Fundamentally, using a generated type vs a named typed implies you have to use generics, but that should be orthogonal to storage?

> 
> I'm looking for a solution where boxing is a no-go (I don't want to allocate) and where `dyn` is a no-go (I don't want virtual function calls).
> 
> > You do need boxing in any case to be able to call poll on the pinned future, no?
> 
> That's not always true. My stream struct is already pinned (via the `pin_project` macro) so I can store futures directly.

I don't quite understand that either. I thought `pin_project` didn't pin anything, it only provided a convenient way to access pinned fields of a struct (I guess it uses `Pin::new_unchecked` for those), meaning you still had to ensure the `pin_project` struct is itself pinned, which you can either do with `Box::pin` or `Pin::new_unchecked`, no?

(To be clear, I'm not pushing back on anything, I'm just trying to understand the issue. As you see, I don't have a lot of experience with the lower levels of async rust.)

PR #451, opened 2021-10-27
Eliminate memory allocation and memory copy for network IO

thirstycrow:
### What does this PR do?

1. Allocate a receive buffer for each `GlommioStream` when it is created, so no more allocation will happen for each read operation. 
2. The interface of receive buffer is defined as the `RxBuf` trait to support future work for automatic buffer selection.
3. A benchmark (`benches/proxy.rs`) is added to compare the performance of glommio/tokio based proxy servers. (It is somewhat upsetting that tokio still out performs glommio in this benchmark with this PR)

### Additional Notes

It is not possible to well align the io-uring model with the `AsyncRead`/`AsyncWrite` traits. Compromise has to be made between performance and safety. The current state of this PR favors performance, assuming the buffer refs from caller of `poll_read`/`poll_write` would be available throughout the lifetime of the IO operation. This is aggressive and will cause problem is the user failed to use it sanely. Maybe a feature could be add to let the user decide what compromise should be make.


thirstycrow:
I reworked the `GlommioStream` to use poll + recv/send. And additionally applied an trick I call early-poll to improve the effectiveness of yolo calls. That is when we know the stream is not ready for the next operation by comparing the returned read/written bytes with the provided buffer length, we do a poll right after the current successful operation. So the next time the stream is polled for data, we should know the readiness immediately, and decide whether to do a yolo call accordingly. With this trick applied, benchmark result for the proxy scenario improved from 80k req/s to 102k req/s. It is, however, still behind tokio, which could do 114k req/s.

I also worked on a io-uring fast-poll implementation with similar tricks applied. It is ~5% slower than the poll + recv one in the proxy benchmark. However, in the echo benchmark (`benches/echo`), the fast-poll version seems slightly faster than the poll + recv one. So which one is better may be case dependent. I choose to submit the poll + recv version, because it is much simpler than the other.

bryandmc:
Is there a way we can just have both implementations? Because I'm wondering if different versions of io_uring might even affect which of these methods works better and I hate the idea of completely removing fast poll versions. Because with the right API the fast poll versions might work better, and I just don't wanna throw the baby out with the bath water (so to speak).. Logically, you would think the fast poll versions would be faster, depending on workload, of course. I will look at the specifics of this patch later but I just wanted to get this point out there quickly because I think it's worth considering.

bryandmc:
> > According to some benchmarks poll events on io_uring are actually faster.
> 
> @bryandmc Do you mean faster than epoll or faster than io-uring's fast poll? Could you provide a reference to the benchmarks?

I don't remember but someone came into here (or the chat site) and had been testing a bunch of stuff and found that the poll + recv worked well.

thirstycrow:
I used to use cargo features to have both implementations, but it's little bit hard to maintain, and we can have only one of them in a single build. There's another two approaches I can think of to make them coexists. One is to use generic and adapters. The other is to introduce some new fast-poll specific traits like the following. In this way we can use both of them even on a single stream, and the requirement for internal buffers could also be removed. 

```rust
pub trait FastRead<'a> {
    type R: FastReader<'a>;

    fn fast_read(&'a mut self, buf: &'a mut [u8]) -> Self::R;
}

pub trait FastReader<'a> {
    fn poll(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<usize>>;
}

pub trait FastWrite<'a> {
    type W: FastWriter<'a>;

    fn fast_write(&'a mut self, buf: &'a [u8]) -> Self::W;
}

pub trait FastWriter<'a> {
    fn poll(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<usize>>;
}

```

bryandmc:
@thirstycrow  -- I agree with your assessment about the cargo features.. In some ways they might just be better at different use-cases, and, if that's in the same app, you'd be stuck so I agree with that.

To do the fastpoll version correctly you would probably need an entirely different API that can eliminate the excessive copying. The only safe way would probably involve moving ownership of whatever buffer (or more accurately, a handle to the buffer) to us in the runtime and returning ownership with the result. 

Point being, these different ways of doing things don't need to both implement the exact same traits. I'm OK with losing AsyncRead for the fast poll version.. if that simplifies our constraint. The tough question of how exactly to implement this in the most ergonomic way, still remains. I suppose we could make it a wrapper type and then make it trivial to convert between them. Or we can make TcpStream further generic and delegate to that specific impl type when we need to make the crucial distinction.. but this will require a lot more work probably. 

We may want to break this all up into parts because there could be a decent chunk of work here, depending which way we go. 

Open to suggestions -- thoughts?

thirstycrow:
@bryandmc I did some experiments with the new traits idea. You're right. It's necessary to transfer the ownership of buffers to make a safe api. Besides, the lifetime constraints add extra difficulties. I'm stuck in the way to make it work with split, because it's not possible to borrow a reference from `RefCell` with specific lifetime. (I've been thinking for a while that since it should be safe to read and write a socket concurrently, but in Rust things could not be borrowed twice, maybe some unsafe code is worthy to achieve real concurrency for socket read and write?)

bryandmc:
Yeah unsafe can be used under the right circumstances and with enough verification. 

Are you referring to how you'd implement a custom split? That's what I'm assuming you mean.

thirstycrow:
Yes, that's exactly what I mean.

thirstycrow:
> With this trick applied, benchmark result for the proxy scenario improved from 80k req/s to 102k req/s. It is, however, still behind tokio, which could do 114k req/s.

According to the flamegraphs of tokio (epoll) and glommio (poll + recv) proxy servers. They both spent ~70% of cputime in `__libc_send`. The tokio server spent 14.78% cputime in `__libc_recv` and 2.31% in `poll_ready`. The glommio server spent ~10.93% of cputime in `__libc_recv`, 1.19% in `poll_read_ready`, 1.83% in `consume_completion_queue`, and 5.07% in `consume_submission_queue`. 

So the difference in performance I guess is primarily a result of the cputime used by `consume_submission_queue`. The one-shot behavior of IORING_OP_POLL_ADD gives us no choice when no more data is available but submitting one more poll request to get notified when data is ready. There's no such cost for an edge-triggered epoll.

thirstycrow:
There's IORING_OP_EPOLL_CTL in io-uring, so maybe we could chaining a epoll fd to get the best of both of the systems

bryandmc:
> There's IORING_OP_EPOLL_CTL in io-uring, so maybe we could chaining a epoll fd to get the best out of both of them.

I don't think it makes any sense to use EPOLL (it would have to have a massive benefit to justify the added complexity).. I think the goal should be to implement this simply with the poll events and the standard traits, and then another patch can handle creating an API to use fastpoll + custom traits. I think this is the only sensible way forward given the bench-marking results we are seeing. Because unfortunately we have to support these popular traits, so we just have to make the best of it.

thirstycrow:
Since there's multiple possible solutions to improve the performance, let's merge the benchmarks first (#455), so we'll have a standard to evaluate the improvements.

bryandmc:
I think that's a good idea, I will review that and we can merge it. I think we can likely break up the idea into two parts after that. One that adds poll based requests for standard traits and just leave the infrastructure of the fastpoll stuff alone, and then one that builds that out. Alternatively we can remove and re add the fast poll stuff but that's just a matter of taste.

thirstycrow:
close this in favor of #462

bryandmc:
Closing this, per request.

PR #457, opened 2021-11-12
fix foreign wakes

HippoBaro:
A subtle bug was introduced in a previous refactor. Two boolean values
were simplified to a single one and a bug sneaked in, marking an
awakened reactor as being asleep to remote ones. That lead to
significant performance slow-downs because shards would issue a syscall
each time they would wake up a foreign waker.

Also, I simplified the code a little and replaced two standalone atomic
load and swap operations with a single CAS one.

The relevant commit was 2fb4fb47ee3d3c0375ad9380c8a1892e4bc52404.

### DIFF
--- a/glommio/src/sys/mod.rs
+++ b/glommio/src/sys/mod.rs
@@ -319,13 +319,11 @@ impl SleepNotifier {
     }
 
     pub(crate) fn notify_if_sleeping(&self) {
-        if self.should_notify.load(Ordering::Relaxed) {
-            self.notify_if_needed();
-        }
-    }
-
-    pub(crate) fn notify_if_needed(&self) {
-        if self.should_notify.swap(false, Ordering::Relaxed) {
+        if self
+            .should_notify
+            .compare_exchange(true, false, Ordering::Relaxed, Ordering::Relaxed)
+            .is_ok()
+        {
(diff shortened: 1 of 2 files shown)


glommer (approved):

glommer:
@thirstycrow ^

HippoBaro:
Btw I should thank @thirstycrow because his commits are super incremental and small in size, so my bug-hunting experience went super smoothly! 

Also, I should add that the significant performance issues I was having were only manifesting when `spin_before_park` was enabled.

thirstycrow:
It is intended to notify on foreign wake to reduce tail latency if the tasks was spawned on a task queue where latency matters. So `notify_if_sleep` and `notify_if_needed` should not be merged as one. This PR effectively cancels the improvement by #434.

HippoBaro:
Ah that's unfortunate! Since tests were passing and it fixed my issue, I assumed It was good. I'll rework the fix.

glommer:
As a general rule, when a problem is found, is best to wait for the one who introduced the problem to comment, especially when I tagged him explicitly =)

That said, the commit did introduce a regression from a released version, so it is ok to effectively revert it until we figure it out.

PR #459, opened 2021-11-14
fix foreign wakes (ditto)

HippoBaro:
A subtle bug was introduced in a previous bug fix. We did fix the bug
but at the cost of negating the performance improvements, the work was
originally pulled for. This commit fixes the fix.

The relevant commit was dcc3ca11c2be6e28517a38c3046a17918716b0d5.

@thirstycrow thanks again for the report; I ran your benchmark with this
patch applied and this is what I get:

```
Latency requirement: NotImportant
Label              Min   Mean    P99  P99.9    Max
glommio->tokio:  0.016  0.796  2.427  3.749  3.833
tokio schedule:  0.001  0.165  0.567  1.022  1.105
tokio process :  0.516  1.807  2.857  3.251  3.413
tokio->glommio:  0.010 94.914 97.599 98.111 99.327

Latency requirement: Matters(100ms)
Label              Min   Mean    P99  P99.9    Max
glommio->tokio:  0.003  0.257  1.075  1.365  1.544
tokio schedule:  0.000  0.126  0.417  0.648  0.847
tokio process :  0.530  1.646  2.419  2.677  2.965
tokio->glommio:  0.111  1.266  2.039  2.415  2.495
```

### DIFF
--- a/glommio/src/channels/shared_channel.rs
+++ b/glommio/src/channels/shared_channel.rs
@@ -246,7 +246,7 @@ impl<T: Send + Sized> ConnectedSender<T> {
         }
         match self.state.buffer.try_push(item) {
             None => {
-                self.notifier.notify_if_sleeping();
+                self.notifier.notify(false);
                 Ok(())
             }
             Some(item) => {
@@ -314,7 +314,7 @@ impl<T: Send + Sized> ConnectedSender<T> {
     pub fn close(&self) {
         if !self.state.buffer.disconnect() {
             if let Some(r) = self.reactor.upgrade() {
-                self.notifier.notify_if_sleeping();
+                self.notifier.notify(false);
                 // wake other tasks `awaiting` the same sender letting them know sender is
                 // closed; we don't `unregister_shared_channel` here because
                 // another task could still be `await`ing this sender, in which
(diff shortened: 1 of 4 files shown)


HippoBaro:
@glommer working on this makes me realize that our shared channel doesn't use wakers internally. This has the disadvantage that the perf improvements offered by #434 don't apply there "for free", which would be nice to have.

Ideally, if a `LatencyMatter` task queue is waiting on a shared channel, then sending something on that channel from another executor should trigger an event on the latency ring even if we are not asleep. IIUC, it doesn't ATM.

glommer:
Yes, that's because the shared channels work predates the waker work. I always wanted to rework it, but never had the time.

glommer:
@HippoBaro you reverted the original work because it regressed a workload for you. Does this fix brings the waker performance back while keeping your workload unaffected?

HippoBaro:
> Does this fix brings the waker performance back while keeping your workload unaffected?

Yeah, this is deployed and gives the best of both worlds

HippoBaro:
@thirstycrow, I'm reluctant to merge this before you give the go-ahead since I f***ed it up once already. Let me know if this looks good to you.

thirstycrow:
@HippoBaro I'm alright with this one. Please go ahead.

PR #460, opened 2021-11-14
allow users to override the default io concurrency of `read_many`

HippoBaro:
The default of `RING_SUBMISSION_DEPTH` is good, but lowering IO
concurrency also limits the opportunities to deduplicate in-flight IO
requests. The appropriate balance between RSS and IO deduplication feels
very application-specific, so exposing some knob to allow the user to
override this is probably a good idea.


bryandmc (approved):
Looks good to me. I agree that this is application and use-case specific 'knob' ..

PR #461, opened 2021-11-15
fast `ReadResult` deref

HippoBaro:
The current `Deref` implementation of `ReadResult` is slow as it
traverses the underlying `ScheduledSource,` `Source,` and performs a
conditional jump depending on the type of buffer (IO buffer, heap, etc.)

This is all pretty expensive! Since a `ReadResult` always carries the
`ScheduledSource` with it, we can materialize the pointer to the
underlying memory and return it directly upon demand.

### DIFF
--- a/glommio/src/io/read_result.rs
+++ b/glommio/src/io/read_result.rs
@@ -23,7 +23,7 @@ impl core::ops::Deref for ReadResult {
 #[derive(Clone, Debug)]
 struct ReadResultInner {
     buffer: ScheduledSource,
-    offset: usize,
+    mem: *const u8,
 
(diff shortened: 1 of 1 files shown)


glommer (approved):

glommio/src/io/read_result.rs
```diff
@@ -23,7 +23,7 @@ impl core::ops::Deref for ReadResult {
 #[derive(Clone, Debug)]
 struct ReadResultInner {
     buffer: ScheduledSource,
-    offset: usize,
+    mem: *const u8,
```
glommer, line 26: can we at least use `NonNull` here?

If this is null your unsafe `deref` crashes.

Issue #458, opened 2021-11-14
Deadlock potential


thirstycrow:
There's a limited number of sqes in a ring. When sqe is running out, glommio submit the prepared sqes, and requests are left in the submission queue until the next time `consume_submission_queue` is called. Depending on the application logic, it is possible that a submitted sqe depends on an operation still in the submission queue to complete. If the executor fell asleep before the dependent operation is submitted, deadlock happens.

HippoBaro:
Would it be easy for you to write a simple example of such a program? I would love to have a unit test for that.

thirstycrow:
It's not an issue of the current version of glommio. Deadlock will only happen if `rush_dispatch` is removed from socket io. (I think that is necessary to improve performance as long as io-uring is used for socket io, because rush_dispatch counterweigh all benefits brought by io-uring for socket only applications) It can be reproduced with the proxy benchmark on this branch: https://github.com/thirstycrow/glommio/tree/tcp_poll_recv_no_rush

glommer:
The executor should not fall asleep if there are elements in the submission queue.
If it does, that's a bug.

glommer:
It could be that we're just forgetting to check that criteria. But "sleep" means "there's no more work to possibly be done". if there are requests sitting in a queue, by definition there is work to do and we can't sleep.

PR #462, opened 2021-11-15
Re-implement standard traits for socket using poll + recv

thirstycrow:
This is a continuation of #451. In this PR, the standard traits for `GlommioStream` is re-implemented using poll + recv. Performance of the `echo` and `proxy` benchmarks improved compared to current master branch, but the result could be even better if invocations to `rush_dispatch` could be removed from `poll_recv`. However, to do that, #458 needs to be addressed first, otherwise there is good probability for the proxy benchmark to deadlock in the 1000-client cases.

I also tested two epoll based implementations, which gives the best performance among all. The efficiency of them comes from edge-triggered epoll: no additional overhead will be introduced after when the socket fd is registered to the epoll. While in the poll + recv implementation, an additional SQE is needed each time we poll the socket fd, because poll on io-uring works in oneshot mode.

There may be a chance for an fast-poll based implementation to surpass the epoll ones in some scenarios. But I believe it could only happen after #458 is handled appropriately so we can avoid `rush_dispatch` in socket io.

Screenshot from 2021-11-15 11-13-07

### DIFF
--- a/glommio/src/io/buffered_file_stream.rs
+++ b/glommio/src/io/buffered_file_stream.rs
@@ -381,7 +381,7 @@ impl StreamWriter {
         res.map(|_x| ())
     }
 
-    fn flush_write_buffer(&mut self, waker: Waker) -> bool {
+    fn flush_write_buffer(&mut self, waker: &Waker) -> bool {
         assert!(self.source.is_none());
         let bytes = self.buffer.consumed_bytes();
         if !bytes.is_empty() {
@@ -409,7 +409,7 @@ impl StreamWriter {
                         .upgrade()
                         .unwrap()
                         .fdatasync(self.file.as_ref().unwrap().as_raw_fd());
-                    source.add_waiter_single(cx.waker().clone());
+                    source.add_waiter_single(cx.waker());
                     self.source = Some(source);
                     Poll::Pending
                 }
@@ -429,7 +429,7 @@ impl StreamWriter {
                     .upgrade()
                     .unwrap()
                     .close(self.file.as_ref().unwrap().as_raw_fd());
-                source.add_waiter_single(cx.waker().clone());
+                source.add_waiter_single(cx.waker());
                 self.source = Some(source);
                 Poll::Pending
             }
@@ -443,7 +443,7 @@ impl StreamWriter {
 
     fn do_poll_flush(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
         match self.source.take() {
-            None => match self.flush_write_buffer(cx.waker().clone()) {
+            None => match self.flush_write_buffer(cx.waker()) {
                 true => Poll::Pending,
                 false => Poll::Ready(Ok(())),
             },
@@ -497,7 +497,7 @@ macro_rules! do_seek {
                         .upgrade()
                         .unwrap()
                         .statx($fileobj.as_raw_fd(), &$fileobj.path().unwrap());
-                    source.add_waiter_single($cx.waker().clone());
+                    source.add_waiter_single($cx.waker());
                     $source = Some(source);
                     Poll::Pending
                 }
@@ -585,7 +585,7 @@ impl AsyncBufRead for StreamReader {
                         self.buffer.max_buffer_size,
                         self.file.file.scheduler.borrow().as_ref(),
                     );
-                    source.add_waiter_single(cx.waker().clone());
+                    source.add_waiter_single(cx.waker());
                     self.io_source = Some(source);
                     Poll::Pending
                 }
@@ -611,7 +611,7 @@ impl AsyncWrite for StreamWriter {
         }
 
         if !self.buffer.data.is_empty() {
-            let x = self.flush_write_buffer(cx.waker().clone());
+            let x = self.flush_write_buffer(cx.waker());
             assert!(x);
             Poll::Pending
         } else {
@@ -670,7 +670,7 @@ impl AsyncBufRead for Stdin {
                         self.buffer.max_buffer_size,
                         None,
                     );
-                    source.add_waiter_single(cx.waker().clone());
+                    source.add_waiter_single(cx.waker());
                     self.source = Some(source);
                     Poll::Pending
                 }
--- a/glommio/src/sys/source.rs
+++ b/glommio/src/sys/source.rs
@@ -40,6 +40,7 @@ use std::{
 pub(crate) enum SourceType {
     Write(PollableStatus, IoBuffer),
     Read(PollableStatus, Option<IoBuffer>),
+    PollAdd,
     SockSend(DmaBuffer),
     SockRecv(Option<DmaBuffer>),
     SockRecvMsg(
@@ -225,10 +226,17 @@ impl Source {
     // adds a single waiter to the list, replacing any waiter that may already
     // exist. Should be used for single streams that map a future 1:1 to their I/O
     // source
-    pub(crate) fn add_waiter_single(&self, waker: Waker) {
+    pub(crate) fn add_waiter_single(&self, waker: &Waker) {
         let mut inner = self.inner.borrow_mut();
-        inner.wakers.waiters.pop();
-        inner.wakers.waiters.push(waker);
+        let waiters = &mut inner.wakers.waiters;
+        match waiters.first_mut() {
+            Some(w) => {
+                if !w.will_wake(waker) {
+                    *w = waker.clone();
+                }
+            }
+            None => waiters.push(waker.clone()),
+        }
         debug_assert_eq!(inner.wakers.waiters.len(), 1)
     }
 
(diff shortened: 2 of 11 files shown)


glommio/src/io/buffered_file_stream.rs
```diff
@@ -409,7 +409,7 @@ impl StreamWriter {
                         .upgrade()
                         .unwrap()
                         .fdatasync(self.file.as_ref().unwrap().as_raw_fd());
-                    source.add_waiter_single(cx.waker().clone());
+                    source.add_waiter_single(cx.waker());
```
glommer, line 412: @HippoBaro please review this, since we had issues before with `add_waiter_single` vs multiple.

glommio/src/sys/source.rs
```diff
@@ -225,10 +226,17 @@ impl Source {
     // adds a single waiter to the list, replacing any waiter that may already
     // exist. Should be used for single streams that map a future 1:1 to their I/O
     // source
-    pub(crate) fn add_waiter_single(&self, waker: Waker) {
+    pub(crate) fn add_waiter_single(&self, waker: &Waker) {
         let mut inner = self.inner.borrow_mut();
-        inner.wakers.waiters.pop();
-        inner.wakers.waiters.push(waker);
+        let waiters = &mut inner.wakers.waiters;
+        match waiters.first_mut() {
```
glommer, line 232: Can you explain the purpose behind this ?

glommer (commented):
Looks straightforward to me.

I think I am a bit lost with the new signature of add_waiter_single and what we're trying to accomplish on that loop.

It used to be a simple push + pop, so if there's more logic now, I think that warrants a comment

glommer:
At some point I think it will be worth to understand why we're still behind in the 1-client case, since improving performance there is likely to lift all boats.

But it is very encouraging that this work move us in the right direction for the multi-client case.

How many CPUs did you have in this benchmark, btw ?

glommio/src/sys/source.rs
HippoBaro: From what I can see, this is meant to add the `Waker` if it's different than the one already in place. 
This looks wrong because `Waker::will_wake` makes zero guarantees. i.e., it could return `true` when two wakers would wake different tasks.

Maybe the usage is safe, and this is sound, but a comment would be great so we can understand what you're trying to do. This code is susceptible, and we introduced some pretty nasty bugs here before :grimacing:

thirstycrow: While I was trying to understand the difference between glommio and tokio's socket implementation, I found that tokio used the same trick to prevent avoid cloning a waker. So I modified `add_waiter_single` in the same way to exclude noise in the benchmark. It turns out it made little difference whether we have this change or not. I keep it however, because it is logically very likely the existing waker and the one being added will wake the same task.

thirstycrow:
In the 1-client case, code in between each task run is evaluated more frequently. The complexity of this part has a stronger influence to performance than in the multi-client case. In glommio, a preempt timer is armed each time before running a task queue, a user timer is armed if there's one and the latency ring is linked to the main ring before going to sleep. These are all done with io-uring. None of them could be submitted in batch, and io-uring enter will be triggered every time. All of these consume CPU cycles and postpone processing of incomming io events. 

In the poll + epoll implementation, there's one more thing, the epoll fd, needs to be linked to the main ring. Although saved the cost of submit sqes for polling, it did even worse in the 1-client case. 

In the reactor on epoll implementation, the rings are registered to epoll, which is waited on sleeping. The linking business are eliminated. Thus the results for 1-client case are improved. The percentage number could be improved up to ~93% if the preempt timer is removed, but in that way we will lose the ability to preempt. (Perhaps it could be optimized in the way by making `preempt_timer_duration` optional, and setting it to `None` for the default task queue)

thirstycrow:
There's 1 thread for the client and 1 for the server in the echo benchmark. There's 2 tokio threads shared by the client and the echo server, and 1 thread for the proxy server in the proxy benchmark.

glommer:
It's ultimately fine, because the 1-client cases are generally uninteresting and I'd gladly sacrifice performance there for other cases that make sense.

But something doesn't add up. This is work we do on idle. Let me think about this a bit more, but this should not block you in any way

bryandmc:
> I also tested two epoll based implementations, which gives the best performance among all. The efficiency of them comes from edge-triggered epoll: no additional overhead will be introduced after when the socket fd is registered to the epoll. While in the poll + recv implementation, an additional SQE is needed each time we poll the socket fd, because poll on io-uring works in oneshot mode.

Multishot poll seems like the solution to this but unfortunately it has to be optional until it becomes more mainstream.. I'm working on adding that but will wait for this to land to integrate it..

> There may be a chance for an fast-poll based implementation to surpass the epoll ones in some scenarios. But I believe it could only happen after #458 is handled appropriately so we can avoid `rush_dispatch` in socket io.

I have high hopes that we can create fast-poll based api's that are the fastest of them all -- but we will see if that can be achieved. In theory it seems like registered buffers fast-poll would be incredibly powerful for proxying if can be done right. I have a design document and I actually think (for the first time) it might be possible.. It's based on an email Jen's himself wrote about the workflow that might enable us to accomplish this without too much excessive traffic on the rings.

Great work btw! I guess we shouldn't be surprised that traits that were designed around epoll would force a usage pattern that favors epoll-style designs.. I'll do a closer review later but so far looks good.

glommio/benches/echo.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    block_on(ready_recv).unwrap();
+}
+
+fn start_tokio_server(runtime: &Runtime, port: u16) {
+    use tokio::{io::copy, net::TcpListener};
+
+    let (ready_send, ready_recv) = oneshot::channel();
+
+    runtime.spawn_blocking(move || {
```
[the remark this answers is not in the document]
thirstycrow: It is used to ensure the tokio servers will not use more than one cpu's resource, or it could be unfair to its glommio competitior. It would be more straightforward to create separate runtimes for client and server, though.

glommio/benches/echo.rs
```diff
@@ -0,0 +1,150 @@
...
+            .build()
+            .unwrap()
+            .block_on(async move {
+                let listener = TcpListener::bind(("0.0.0.0", port)).await.unwrap();
+                ready_send.send(()).unwrap();
+
+                let handle = Handle::current();
+                while let Ok((mut stream, _)) = listener.accept().await {
+                    handle.spawn(async move {
+                        let (mut inbound, mut outbound) = stream.split();
```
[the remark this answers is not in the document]
thirstycrow: Yes, it could be optimized once we have a specialized split alternative. And copy is ok here, since the streams are non-buffered by default now, there's not data copy in the user space any more.

glommio/benches/echo.rs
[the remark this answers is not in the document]
bryandmc: I just don't have a good idea of the implications, personally.. It just struck me as suspicious in some way. It could be fine, I just don't know if it will taint the output.

glommer:
@bryandmc do you still have any major concerns here?
 I like the direction and would like to see some progress here.

Any objections to merge?

glommer:
@thirstycrow if I understand correctly, the biggest performance is by just not using io_uring at all and going with epoll?
That's not what's going on in this patch, is it ?

glommer:
Also #458 should be handled by now.

bryandmc (approved):
@glommer  - looks good in my opinion. This doesn't add any epoll or anything like that just does the poll + nonblocking recv. So I say we ship it... its a strong basis for further work we've discussed in a few of these issues. 

EDIT:  techinically epoll works the best for these types of traits and I think that makes sense if you consider their design and how epoll works. They were designed to work with epoll specifically.

thirstycrow:
@glommer Yes, epoll matches better with the existing traits. But the combination of `fast-poll` + `multishot poll` looks more promising. Once that is realized, I think it's fine to have a slightly less efficient alternative with the old traits. So, I'll keep the epoll based implementations just for comparison of benchmark results, and not going to submit them for the moment.

HippoBaro:
I am okay with merging this, but the branch is out-of-sync with master. @thirstycrow can you rebase and make sure each commit passes fmt/clippy/tests individually? I would prefer to rebase this instead of squashing since this is not a trivial change :)

thirstycrow:
@HippoBaro It is rebased now. The commits except first one are used to fix fmt/clippy issues, so I simply squashed them.

@glommer Rushed dispatches are commented out in the last commit. I tested this version in a long running benchmark of our application, and no deadlock happened. Please let me know if you want it rolled-back.

thirstycrow:
I feel it annoying sometimes when `LICENSE-3rdparty.csv` is outdated by some updated dependency crates. Maybe it would be better if this file is maintained as a build artifact instead part of the source code.

glommer:
Agree about the license. I asked @HippoBaro to review that.

I am afraid lack of rushed dispatch may hurt throughput oriented benchmarks with a single socket, but for now let's lean on the data that we do have, which indicates that their removal helps.

PR #464, opened 2021-11-16
Prevent sleeping if the submission queue is not empty

### LINKED ISSUE #458: Deadlock potential

thirstycrow:
### What does this PR do?

Fix #458


glommer (commented):
Isn't it simpler to just check the size of the submission queues right before going to sleep ?

thirstycrow:
You're right. It is simpler and more efficient.

glommer (approved):

PR #467, opened 2021-11-26
use a drain iterator to empty the `IOVecMerger`s instead of replacing them

HippoBaro:
Using a drain iterator allows us to keep the internal memory around and 
avoid spurious `realloc`s in the subsequent iterations of the 
`read_many` stream.


glommer (approved):
awesome

PR #468, opened 2021-11-26
add fmt and clippy tests as part of the test process

glommer:
So we don't have to have a separate CI target.
This is easier to reason, and it's a single test

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/tests/linters.rs
+++ b/glommio/tests/linters.rs
@@ -1,0 +1,49 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2020
+// Datadog, Inc.
+#[cfg(test)]
+mod tests {
+    use std::process::Command;
+    #[test]
+    fn check_formating() {
+        let status = Command::new("cargo")
+            .args(["+nightly", "fmt", "--all", "--", "--check"])
+            .status()
+            .unwrap();
+        assert!(
+            status.success(),
+            "cargo fmt failed. Note that glommio uses nightly for formatting, so please invoke \
+             cargo with +nightly"
+        );
+    }
+
+    #[test]
+    fn check_clippy() {
+        let status = Command::new("cargo")
+            .args([
+                "clippy",
+                "--all-targets",
+                "--all-features",
+                "--",
+                "-D",
+                "warnings",
+            ])
+            .status()
+            .unwrap();
+        assert!(status.success());
+    }
+
+    #[test]
+    fn check_dependencies_sorted() {
+        let status = Command::new("cargo")
+            .args(["sort", "-w", "-c"])
+            .status()
+            .unwrap();
+        assert!(
+            status.success(),
+            "cargo-sort not installed or cargo.toml dependencies not sorted"
+        );
+    }
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,26 +10,6 @@ on:
     branches: [ master ]
 
 jobs:
-  fmt:
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Cache setup
-        uses: ./.github/actions/cache-setup
-
-      - name: Install nightly for rustfmt features
-        run: |
-          FMT_NIGHTLY=nightly-$(curl -s https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu/rustfmt)
-          rustup install nightly
-          rustup default "$FMT_NIGHTLY"
-          rustup component add rustfmt
-
-      - name: Check formatting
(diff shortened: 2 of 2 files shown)


duarten:
I feel this is harder to reason, as it replaces something simple, expected, and obvious (ci jobs) with something hidden and unexpected (tests calling into cargo) 😅

glommer:
Then let's give others some time to comment before I debug more why this is failing.

But the reason I like this approach is that `cargo test` should be part of everyone's local development workflow.
Many a time I had run cargo test, sent the PR, just to see that the PR failed on fmt or clippy. Integrating those reduces the surprise factor.

duarten:
> But the reason I like this approach is that `cargo test` should be part of everyone's local development workflow. Many a time I had run cargo test, sent the PR, just to see that the PR failed on fmt or clippy. Integrating those reduces the surprise factor.

True. Another approach for this would be to add a pre-commit hook.

glommer:
Adding @HippoBaro @bryandmc @thirstycrow

HippoBaro:
I believe this approach doesn't work because we rely on rust `nightly` for formatting. Since we use `stable` to run the tests, we end up with an incorrect result. 

> PS: Now that Rust 2021 has been released, maybe we should investigate whether we still need to use that weird setup. Is it possible the formatting options we use have been included in the stable channel?

glommer:
Well, the approach itself works. The current code indeed doesn't, but I am one debugging session away from figuring out why =)

glommio/tests/linters.rs
```diff
@@ -0,0 +1,30 @@
...
+#[cfg(test)]
+mod tests {
+    use std::process::Command;
+    #[test]
+    fn check_formating() {
+        let status = Command::new("cargo")
+            .args(["+nightly", "fmt", "--all", "--", "--check"])
+            .status()
+            .unwrap();
+        assert!(status.success());
```
HippoBaro, line 12: Can you add a failed assertion message here? A common mistake is to format using the stable channel so a little hint there would help people understand what's going on!

HippoBaro (changes requested):

HippoBaro:
Overall I like the idea + it doesn't prevent people used to the fmt+clippy workflow to keep doing what they are currently used to

HippoBaro:
Is there a reason why the `doc` test isn't part of this change? It would be nice for `cargo test` to fail if some APIs are not properly documented.

glommio/tests/linters.rs
```diff
@@ -0,0 +1,49 @@
...
+                "-D",
+                "warnings",
+            ])
+            .status()
+            .unwrap();
+        assert!(status.success());
+    }
+
+    #[test]
+    fn check_sort() {
```
HippoBaro, line 39: Can we rename this to something like `check_dependencies_sorted`?

glommer: done

glommer:
> Is there a reason why the `doc` test isn't part of this change? It would be nice for `cargo test` to fail if some APIs are not properly documented.

Not sure if that buys us much. I just removed a documentation from a public type, and got this:

```
$ cargo doc --all-features --no-deps
warning: missing documentation for an enum
   --> glommio/src/io/dma_file.rs:29:1
    |
29  | pub enum CloseResult {
    | ^^^^^^^^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> glommio/src/lib.rs:288:9
    |
288 | #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
    |         ^^^^^^^^^^^^

warning: `glommio` (lib doc) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
[glauber@trueserverstrongandfree glommio]$ echo $?
0
```

More importantly: 
1. both clippy and build complained about that as well.
2. `cargo doc` has side-effects (generates a whole doc directory). Not sure if this is advisable during test.

HippoBaro:
Ah, I see, in that case, this LGTM! If that's okay with you, I'll merge when the test all passes (except for the one you removed).

bryandmc:
maybe I should have mentioned this sooner but the other option is a git hook, right? Like a "push" hook that runs everything.. though people can always ignore them.. But it takes extra flags to their command.

HippoBaro:
I would be fine with also adding a hook that runs `cargo test` as part of the PR; @glommer wdyt?

glommer:
not sure what that buys us

HippoBaro:
> maybe I should have mentioned this sooner but the other option is a git hook, right? Like a "push" hook that runs everything.. though people can always ignore them.. But it takes extra flags to their command.

IMO these two things are orthogonal, this PR can be merged and is useful on its own

bryandmc:
> > maybe I should have mentioned this sooner but the other option is a git hook, right? Like a "push" hook that runs everything.. though people can always ignore them.. But it takes extra flags to their command.
> 
> IMO these two things are orthogonal, this PR can be merged and is useful on its own

Sorry misunderstood what this was for and only looked at the conversation not the code -- ignore me haha

PR #469, opened 2021-11-26
allow liburing to be vendored

### LINKED ISSUE #466: Support for offline vendored builds
The build.rs currently requires Internet access to get some required dependencies at build time. There should be a path for building from pre-fetched vendored sources. Currently unsure on the best approach at the moment.

This is required before we can start using this for projects in Pop!_OS. Interested in experimenting with a Linux-only async runtime with native io_uring support.

glommer:
A better long term approach is to move to a full rust implementation.
For now, allow liburing to be vendored by reading an environment
variable with its location. In the absence of that variable, we'll
still resort to the standard submodule update.

Fixes #466 (for now)


glommer:
@mmstick 
for your consideration

bryandmc (approved):
We can wait for the person who requested it but this looks good.

bryandmc:
are we good here?

PR #474, opened 2021-11-29
make it easy to consume a `DmaStreamReader` without copying

HippoBaro:
The `get_buffer_aligned` API could be better. It relies on user code to be
aware of buffer boundaries locations. This makes user code more
complicated and brittle because the buffer size will generally be a
configurable knob, subject to change.

Fear not, however, for there is a better way!

Instead of failing when reading too much in `get_buffer_aligned,` read
whatever we can from the current buffer and return the number of
remaining bytes. This allows the following:
* If the read fits inside a buffer, the user can consume the result
directly without any copy;
* If the read crosses buffer boundaries and the user can consume partial
results, it can do so in a loop without any copies;
* If the read crosses buffer boundaries and the user needs a complete
result, `get_buffer_aligned` may be used trivially in a loop that
concatenates the results in a reusable user-allocated buffer.

This PR also replaces the ad-hoc `poll_read` implementation for a proxy
call to `get_buffer_aligned`, resulting in a very nice reduction in complexity.

### DIFF
--- a/glommio/src/io/dma_file_stream.rs
+++ b/glommio/src/io/dma_file_stream.rs
@@ -9,8 +9,6 @@ use crate::{
     task,
     ByteSliceExt,
     ByteSliceMutExt,
-    GlommioError,
-    ResourceType,
 };
 use ahash::AHashMap;
 use core::task::Waker;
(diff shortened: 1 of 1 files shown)


HippoBaro:
Reviews are welcome but let's not merge this right now: I want to give this a spin on some real workload first :)

HippoBaro:
This code has been deployed for a little while now, and it seems to be working as expected. We have streams that go over every variable-sized item in a file. Their throughput has doubled, primarily because calls to `memcpy` have almost disappeared (some copying is still necessary when crossing buffers).

All in all pretty happy with the result!

glommio/src/io/dma_file_stream.rs
```diff
@@ -519,25 +493,20 @@ impl DmaStreamReader {
...
 
     /// Allows access to the buffer that holds the current position with no
-    /// extra copy
-    /// In order to use this API, one must guarantee that reading the specified
-    /// length may not cross into a different buffer.  Users of this API are
-    /// expected to be aware of their buffer size (selectable in the
-    /// [`DmaStreamReaderBuilder`]) and act accordingly.
+    /// extra copy.
+    ///
+    /// This function returns a tuple of a [`ReadResult`] and a [`usize`]. The
```
glommer, line 498: I am a fan of the approach. Well done.

Not sure I am a fan of the unnamed tuple. Why do you think this is preferable to a specialized struct ?

HippoBaro: Right, I agree. 

Come to think of it, this API should work like POSIX `read,` i.e., it should return what was read instead of what is left to read. What is left to read is actually very hard to know, given we could hit EOF at any time.

If I return what was read instead, then I don't need to return a tuple because, by definition, you can already get that with `ReadResult::len().` That will simplify things even more. Nice!

HippoBaro: Done!

HippoBaro:
An additional change I am considering making would be to change the name of the two functions `poll_get_buffer_aligned` and `get_buffer_aligned` in favor of `poll_read_direct` and `read_direct.` This makes sense because they are now direct replacements for the `AsyncRead` copying equivalents.

We could even create two traits: `AsyncReadDirect` and `AsyncReadDirectExt,` which would work much like the ones from `futures_io` do but would be copy-free. What do you guys think?

glommer:
We can add the traits, but the biggest advantage of having traits in this case is if there are other implementations. So probably best to wait until the need shows up

glommer (approved):

PR #476, opened 2021-11-30
add a way to measure and report IO latency

HippoBaro:
Add the plumbings necessary to measure various kinds of IO latency,

We now keep track of two kinds of IO latency: 
* The time sources spend in the ring, including the submission queue;
* The post-reactor scheduler latency, or the time the scheduler takes to
come back at the tasks that consume the IO;

We now export them as distribution in the `IoStats` struct. And this PR
adds distribution support to Glommio. We use "sketches" that give a 
statistical approximation of quantiles because we want the overhead 
(space and time) to be low when recording latencies.

A side effect of this is that pulling the stats from the reactor now
clears them (because a distribution needs to be cleared regularly to be
helpful). I don't expect this to be problematic because, in practice,
users want rates from the stats and keep the previous values in memory
to compute deltas. This should, therefore, greatly simplify this logic.

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -35,6 +35,7 @@ pin-project-lite = "~0.2"
 rlimit = "~0.6"
 scoped-tls = "1.0.0"
 scopeguard = "1.1"
+sketches-ddsketch = "0.1"
 smallvec = { version = "1.7", features = ["union"] }
 socket2 = { version = "~0.3", features = ["unix", "reuseport"] }
 tracing = "~0.1"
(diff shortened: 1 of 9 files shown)


glommer:
Is there any way we can make this something the user enable with a flag?
When I was playing with the stall detector, I made it so that it would be enabled/disabled dynamically.

For this, we should at least do it at executor creation time, but not unconditionally. Measuring time can be incredibly expensive.

The rest of the code looks ok and ready to go.

HippoBaro:
> Is there any way we can make this something the user enables with a flag? When I was playing with the stall detector, I made it so that it would be enabled/disabled dynamically.
> 
> For this, we should at least do it at executor creation time, but not unconditionally. Measuring time can be incredibly expensive.
> 
> The rest of the code looks ok and ready to go.

Fair point, I'll add an executor config to disable all that stuff!

Issue #477, opened 2021-12-10
Cargo test hangs


topecongiro:
Running `cargo test` hangs with test failures here and there in my environment.

<details><summary>cargo test output</summary><div>

```
running 360 tests
test channels::spsc_queue::tests::test_try_poll ... ok
test channels::spsc_queue::tests::test_try_push ... ok
test channels::spsc_queue::tests::test_wrap - should panic ... ok
test error::test::composite_error_from_into ... ok
test error::test::channel_wouldblock_err_msg - should panic ... ok
test error::test::channel_closed_err_msg - should panic ... ok
test error::test::enhance_error ... ok
test error::test::enhance_error_converted ... ok
test error::test::extended_file_err_msg - should panic ... ok
test error::test::extended_closed_file_err_msg - should panic ... ok
test error::test::extended_file_err_msg_unwrap - should panic ... ok
test error::test::file_wouldblock_err_msg - should panic ... ok
test error::test::queue_still_active_err_msg - should panic ... ok
test error::test::queue_not_found_err_msg - should panic ... ok
test error::test::rwlock_wouldblock_err_msg - should panic ... ok
test error::test::rwlock_closed_err_msg - should panic ... ok
test error::test::semaphore_closed_err_msg - should panic ... ok
test error::test::semaphore_wouldblock_err_msg - should panic ... ok
test executor::placement::test::cpuset_difference ... ok
test executor::placement::test::cpuset_disjoint ... ok
test executor::placement::test::cpuset_intersection ... ok
test executor::placement::test::cpuset_iterate ... ok
test executor::placement::test::cpuset_subset_and_superset ... ok
test executor::placement::test::cpuset_symmetric_difference ... ok
test executor::placement::pq_tree::test::construct_pq_tree_spread ... ok
test executor::placement::test::cpuset_union ... ok
test executor::placement::pq_tree::test::construct_pq_tree_pack ... ok
test executor::placement::test::custom_placement ... ok
test executor::placement::test::max_packer_numa_in_pkg ... ok
test executor::placement::test::max_packer_with_topology_0 ... ok
test executor::placement::test::max_packer_with_topology_empty ... ok
test executor::placement::test::max_spreader_with_topology_0 ... ok
test executor::placement::test::max_spreader_with_topology_1 ... ok
test executor::placement::test::max_spreader_with_topology_2 ... ok
test executor::placement::test::max_spreader_with_topology_empty ... ok
test channels::spsc_queue::tests::test_threaded ... ok
[...]

PR #479, opened 2021-12-11
Allow test to succeed if the shares reaches max

topecongiro:
### What does this PR do?

Fix a consistent failure of `deadline_queue_second_queued_item_increases_slope`.

A quick print debugging revealed that the value of `shares_first` is near 1000 (992 almost every time). Since the max value of shares is 1000, the assertion always fails.

This PR allows the test to succeed if the shares have reached to its maximum.
 
### Related issues

#477

### Additional Notes

I tried to minimize the value of `shares_first` by tweaking some parameters, but this attempt failed.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         uses: ./.github/actions/cache-setup
 
       - name: Install cargo-deadlinks
-        run: cargo install cargo-deadlinks
+        run: which deadlinks || cargo install cargo-deadlinks
 
       - name: Generate documentation
         run: cargo doc --all-features
--- a/glommio/src/controllers/deadline_queue.rs
+++ b/glommio/src/controllers/deadline_queue.rs
@@ -757,7 +757,7 @@ mod test {
             Timer::new(Duration::from_millis(2)).await;
             let shares_second = queue.queue.shares();
             // The second element that we push should rush the first.
-            assert!(shares_second >= shares_first * 20);
+            assert!(shares_second == 1000 || shares_second >= shares_first * 20);
             tq.await;
             tq2.await;
         });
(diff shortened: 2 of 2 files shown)


glommer:
It makes sense that in a very loaded system shares would go higher like that.
But it is likely a more robust condition to just test for `second >= first` instead of the 20x factor.

topecongiro:
Thank you for the review. I agree with your point. I have updated the test accordingly.

glommer (approved):

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         uses: ./.github/actions/cache-setup
 
       - name: Install cargo-deadlinks
-        run: cargo install cargo-deadlinks
+        run: which deadlinks || cargo install cargo-deadlinks
```
glommer, line 24: this seems unrelated to the work. Can you provided an updated branch without this commit?

Note that this only runs on CI, and CI has its own logic for caching, installing packages, etc.

topecongiro: I updated this line because the CI has [failed](https://github.com/DataDog/glommio/runs/4493424152?check_suite_focus=true) in the first attempt.

```
Run cargo install cargo-deadlinks
    Updating crates.io index
error: binary `cargo-deadlinks` already exists in destination
binary `deadlinks` already exists in destination
Add --force to overwrite
Error: Process completed with exit code 101.
```

Since the error didn't occur in other CIs, it shouldn't happen often, so simply rerunning the CI would have succeeded. At the same time, I didn't feel comfortable ignoring it, so I fixed it in this PR.

I have no objection to removing the commit from this PR and moving it to the other PR.

glommer: Let's remove it for now, if this gets done, best done in a separate PR. I am bringing this now to @HippoBaro 's attention.

glommer: I merged for now, because it is indeed failing without it.

PR #480, opened 2021-12-12
Support CPU topology with non-unique core ids

topecongiro:
### What does this PR do?

Support CPU topology with non-unique core ids.

The basic strategy is to sort CPUs by their (NUMA node id, core id) and assign virtual core id in this order. Note we need to ensure that the CPUs on the same core will have the same core id.

### Motivation

To make `cargo test` pass in my environment.

### Related issues

- #477

### Additional Notes

I first tried to fix the issue by calculating the virtual core id from the equation `core id + a number of cores in a NUMA node * NUMA node id`; this did not work for the following reasons:

- Core id is not consecutive within a single NUMA node (e.g., `0, 1, 2, 4, 5, 6` instead of `0, 1, 2, 3, 4, 5`)
- Each node may have different number of core

### Checklist

### DIFF
--- a/glommio/src/sys/hardware_topology.rs
+++ b/glommio/src/sys/hardware_topology.rs
@@ -93,17 +93,6 @@ pub fn get_machine_topology_unsorted() -> io::Result<Vec<CpuLocation>> {
         }
     }
 
-    // Make sure that differnt NUMA nodes do not contain the same core id.
-    let mut core_id_not_unique = false;
-    let mut core_to_numa = HashMap::new();
-    for l in &cpu_locations {
-        if *core_to_numa.entry(l.core).or_insert(l.numa_node) != l.numa_node {
-            core_id_not_unique = true;
-            break;
-        }
-    }
-
-    if core_id_not_unique {
         // Assign a virtual core id to each CPU. The basic strategy is to sort CPUs 
         // by their (NUMA node id, core id) and assign virtual core id in this order.
         // Note we need to ensure that the CPUs on the same core will have the same core id.
@@ -132,7 +121,6 @@ pub fn get_machine_topology_unsorted() -> io::Result<Vec<CpuLocation>> {
         for cpu_location in &mut cpu_locations {
             cpu_location.core = *cpu_to_vcore.get(&cpu_location.cpu).unwrap();
         }
-    }
 
     Ok(cpu_locations)
 }
--- a/glommio/src/sys/hardware_topology.rs
+++ b/glommio/src/sys/hardware_topology.rs
@@ -6,7 +6,7 @@
 //
 
 use std::{
-    collections::{HashMap, HashSet},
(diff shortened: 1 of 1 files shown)


glommer:
Hi. A small comment, see please that this fails the CI test:

```
thread 'tests::check_formating' panicked at 'cargo fmt failed. Note that glommio uses nightly for formatting, so please invoke cargo with +nightly', glommio/tests/linters.rs:15:9
```

`cargo test` locally should have generated the same output.

Your reasoning for the design you chose makes sense. I will take a deeper look at the code now

glommio/src/sys/hardware_topology.rs
```diff
@@ -93,6 +93,47 @@ pub fn get_machine_topology_unsorted() -> io::Result<Vec<CpuLocation>> {
...
+    let mut core_id_not_unique = false;
+    let mut core_to_numa = HashMap::new();
+    for l in &cpu_locations {
+        if *core_to_numa.entry(l.core).or_insert(l.numa_node) != l.numa_node {
+            core_id_not_unique = true;
+            break;
+        }
+    }
+
+    if core_id_not_unique {
```
glommer, line 106: this would work as well if the core id is unique, right?
To keep things simpler and easier to reason, I'd just invoke this all the time so we have the same code paths in all cases.

glommer (changes requested):

glommio/src/sys/hardware_topology.rs
topecongiro: Yes, it works.

topecongiro: Fixed in https://github.com/DataDog/glommio/pull/480/commits/4aac955cf464a4acee1703ea3ede2ff933f86397.

glommer:
There are a lot of CI failures here that look suspicious. @HippoBaro can you take a look ?

glommer:
This also has the issue with the cargo-deadlinks you pointed at.

My best guess is that we are not pinning the rust version, and it now started failing as CI picks up a new version.
I am okay with the PR as is right now, let me just hear from @HippoBaro our CI Lord about what's going on here.

HippoBaro:
> My best guess is that we are not pinning the rust version, and it now started failing as CI picks up a new version.

Sounds very plausible! I'm AFK until tomorrow, but I will take a look ASAP. In the meantime, if you are confident this is valid code, we can bypass CI and merge. @glommer let me know how that sounds.

HippoBaro:
#483 is now merged; @topecongiro, if you rebase on master CI should be happier. Sorry for the delay!

glommer (approved):

glommer:
@HippoBaro for cases like that we can do the update directly on github and if there are no conflicts merge right away. It's one button =) I just did that and Ci is running on it now

PR #483, opened 2021-12-23
update to rust-1.57 

HippoBaro:
Rust 1.57 adds new linter checks in Clippy, namely a better dead code
detector. It turns out we have quite a few dead variables around. I am
not sure if they are helpful or if we should remove them instead. I
prepended them with an underscore for now.

This PR also applies the linter suggestions from 1.59-nightly.

### DIFF
--- a/glommio/benches/sharding.rs
+++ b/glommio/benches/sharding.rs
@@ -22,7 +22,7 @@ fn main() {
 
     #[derive(Clone)]
     struct RequestHandler {
-        nr_shards: usize,
+        _nr_shards: usize,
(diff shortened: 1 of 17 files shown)


HippoBaro:
@glommer, let me know if we'd rather remove the unused symbols or if it makes sense to keep them around as I did here.

glommer (approved):

glommer:
I merged this so it unblocks other work. Some stuff we want to get rid of, but I'd have to comment on each individually. It's best if I do a separate pass.

PR #484, opened 2021-12-23
improve latency of latency-sensitive task queues performing IO

HippoBaro:
This fixes an issue whereby a task queue waiting for external events
(those coming from the reactor, like IO or timers) would not be woken up
 according to its latency requirement.

Consider the following setup: we have two task queues, `(A)` and `(B)`.
`(A)` is latency-sensitive with a preemption timer of 1ms and performs
IO. `(B)` doesn't have a latency requirement and therefore defaults to
the executor's timer: 100ms. `(B)` is CPU intensive but highly
cooperative.

And now the following sequence of events:
* `(A)` and `(B)` are both `READY`; therefore, the preemption timer is
  set to the minimum of the two, 1ms;
* `(A)` runs and schedules all the IO it can. The task returns
  `Poll::Pending`;
* `(B)` runs and eventually gets preempted after ~1ms;
* The reactor enters the rings and posts the IO from `(A)`;
* `(A)` is `PENDING` and `(B)` is `READY,` so the preemption timer is
  set to `(B)`: 100ms;
* `(B)` runs for 100ms, until it is preempted;
* The reactor enters the rings and collects the IO results, waking up
  `(A)` in the process;
* `(A)` and `(B)` are both `READY`; therefore, the preemption timer is
  set to the minimum of the two, 1ms;
* `(A)` collects the result of the IOs
* ...

This isn't right: `(A)` wants to be woken up within 1ms of being marked
`READY`! To fix this issue durably, all IO from latency-sensitive queues
are now posted on the latency ring. This means that as soon as the
kernel creates a CQE on the latency ring, tasks get preempted by the
next call to `yield_if_needed.`

This improves the latency of these IO operations significantly. Right
now, this means the poll ring is only used when IO is submitted from
a throughput task queue. We may try to improve upon this later on.

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -1200,8 +1200,7 @@ impl LocalExecutor {
                 // requests that are latency sensitive we want them out of the
                 // ring ASAP (before we run the task queues). We will also use
                 // the opportunity to install the timer.
-                let duration = self.preempt_timer_duration();
-                self.parker.poll_io(duration);
+                self.parker.poll_io(|| Some(self.preempt_timer_duration()));
                 let run = self.run_task_queues();
                 let cur_time = Instant::now();
                 self.queues.borrow_mut().stats.total_runtime += cur_time - pre_time;
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -1189,6 +1189,8 @@ impl Reactor {
         let main_ring =
             SleepableRing::new(ring_depth, "main", allocator.clone(), source_map.clone())?;
         let poll_ring = PollRing::new(ring_depth, allocator.clone(), source_map.clone())?;
+        let mut latency_ring =
+            SleepableRing::new(ring_depth, "latency", allocator.clone(), source_map.clone())?;
 
         match main_ring.registrar().register_buffers_by_ref(&registry) {
             Err(x) => warn!(
@@ -1204,13 +1206,23 @@ impl Reactor {
                     main_ring.registrar().unregister_buffers().unwrap();
                 }
                 Ok(_) => {
-                    allocator.activate_registered_buffers(0);
+                    match latency_ring.registrar().register_buffers_by_ref(&registry) {
+                        Err(x) => {
+                            warn!(
+                                "Error: registering buffers in the poll ring. Skipping{:#?}",
+                                x
+                            );
+                            poll_ring.registrar().unregister_buffers().unwrap();
+                            main_ring.registrar().unregister_buffers().unwrap();
+                        }
+                        Ok(_) => {
+                            allocator.activate_registered_buffers(0);
+                        }
+                    };
                 }
             },
         }
 
-        let mut latency_ring =
-            SleepableRing::new(ring_depth, "latency", allocator.clone(), source_map.clone())?;
         let link_fd = latency_ring.ring_fd();
 
         let eventfd_src = Source::new(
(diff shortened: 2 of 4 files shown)


HippoBaro:
I wrote a little program that demonstrates the bug; I'll clean up my stash and share it later

glommer:
Hippo, first of all, great work investigating this. PI issues are the hardest, and that's a big improvement.
I don't like the fact that we're not posting to the poll ring when a task has latency requirements. It is arguably even more important to post latency-bound tasks into the poll ring.

I understand why you do it, though: if we post to the poll ring, now we have to wait 100ms because `A` is `Pending`.
But that just means that you need to keep the preempt timer at 1ms if `A` has tasks that are in-flight. 

Take a look at how we prevent the reactor from sleeping if there are tasks in the poll ring. It's a similar idea. A useful mental model is that a task that has to be polled still needs care an attention until it is fully ready, dispatching is not enough.

I will now go through the code for specific comments

glommio/src/sys/uring.rs
```diff
@@ -1568,14 +1569,19 @@ impl Reactor {
         let mut main_ring = self.main_ring.borrow_mut();
         let mut lat_ring = self.latency_ring.borrow_mut();
 
+        // we need to make sure we consume the latency ring before setting the
+        // preemption timer because consuming this ring may wake up some previously
+        // asleep task queues.
+        consume_rings!(into &mut woke; lat_ring);
```
glommer, line 1584: commenting here because github doesn't allow to comment on the patch description:

`consume_rings` doesn't (usually) issue a syscall. syscalls are needed to submit new events, but the path to fetch results is syscall-free

glommio/src/executor/mod.rs
```diff
@@ -1077,6 +1077,10 @@ impl LocalExecutor {
             } else {
                 ran = true;
             }
+
+            if self.reactor.sys.needs_expedited_kernel_enter() {
```
glommer, line 1081: I would prefer we don't include this work.
the reason is that if you have a workload that is comprised of a latency-sensitve queue with a lot of tasks each submitting ping-pong like I/O you will go to the reactor all the time and end up hurting your latency.

The task queue should be allowed to submit until `need_preempt`

glommer (changes requested):
I'd prefer we skip the last commit. The rest looks great.

HippoBaro:
@glommer I took a stab at implementing the suggested fix in #486 

I created a separate patch-set because, after testing, some of my concerns about the approach materialized. Specifically that when competing against other tasks, submitting on the latency ring yields better results. I wrote a benchmark in #487 that gives the following results:

This PR:
```
 --- iteration: 0 ---
performed 1000000 read IO at 278305 IOPS (took 3.59s)
[measured] lat (end-to-end):	min:  170us	p50:  465us	p99:  639us	p99.9:  693us	p99.99: 1884us	p99.999: 2234us	 max: 2235us
```

#486:

```
 --- iteration: 0 ---
performed 1000000 read IO at 57930 IOPS (took 17.26s)
[measured] lat (end-to-end):	min: 2137us	p50: 2205us	p99: 4354us	p99.9: 4376us	p99.99: 5210us	p99.999: 6340us	 max: 8299us
```

Your solution does improve things substantially, but my approach seems to perform best. Note that the tests are done with a preemption timer of 1ms for the IO queue. Reducing that marginally improves results, but anything below 500us hurts performance.

I think that the poll ring will work best in uncontended workloads, but we need some heuristic to decide at runtime what ring to post on. Another avenue I'm looking into is to consume the rings more often. 

For reference, this is HEAD (I just did 10k IO otherwise it takes forever):
```
 --- iteration: 0 ---
performed 10000 read IO at 657 IOPS (took 15.20s)
[measured] lat (end-to-end):	min: 101180us	p50: 201175us	p99: 402287us	p99.9: 402751us	p99.99: 402815us	p99.999: 402815us	 max: 402815us
```

glommer:
Okay, let's merge this. I see you removed the patch that I found problematic, so there should be no more concerns.

glommer (approved):

PR #485, opened 2021-12-23
update doc for previously blocking synchronous operations

HippoBaro:
`io_uring` doesn't support all the regular file operations yet. In the
past, we were filling the gaps by issuing blocking syscalls directly,
blocking the reactor thread. Because this kind of sucks, we added a
warning in the docs to warn application writers.

Nowadays, we perform these blocking operations in a background thread.
Therefore, we no longer block the reactor thread, and the warning is no
longer relevant.

### DIFF
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -199,14 +199,16 @@ impl BufferedFile {
 
     /// Truncates a file to the specified size.
     ///
-    /// **Warning:** synchronous operation, will block the reactor
+    /// Note: this syscall might be issued in a background thread depending on
+    /// the system's capabilities.
     pub async fn truncate(&self, size: u64) -> Result<()> {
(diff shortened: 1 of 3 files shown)


bryandmc:
Should we at least mention that they aren't going over io_uring, at least? I dunno might be worth pointing out at least, especially because some of them are, or will, be available on io_uring and we don't want people confused by that.

HippoBaro:
> Should we at least mention that they aren't going over io_uring, at least? I dunno might be worth pointing out at least, especially because some of them are, or will, be available on io_uring and we don't want people confused by that.

Alright, I'll rework this and add a note about the background thread 🙇

glommer:
I am okay with the note, but don't say it goes to a background thread, but rather than it "might".
io_uring in the latest kernel even have some of those operations already and eventually we'll want them in. It just wasn't a priority so far.

PR #486, opened 2021-12-26
[alt] improve latency of latency-sensitive task queues performing IO

HippoBaro:
This is an alternative implementation of #484 as per @glommer insights.

### DIFF
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -18,6 +18,7 @@ use std::{
     fmt,
     io,
     io::{Error, ErrorKind},
+    num::NonZeroI64,
     os::unix::io::RawFd,
     panic,
     ptr,
@@ -626,6 +627,8 @@ trait UringCommon {
     fn consume_one_event(&mut self) -> Option<bool>;
     fn name(&self) -> &'static str;
     fn registrar(&self) -> iou::Registrar<'_>;
+    fn register_latency_requirement(&mut self, req: Latency);
+    fn latency_requirement(&self) -> Option<Duration>;
 
     fn consume_sqe_queue(
         &mut self,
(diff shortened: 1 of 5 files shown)


glommer:
It would be helpful to know if your results were different using the alternative version, and if yes better, worse, by how much, etc

HippoBaro:
> It would be helpful to know if your results were different using the alternative version, and if yes better, worse, by how much, etc

I posted this on the other PR where the history of the issue is: https://github.com/DataDog/glommio/pull/484#issuecomment-1001272157

glommio/src/sys/uring.rs
```diff
@@ -1042,6 +1082,10 @@ impl UringCommon for SleepableRing {
                 _ => None,
             },
             |mut src, res| {
+                self.lat_req
+                    .entry(src.io_requirements.latency_req)
```
glommer, line 1086: I have a really bad feeling about this.
There are situations like linked sqe, cancellations, etc, that may create a mismatch here.
It is potentially safe but quite fragile, and at least we should have low level unit tests that test those cases specifically.

This is very tricky, because if this counter has a mismatch, it won't present itself as a bug, but as a very weird performance issue.

My idea of implementing this was to do this at a higher (executor) layer, and not at the uring layer. If requests are in-flight the task will be `Pending`, and we would include the presence of pending tasks when calculating the preempt timer.

HippoBaro: I agree. I felt the same when I wrote this. 

Doing this at the executor level means we would be including requests going through the latency ring. We would instead not influence the preemption timer since the latency ring already does that for us. 
But, because what goes in what ring is an implementation detail of the reactor, I felt like this stuff belonged there instead.

Now that this code, I like the idea of keeping two kinds of latency requirements: (1) the scheduler latency requirement and (2) the reactor latency requirement. I feel those two things are somewhat orthogonal. 

Your point is very much valid, and if we decide to go the route of this PR, this will require a rework.

bryandmc:
Is this patch still in play?

HippoBaro:
> Is this patch still in play?

A previous patch set disabled the poll ring for latency queues (#484). It would be great to find a way to enable it back in the future. I think this PR could be the basis of that. I don't have the bandwidth to investigate this myself for now (+ a realistic production environment would use a RAID device, and they don't support polling yet). 

I'll close this and open an issue instead. Thanks!

PR #487, opened 2021-12-27
add a benchmark that measures how IO performs when competing with CPU

HippoBaro:
This adds a new benchmark that creates a CPU intensive, throughput
oriented task queue alongside a latency-sensitive IO queue.

### DIFF
--- a/glommio/benches/competing_io.rs
+++ b/glommio/benches/competing_io.rs
@@ -1,0 +1,147 @@
+use futures::future::join;
+use futures_lite::{stream, AsyncWriteExt, StreamExt};
+use glommio::{
+    enclose,
+    io::{ImmutableFile, ImmutableFileBuilder},
+    Latency,
+    LocalExecutorBuilder,
+    Placement,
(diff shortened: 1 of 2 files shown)


glommio/benches/competing_io.rs
```diff
@@ -0,0 +1,146 @@
...
+                        Shares::Static(1000),
+                        Latency::Matters(Duration::from_millis(1)),
+                        "lat",
+                    );
+                    let throughput_tq = glommio::executor().create_task_queue(
+                        Shares::Static(1000),
+                        Latency::NotImportant,
+                        "throughput",
+                    );
+                    let gate = Rc::new(RefCell::new(true));
```
glommer, line 57: given the name, could you use https://docs.rs/glommio/0.6.0/glommio/sync/struct.Gate.html ?
If not, at least use a `Cell` instead of a `RefCell`

glommer (approved):

glommer (approved):

PR #489, opened 2021-12-27
probe `sysfs` to detect the minimum alignment io size

HippoBaro:
From https://people.redhat.com/msnitzer/docs/io-limits.txt:
> Storage vendors can also supply "I/O hints" about a device's preferred
minimum unit for random I/O ('minimum_io_size') and streaming I/O
('optimal_io_size'). For example, these hints may correspond to a RAID
device's chunk size and stripe size respectively.

### DIFF
--- a/glommio/src/io/dma_file.rs
+++ b/glommio/src/io/dma_file.rs
@@ -158,7 +158,9 @@ impl DmaFile {
 
         Ok(DmaFile {
             file,
-            o_direct_alignment: sysfs::BlockDevice::minimum_io_size(major, minor) as u64,
+            o_direct_alignment: (sysfs::BlockDevice::minimum_io_size(major, minor) as u64)
+                .max(sysfs::BlockDevice::logical_block_size(major, minor) as u64)
+                .max(512), // make sure the alignment is at least 512 in any case
             max_sectors_size,
             max_segment_size,
             pollable,
--- a/glommio/src/io/dma_file.rs
+++ b/glommio/src/io/dma_file.rs
@@ -72,7 +72,6 @@ pub struct DmaFile {
 }
 
 impl DmaFile {
-    // FIXME: Don't assume 512, we can read this info from sysfs
     /// align a value up to the minimum alignment needed to access this file
     pub fn align_up(&self, v: u64) -> u64 {
         align_up(v, self.o_direct_alignment)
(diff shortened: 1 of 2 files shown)


HippoBaro:
From the wording of the documentation, I feel like it's okay for the storage devices to leave `minimum_io_size` undefined (it would then be zero). Maybe we should fall back on `physical_block_size` when that happens?

HippoBaro:
From [man open(2)](https://man7.org/linux/man-pages/man2/open.2.html#NOTES):
```
Under Linux 2.4, transfer sizes, the alignment of the user
buffer, and the file offset must all be multiples of the logical
block size of the filesystem.  Since Linux 2.6.0, alignment to
the logical block size of the underlying storage (typically 512
bytes) suffices. 
```

I'll make the code fallback on `logical_block_size` :+1:

glommio/src/io/dma_file.rs
```diff
@@ -163,11 +163,7 @@ impl DmaFile {
             | (opts.custom_flags as libc::c_int & !libc::O_ACCMODE);
 
         let res = DmaFile::open_at(dir, path, flags, opts.mode).await;
-        let mut f = enhanced_try!(res, opdesc, Some(path), None)?;
-        // FIXME: Don't assume 512 or 4096, we can read this info from sysfs
-        // currently, we just use the minimal {values which make sense}
-        f.o_direct_alignment = if opts.write { 4096 } else { 512 };
```
HippoBaro, line 169: @glommer this PR removes this in favor of always using the alignment from sysfs. It will be `512` in most cases. I don't remember why we used `4096` when the write bit is set. Was it for compatibility?

glommer: 512 writes are expensive even if allowed. The logical block is not 512-bytes, so you have to RMW internally.

glommer: Also, my memory is failing me a bit here, but I don't think the OS will like a sub-4096 write even if the media allows it.

I'd like to be defensive here and keep writes at a minimum of 4k.

HippoBaro: Ok. One thing that bothers me is that we have one alignment setting per file so if a file is opened with the write bit, then reads will be aligned on 4k too. We could change that though.

As for write alignment, there are two write patterns:
* Most people use linear writes and use large buffers, much larger than the minimum. Does it make a difference if a 128k buffer is aligned on 512 or 4k? When you talk about alignment I feel you compound that with the minimum IO buffer size. Can we make 4k buffers the minimum with a 512b alignment if that's what sysfs says? 
* A few people will use random writes and for them, it makes sense to expose the very minimum alignment the system supports IMO. 

What do you think?

HippoBaro: > 512 writes are expensive even if allowed. The logical block is not 512-bytes, so you have to RMW internally.

Even if sysfs puts `512` in `logical_block_size`? That sounds wild

glommer: I don't think the random writes case is very interesting at all. It seldom shows up.

davidblewett:
This PR fixes a bug we encountered where the 512 byte read from `ImmutableFile` throws errors due to the block size being 4096.

glommer:
Even I always forget the rules here =p but that doesn't add up. I thought we could always read 512-bytes from a file, despite of the block size. memory buffers must be 4k-aligned, though. Are you sure this was not the issue ?

PR #490, opened 2021-12-29
fix preemption timer when using `rush_dispatch()`

HippoBaro:
The idea behind `rush_dispatch` is great. Sometimes, it is advantageous 
to rush to the rings and performs some io_uring transactions
immediately.

Our implementation suffers from one deadly bug, however. What if, while
consuming the CQEs, we get the preemption timer back? Uh oh! If that
happens, we silently throw it away. And because the pointers to the ring
have now changed (and are equal), `need_preempt` returns false. Uh oh!!
Even worse, we now don't have a preemption timer, so we will run every
task in the system to completion (or until `Poll::Pending`) before going
to the reactor once again.

The fix is to check whether the timer has returned after rushing to the
rings. We also add a new variant of `need_preempt(),`
`need_preempt_check_timer()` that explicitly checks the timer too. This
more expensive check is only used in-between scheduling two queues to
avoid undue overhead.

Therefore, the control flow goes like this:
* `rush_dispatch` is called, and the timer is collected during the
  flush;
* `rush_dispatch` detects that the timer has returned and marks the task
  queue for yield;
* the scheduler calls `need_preempt_check_timer()` before scheduling the
 next queue and short-circuits back to the reactor;
* a new preemption timer is armed, and we go on our merry way.

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -1065,8 +1065,9 @@ impl LocalExecutor {
 
     fn mark_me_for_yield(&self) {
         let queues = self.queues.borrow();
-        let mut me = queues.active_executing.as_ref().unwrap().borrow_mut();
-        me.yielded = true;
+        if let Some(me) = queues.active_executing.as_ref() {
+            me.borrow_mut().yielded = true;
+        }
     }
 
     fn spawn<T>(&self, future: impl Future<Output = T>) -> multitask::Task<T> {
@@ -1110,11 +1111,21 @@ impl LocalExecutor {
         self.reactor.need_preempt()
     }
 
+    pub(crate) fn need_preempt_check_timer(&self) -> bool {
+        let timer = if let Some(src) = &*self.reactor.sys.timeout_source() {
+            src.result().is_some()
+        } else {
+            false
+        };
+
+        timer || self.need_preempt()
+    }
+
     fn run_task_queues(&self) -> bool {
         let mut ran = false;
         loop {
             self.reactor.sys.install_eventfd();
-            if self.need_preempt() {
+            if self.need_preempt_check_timer() {
                 break;
             }
             if !self.run_one_task_queue() {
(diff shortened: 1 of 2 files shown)


glommio/src/sys/uring.rs
```diff
@@ -1512,10 +1513,24 @@ impl Reactor {
...
             None => Self::simple_poll(&self.poll_ring, woke),
             Some(Latency::NotImportant) => Self::simple_poll(&self.main_ring, woke),
-            Some(Latency::Matters(_)) => Self::simple_poll(&self.latency_ring, woke),
+            Some(Latency::Matters(_)) => {
+                let ret = Self::simple_poll(&self.latency_ring, woke);
+                if let Some(src) = &*self.timeout_src.borrow() {
+                    if src.result().is_some() {
+                        // the preemption timer has fired during the last poll
+                        // we need to communicate this to the scheduler
+                        let _ = executor().yield_task_queue_now();
```
glommer, line 1522: I don't understand why this is being tested only in the `Latency::Matters` case. 
Even if you don't have a latency requirement, other task queues might, so if the timer fired you have to bail.

am I missing anything ?

glommio/src/executor/mod.rs
```diff
@@ -1110,11 +1111,21 @@ impl LocalExecutor {
         self.reactor.need_preempt()
     }
 
+    pub(crate) fn need_preempt_check_timer(&self) -> bool {
```
glommer, line 1114: I am not a fan of this, but I have to think a bit more about the best way to tackle it.

Is the reason you need it the fact that by consuming the cqe in the latency ring, need preempt is now zero ?

One of the reasons I don't like it, besides the special casing, is that `need_preempt()` is supposed to be about more than timers. But rather, any event in the latency ring.

Also the fact that this works "only between task queues" is an artifact of the current implementation. Up until very recently rush_dispatch could have been called inside a task queue (thirstycrow temporarily removed it)

glommer (changes requested):

glommio/src/sys/uring.rs
HippoBaro: Yeah, the preemption timer is added to the latency ring so if we rush to any other ring, that's not an issue.

glommio/src/executor/mod.rs
HippoBaro: > Is the reason you need it the fact that by consuming the cqe in the latency ring, need preempt is now zero ?

Yes. `need_preempt` works under the assumption that a timer is always scheduled on the latency ring on the way out of the reactor. If it isn't there (because we polled the ring from a task, for instance) then the preemption mechanism breaks. 

I added this variant to be more robust and cover that use case.

> Also the fact that this works "only between task queues"

It works from everywhere. I use it between queues because that's enough in practice.

>  Up until very recently rush_dispatch could have been called inside a task queue (thirstycrow temporarily removed it)

Are you sure? From what I can see `rush_dispatch` is being used in the net code

PR #493, opened 2022-01-06
fast IO cancellation

HippoBaro:
So far, to cancel a source in glommio, the code would:
* Probe the submitting queue of the associated ring to find its
  corresponding `UringOpDescriptor`;
* If one is found, it is removed;
* Otherwise, a cancellation request is sent on the ring.

The issue with this way of handling cancellation is that the first step
is a linear probing over a potentially large number of descriptors.
Specifically, it means that canceling a source is _O(n)_ where _n_ is
the total number of queued sources in the ring.

Therefore, a realistic application canceling a query (that issues
many IOs) is an _O(kn)_ operation where _k_ is the number of sources
created by the request. Uh oh. That's a lot of work done in a destructor
where we can't cooperate and yield.

Instead, this commit changes the behavior of source cancellation to be
lazy. When dropped, a source marks itself as "canceled" if it wasn't
submitted yet. Meanwhile, the reactor now filters out canceled sources.
If the source was dispatched, we submit a cancel request like before.

Co-authored with @glommer

### DIFF
--- a/glommio/src/iou/sqe.rs
+++ b/glommio/src/iou/sqe.rs
@@ -680,6 +680,12 @@ impl<'ring> Iterator for SQEs<'ring> {
     }
 }
 
+impl<'ring> Drop for SQEs<'ring> {
+    fn drop(&mut self) {
+        self.sq.sqe_tail -= self.remaining()
+    }
+}
+
 /// An Iterator of [`SQE`]s which will be hard linked together.
 pub struct HardLinked<'ring, 'a> {
     sqes: &'a mut SQEs<'ring>,
--- a/glommio/src/sys/source.rs
+++ b/glommio/src/sys/source.rs
@@ -31,6 +31,7 @@ use std::{
     mem::MaybeUninit,
     os::unix::io::RawFd,
     path::PathBuf,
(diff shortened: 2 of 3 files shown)


duarten (approved):
Nice!

glommer (approved):

glommer:
Code is good, but please fix my email prior to merging

HippoBaro:
It is worth mentioning that the current implementation is not optimal. An optimal implementation would use an intrusive linked list to store the `UringOpDescriptor`s such that `drop` of `Source` could remove itself from the list in _O(1)_. This would simplify the dispatching code in the reactor because having to filter out canceled sources is complicated. 

Intrusive data structures are a real pain in rust, so I went the easy way.

glommer:
Intrusive is easy.

The issue here is not lack of intrusive, but that today, `UringOpDescriptor` doesn't allocate and it would be forced to do it if we were to add an intrusive element there.

The `InnerSource` can hold the link, but we don't dispatch from the source, and there are actual dispatches without a source (for example in destructors).

So it's all possible, but to do it efficiently it's a bit tricky.

PR #496, opened 2022-01-07
measure the pre-reactor IO latency

HippoBaro:
We now record three kinds of source latencies:
* pre-reactor: the time the source spent queued (this commit);
* io: the time the source spent in the kernel;
* post-reactor: the time the scheduler took to get back at the consuming
  task.

Previously, the time the source was queued counted towards IO
latency (second item in the list).

### DIFF
--- a/glommio/src/sys/mod.rs
+++ b/glommio/src/sys/mod.rs
@@ -451,14 +451,30 @@ impl From<Duration> for TimeSpec64 {
     }
 }
 
+pub(super) struct Latencies {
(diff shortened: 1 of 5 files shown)


glommio/src/sys/mod.rs
```diff
@@ -487,6 +492,18 @@ impl Wakers {
             true
         }
     }
+
+    fn timestamps(&mut self) -> Option<(Instant, Instant, Instant)> {
```
davidblewett, line 496: This is kind of unwieldy. Could we make it a struct?

PR #497, opened 2022-01-08
better `read_many` interface and defaults 

HippoBaro:
* Provide a better, less confusing interface using named enums. This
  makes the code a lot more readable;
* Read sysfs to give good device-specific options for request merging 
  (if we merge too much, we may end up with a request so large the kernel has 
  to split it underneath our feet, which defeats the purpose).

### DIFF
--- a/glommio/src/io/bulk_io.rs
+++ b/glommio/src/io/bulk_io.rs
@@ -14,6 +14,59 @@ use std::{
     rc::Rc,
 };
 
+/// Set a limit to the size of merged IO requests.
+#[derive(Debug)]
+pub enum MergedBufferLimit {
+    /// Disables request coalescing
+    NoMerging,
+
+    /// Sets the limit to the maximum the kernel allows for the underlying
+    /// device without breaking down the request into smaller ones
+    /// (/sys/block/.../queue/max_sectors_kb)
+    DeviceMaxSingleRequest,
+
+    /// Sets a custom limit.
+    /// This value should be a multiple of the file's alignment. If it is not,
+    /// it'll be aligned down.
+    Custom(usize),
+}
+
+impl Default for MergedBufferLimit {
+    fn default() -> Self {
+        Self::DeviceMaxSingleRequest
+    }
+}
+
+/// Set a limit to the amount of read amplification in-between two mergeable IO
+/// requests.
+#[derive(Debug)]
+pub enum ReadAmplificationLimit {
+    /// Deny read amplification.
+    ///
+    /// Note that, because IO request coalescing is done _before_ buffers are
+    /// aligned, requests may be merged if they are smaller than the minimum IO
+    /// size. For instance, if the minimum IO size is 4KiB and the user reads
+    /// [0..256] and [2048..2560] then the two will be merged into [0..4096] to
+    /// accommodate the 4KiB minimum IO size.
+    NoAmplification,
+
+    /// Merge two consecutive IO requests if the read amplification is below a
+    /// limit.
+    ///
+    /// This value doesn't have any alignment constrain.
+    Custom(usize),
+
+    /// Always merge successive IO requests if possible, no matter the distance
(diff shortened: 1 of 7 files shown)


glommio/src/sys/sysfs.rs
```diff
@@ -145,6 +153,14 @@ impl BlockDevice {
         block_property!(DEV_MAP, physical_block_size, major, minor)
     }
 
+    pub(crate) fn max_sectors_size(major: usize, minor: usize) -> usize {
+        block_property!(DEV_MAP, max_sectors_size, major, minor)
+    }
+
+    pub(crate) fn max_segment_size(major: usize, minor: usize) -> usize {
+        block_property!(DEV_MAP, max_sectors_size, major, minor)
```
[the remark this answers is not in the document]
HippoBaro: Oh, thanks! That would have been unfortunate.

glommio/src/io/bulk_io.rs
```diff
@@ -14,6 +14,51 @@ use std::{
     rc::Rc,
 };
 
+/// Set a limit to the size of merged IO requests.
+#[derive(Debug)]
```
glommer, line 18: we should probably implement Default for this. I think `DeviceMaxSingleRequest` is a good default.

glommio/src/io/bulk_io.rs
```diff
@@ -14,6 +14,51 @@ use std::{
...
+
+    /// Sets a custom limit.
+    /// This value should be a multiple of the file's alignment. If it is not,
+    /// it'll be aligned down.
+    Custom(usize),
+}
+
+/// Set a limit to the amount of read amplification in-between two mergeable IO
+/// requests.
+#[derive(Debug)]
```
glommer, line 40: same here, implement Default, and `NoAmplification` is a good default

glommio/src/io/bulk_io.rs
```diff
@@ -14,6 +14,51 @@ use std::{
...
+    NoMerging,
+
+    /// Sets the limit to the maximum the kernel allows for the underlying
+    /// device without breaking down the request into smaller ones
+    /// (/sys/block/.../queue/max_sectors_kb)
+    DeviceMaxSingleRequest,
+
+    /// Sets the limit to the maximum the kernel allows for the underlying
+    /// device (/sys/block/.../queue/max_segment_size)
+    DeviceMax,
```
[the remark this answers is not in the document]
HippoBaro: The old behavior had "unlimited" but when I discovered that the kernel actually has a hard limit I introduced this instead to prevent the stream from failing.  Esp with in-memory files, I think such fat requests like this _could_ happen. 

PS: It is the same limit on my side and on an i3.

glommio/src/io/bulk_io.rs
HippoBaro: I can't implement `Default` because the value of `DeviceMaxSingleRequest` depends on the major/minor device ids

glommer: what do you mean? 

The actual value changes but all you're doing in this enum is to pick the variant. You don't need to fill in the value.

glommio/src/io/bulk_io.rs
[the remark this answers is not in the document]
glommer: I am not sure what this buys us. It's a property we don't fully understand, that will essentially be unlimited until it is zero in someone's system for whatever reason you didn't anticipate.

glommio/src/io/bulk_io.rs
HippoBaro: Indeed 🤦‍♂️

glommio/src/io/bulk_io.rs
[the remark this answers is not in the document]
HippoBaro: Do you think I should remove the option and limit values we are given under the hood? Or just not use `max_segment_size`?

PR #498, opened 2022-01-10
fix racy unit test

HippoBaro:
This test keeps borrows across await points, which makes it flaky.


bryandmc (approved):
yeah this makes sense.. looks good

PR #499, opened 2022-01-13
add a stall detector that logs stacktraces of unyielding tasks, redux

davidblewett:
### What does this PR do?

This PR supercedes https://github.com/DataDog/glommio/pull/417 . I believe all the concerns raised there are addressed in this implementation.

This PR adds a rudimentary stall detection mechanism to Glommio.

It measures the runtime of task queues and logs if they don't yield when
they are due.

Knowing if a queue is stalling the reactor is nice, but finding the
exact code location where the stall occurs remains very hard. Especially
if a task queue hosts many concurrent fibers.

To help with that, this PR introduces a stall detection mechanism that
records stack traces of stalling tasks. It works as follows:

- When a task queue is scheduled for execution, we set up a timer that
triggers some time after the queue is expected to yield (we add a 10%
error margin to avoid false positives). A background thread collocated
with the local executor waits on the timer at all times.
- When the timer fires, the thread sends a signal (SIGUSR1) to the
local executor thread. Upon receiving the signal, the local executor
records a complete trace of the local stack. Here we take advantage of
the fact that by default, the kernel invokes signal handlers on top of
the existing stack. i.e. the frames we record are those of the
problematic user code that was meant to yield. The recorded frames are
pushed on a non-blocking communication channel that links the signal
handler and the local executor.
- When a task queue yields, the local executor disarm the timer and
checks the communication channel for potential recorded frames, if there
are any then we can conclude that the queue stalled, so we log them.

This code works in practice but has two major drawbacks:

- The timer dance is expensive; expect a high number of syscalls.
Because of this runtime overhead, the stall detector is disabled by
default. To opt-in, you  must pass in something that implements
`StallDetectionHandler` when constructing the `LocalExecutor`. There
is a concrete `DefaultStallDetectionHandler` that uses `SIGUSR1` and 
logs a warning log message.
- We log stalls only after the queue yield. Therefore, if there is a bug
in your code and your queue never yields, the stall detector will never
log the code location that's at fault (even though we probably have
recorded the trace by then). The reason for this is that logging from a
signal handler is illegal.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+too-many-arguments-threshold = 8
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,0 @@
-too-many-arguments-threshold = 8
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -17,6 +17,7 @@ readme = "../README.md"
 
 [dependencies]
 ahash = "~0.7"
+backtrace = { version = "0.3.61", optional = true }
 bitflags = "1.3"
 bitmaps = "3.1.0"
 buddy-alloc = "~0.4"
@@ -35,6 +36,7 @@ pin-project-lite = "~0.2"
 rlimit = "~0.6"
 scoped-tls = "1.0.0"
 scopeguard = "1.1"
+signal-hook = { version = "0.3.10", optional = true }
 sketches-ddsketch = "0.1"
 smallvec = { version = "1.7", features = ["union"] }
 socket2 = { version = "~0.3", features = ["unix", "reuseport"] }
@@ -45,6 +47,7 @@ typenum = "1.14"
 fastrand = "1.5"
 futures = "~0.3"
 hdrhistogram = "7.3.0"
+logtest = "2.0.0"
 pretty_env_logger = "~0.4"
 rand = "~0.8"
 tokio = { version = "1.12", default-features = false, features = ["rt", "macros", "rt-multi-thread", "net", "io-util", "time", "sync"] }
@@ -56,7 +59,7 @@ cc = "1.0"
 [features]
 bench = []
 debugging = []
-stall-detection = []
+stall-detection = ["signal-hook", "backtrace"]
 
 [[bench]]
 name = "executor"
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -16,41 +16,41 @@ categories = ["asynchronous", "concurrency", "os", "filesystem", "network-progra
 readme = "../README.md"
 
 [dependencies]
-ahash = "~0.7"
-backtrace = { version = "0.3.61" }
+ahash = "0.7"
+backtrace = { version = "~0.3.58" }
 bitflags = "1.3"
-bitmaps = "3.1.0"
-buddy-alloc = "~0.4"
+bitmaps = "3.1"
+buddy-alloc = "0.4"
 concurrent-queue = "1.2"
-crossbeam = "~0.8"
+crossbeam = "0.8"
 enclose = "1.1"
 futures-lite = "1.12"
-intrusive-collections = "~0.9"
+intrusive-collections = "0.9"
 lazy_static = "1.4"
-libc = "~0.2"
-lockfree = "~0.5"
-log = "~0.4"
-membarrier = "~0.2"
-nix = "~0.23.0"
-pin-project-lite = "~0.2"
-rlimit = "~0.6"
-scoped-tls = "1.0.0"
+libc = "0.2"
+lockfree = "0.5"
+log = "0.4"
+membarrier = "0.2"
+nix = "0.23"
+pin-project-lite = "0.2"
+rlimit = "0.6"
+scoped-tls = "1.0"
 scopeguard = "1.1"
 signal-hook = { version = "0.3" }
 sketches-ddsketch = "0.1"
 smallvec = { version = "1.7", features = ["union"] }
-socket2 = { version = "~0.3", features = ["unix", "reuseport"] }
-tracing = "~0.1"
-typenum = "1.14"
+socket2 = { version = "0.3", features = ["unix", "reuseport"] }
+tracing = "0.1"
+typenum = "1.15"
 
 [dev-dependencies]
-fastrand = "1.5"
-futures = "~0.3"
-hdrhistogram = "7.3.0"
-pretty_env_logger = "~0.4"
-rand = "~0.8"
-tokio = { version = "1.12", default-features = false, features = ["rt", "macros", "rt-multi-thread", "net", "io-util", "time", "sync"] }
-tracing-subscriber = { version = "~0.3", features = ["env-filter"] }
+fastrand = "1"
+futures = "0"
+hdrhistogram = "7"
+pretty_env_logger = "0"
+rand = "0"
+tokio = { version = "1", default-features = false, features = ["rt", "macros", "rt-multi-thread", "net", "io-util", "time", "sync"] }
+tracing-subscriber = { version = "0", features = ["env-filter"] }
 
 [build-dependencies]
 cc = "1.0"
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -36,7 +36,7 @@ pin-project-lite = "~0.2"
 rlimit = "~0.6"
 scoped-tls = "1.0.0"
 scopeguard = "1.1"
-signal-hook = { version = "0.3.10" }
+signal-hook = { version = "0.3" }
 sketches-ddsketch = "0.1"
 smallvec = { version = "1.7", features = ["union"] }
 socket2 = { version = "~0.3", features = ["unix", "reuseport"] }
@@ -47,7 +47,6 @@ typenum = "1.14"
 fastrand = "1.5"
 futures = "~0.3"
 hdrhistogram = "7.3.0"
-logtest = "2.0.0"
 pretty_env_logger = "~0.4"
 rand = "~0.8"
 tokio = { version = "1.12", default-features = false, features = ["rt", "macros", "rt-multi-thread", "net", "io-util", "time", "sync"] }
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -17,7 +17,7 @@ readme = "../README.md"
 
 [dependencies]
 ahash = "~0.7"
-backtrace = { version = "0.3.61", optional = true }
+backtrace = { version = "0.3.61" }
 bitflags = "1.3"
 bitmaps = "3.1.0"
 buddy-alloc = "~0.4"
@@ -36,7 +36,7 @@ pin-project-lite = "~0.2"
 rlimit = "~0.6"
 scoped-tls = "1.0.0"
 scopeguard = "1.1"
-signal-hook = { version = "0.3.10", optional = true }
+signal-hook = { version = "0.3.10" }
 sketches-ddsketch = "0.1"
 smallvec = { version = "1.7", features = ["union"] }
 socket2 = { version = "~0.3", features = ["unix", "reuseport"] }
@@ -59,7 +59,6 @@ cc = "1.0"
 [features]
 bench = []
 debugging = []
-stall-detection = ["signal-hook", "backtrace"]
 
 [[bench]]
 name = "executor"
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -56,6 +56,7 @@ cc = "1.0"
 [features]
 bench = []
 debugging = []
+stall-detection = []
 
 [[bench]]
 name = "executor"
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -44,9 +44,11 @@ use tracing::trace;
 use std::{
     cell::RefCell,
     collections::{hash_map::Entry, BinaryHeap},
+    fmt,
     future::Future,
     io,
     marker::PhantomData,
+    ops::Deref,
     pin::Pin,
     rc::Rc,
     sync::Arc,
@@ -60,9 +62,9 @@ use scoped_tls::scoped_thread_local;
 
 use log::warn;
 
-use crate::executor::stall::StallDetector;
 use crate::{
     error::BuilderErrorKind,
+    executor::stall::StallDetector,
     parking,
     reactor,
     sys,
@@ -444,7 +446,7 @@ pub struct LocalExecutorBuilder {
     record_io_latencies: bool,
     /// Whether to detect stalls in unyielding tasks.
     /// This installs a signal handler for SIGUSR1, so is disabled by default.
-    detect_stalls: bool,
+    detect_stalls: Option<Box<dyn stall::StallDetectionHandler + 'static>>,
 }
 
 impl LocalExecutorBuilder {
@@ -462,7 +464,7 @@ impl LocalExecutorBuilder {
             ring_depth: DEFAULT_RING_SUBMISSION_DEPTH,
             preempt_timer_duration: DEFAULT_PREEMPT_TIMER,
             record_io_latencies: false,
-            detect_stalls: false,
+            detect_stalls: None,
         }
     }
 
@@ -534,8 +536,11 @@ impl LocalExecutorBuilder {
     /// Whether to detect stalls in unyielding tasks.
     /// This installs a signal handler for SIGUSR1, so is disabled by default.
     #[must_use = "The builder must be built to be useful"]
-    pub fn detect_stalls(mut self, enabled: bool) -> Self {
-        self.detect_stalls = enabled;
+    pub fn detect_stalls(
+        mut self,
+        handler: Option<Box<dyn stall::StallDetectionHandler + 'static>>,
+    ) -> Self {
+        self.detect_stalls = handler;
         self
     }
 
@@ -621,8 +626,8 @@ impl LocalExecutorBuilder {
         let ring_depth = self.ring_depth;
         let preempt_timer_duration = self.preempt_timer_duration;
         let spin_before_park = self.spin_before_park;
-        let record_io_latencies = self.record_io_latencies;
         let detect_stalls = self.detect_stalls;
+        let record_io_latencies = self.record_io_latencies;
 
         Builder::new()
             .name(name)
@@ -675,7 +680,6 @@ impl Default for LocalExecutorBuilder {
 ///
 /// handles.join_all();
 /// ```
-#[derive(Debug)]
 pub struct LocalExecutorPoolBuilder {
     /// Spin for duration before parking a reactor
     spin_before_park: Option<Duration>,
@@ -700,7 +704,29 @@ pub struct LocalExecutorPoolBuilder {
     record_io_latencies: bool,
     /// Whether to detect stalls in unyielding tasks.
     /// This installs a signal handler for SIGUSR1, so is disabled by default.
-    detect_stalls: bool,
+    handler_gen: Option<Box<dyn Fn() -> Box<dyn stall::StallDetectionHandler + 'static>>>,
+}
+
+impl fmt::Debug for LocalExecutorPoolBuilder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let gen = self
+            .handler_gen
+            .as_ref()
+            .map(|_| "Some(Box<dyn Fn() -> Box<dyn stall::StallDetectionHandler + 'static>)");
+        write!(
+            f,
+            "spin_before_park:{:?}, name:{}, io_memory:{}, ring_depth:{}, \
+             preempt_timer_duration:{:?}, placement:{:?}, record_io_latencies:{}, handler_gen:{:?}",
+            self.spin_before_park,
+            self.name,
+            self.io_memory,
+            self.ring_depth,
+            self.preempt_timer_duration,
+            self.placement,
+            self.record_io_latencies,
+            gen
+        )
+    }
 }
 
 impl LocalExecutorPoolBuilder {
@@ -718,7 +744,7 @@ impl LocalExecutorPoolBuilder {
             preempt_timer_duration: DEFAULT_PREEMPT_TIMER,
             placement,
             record_io_latencies: false,
-            detect_stalls: false,
+            handler_gen: None,
         }
     }
 
@@ -778,8 +804,11 @@ impl LocalExecutorPoolBuilder {
     /// Whether to detect stalls in unyielding tasks.
     /// This installs a signal handler for SIGUSR1, so is disabled by default.
     #[must_use = "The builder must be built to be useful"]
-    pub fn detect_stalls(mut self, enabled: bool) -> Self {
-        self.detect_stalls = enabled;
+    pub fn detect_stalls<G: 'static>(
+        mut self,
+        handler_gen: Option<Box<dyn Fn() -> Box<dyn stall::StallDetectionHandler + 'static>>>,
+    ) -> Self {
+        self.handler_gen = handler_gen;
         self
     }
 
@@ -845,7 +874,7 @@ impl LocalExecutorPoolBuilder {
             let preempt_timer_duration = self.preempt_timer_duration;
             let spin_before_park = self.spin_before_park;
             let record_io_latencies = self.record_io_latencies;
-            let detect_stalls = self.detect_stalls;
+            let detect_stalls = self.handler_gen.as_ref().map(|x| (*x.deref())());
             let latch = Latch::clone(latch);
 
             move || {
@@ -969,7 +998,7 @@ pub struct LocalExecutor {
     reactor: Rc<reactor::Reactor>,
 
     signal_id: Option<signal_hook::SigId>,
-    stall_detector: StallDetector,
+    stall_detector: Option<StallDetector>,
 }
 
 impl LocalExecutor {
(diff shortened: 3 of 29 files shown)


glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,305 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the MIT/Apache-2.0 License, at your convenience
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
```
glommer, line 4: 2022

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,305 @@
...
+//
+#![allow(dead_code)]
```
glommer, line 6: this is not ideal. If you absolutely must mask something, best to mask it at the use-site level, not the file level

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,305 @@
...
+        })
+    }
+
+    pub(crate) fn enter_task_queue(
+        &self,
+        queue_name: String,
+        start: Instant,
+        max_expected_runtime: Duration,
+    ) -> nix::Result<StallDetectorGuard<'_>> {
+        // We consider a queue to be stalling the system if it failed to yield in due
```
glommer, line 114: on the other hand for a 500us 10% does little if anything.

The application will know better what threshold to use, so I think it is best to have something user-defined than this heuristic.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl<'detector> StallDetectorGuard<'detector> {
+    fn new(
+        detector: &'detector StallDetector,
+        queue_name: String,
+        start: Instant,
+        threshold: Duration,
+    ) -> nix::Result<Self> {
+        detector.disarm().unwrap();
+        detector.arm(threshold).unwrap();
```
glommer, line 180: Don't quote me on that because my memory is failing on this, but I am pretty sure we should be able to do a rearm with a single syscall?

disarm + arm, if it does two syscalls, that's a bit much

glommer:
Generally good, minor comments only. The most important one is the heuristics about when to trigger. I think it is impossible that we'll get this right at the framework level, and the user has to tell us when to fire as they turn this on.

(also CI is not passing)

glommio/src/executor/stall.rs
HippoBaro: Agreed. @davidblewett let's use the handler object for that like we for deciding which signal to use.

glommio/src/executor/stall.rs
HippoBaro: Rearming will implicitly override the old timer. IIRC that's how my original work did it and it worked fine

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,305 @@
...
+}
+
+impl fmt::Debug for StallDetection<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "[stall-detector -- executor {}] task queue {} went over-budget: {:#?} (budget: \
+             {:#?}). Backtrace: {:#?}",
+            self.executor, self.task_queue, self.overage, self.budget, self.trace,
+        )
```
HippoBaro, line 34: This works but that's not the best way to do this. Here's a better way:
```
fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_struct("RingIoStats")
            .field("executor", &self.executor)
            // ...
            .finish()
}
```

HippoBaro: You should also implement `Display` for this one. And the above is probably a good output for that!

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,305 @@
...
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread::JoinHandle,
+    time::{Duration, Instant},
+};
+
+pub struct StallDetection<'a> {
+    executor: usize,
+    task_queue: &'a str,
```
HippoBaro, line 21: the task queue name is fine, but you definitely want the task queue handle in there as well.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,305 @@
...
+
+#[derive(Debug)]
+pub(crate) struct StallDetector {
+    timer: Arc<sys::timerfd::TimerFd>,
+    stall_handler: Box<dyn StallDetectionHandler + 'static>,
+    timer_handler: Option<JoinHandle<()>>,
+    id: usize,
+    terminated: Arc<AtomicBool>,
+    //NOTE: we don't use signal_hook::low_level::channel as backtraces
+    //      have too many elements
```
HippoBaro, line 67: nit formatting

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,335 @@
...
+        // triggers if it doesn't yield after running for 110ms.
+        // * If a task queue has a preempt timer of 5ms the the stall detector
+        // triggers if it doesn't yield after running for 15ms.
+        Some(Duration::from_millis((max_expected_runtime.as_millis() as f64 * 0.1) as u64)
+            .max(Duration::from_millis(10)))
+    }
+
+    /// The default signal is SIGUSR1.
+    fn signal(&self) -> u8 {
+        nix::libc::SIGUSR1 as u8
```
HippoBaro, line 86: Same here, let's make this configurable

HippoBaro (changes requested):
Some nits.

My one concern is that now that I am reading the code, I realize why I used TLS for the channels in my original PR. The reason is that signal handlers are global entities. Therefore I see two issues with this version of the code:

* First, you install a signal handler every time an executor is created. So either (1) it is replaced each time or (2) the library does something clever underneath. Let's make sure it's the former. 
* Second, only a single executor will ever be signaled (the last one), and it will forward the traces to itself (because the handler captures the channels to send to). The only way for this to work is for the signal handler to fetch the channel to send to using TLS. Fortunately, the executor is hosted using TLS, so this should be trivial (Look at `LOCAL_EX`).

I think the best way to install the signal handler is to use some global entirety with a lazy initialization (you need it to be lazy since we can configure what signal to use at run time.)

One last thing... What happens if two executors specify different signals? That's a problem. I say panic in that case. If they are doing that they were looking for trouble.

clippy.toml
```diff
@@ -0,0 +1 @@
+too-many-arguments-threshold = 8
```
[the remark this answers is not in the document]
glommer: on your comment above, Hippo, maybe the executors shouldn't be able to specify the signal.
If we really want this to be configurable, an env variable is likely the way to go.

glommio/Cargo.toml
```diff
@@ -17,6 +17,7 @@ readme = "../README.md"
 
 [dependencies]
 ahash = "~0.7"
+backtrace = { version = "0.3.61" }
```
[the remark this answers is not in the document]
glommer: in this case we likely need a minimum patchlevel, though. There is a fix that I added to the backtrace crate without which this won't work (although I wouldn't be able to tell you which version has that)

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,335 @@
...
+        // (and an absolute minimum of 10ms) after which we record a stacktrace. i.e. a
+        // task queue has should return shortly after `need_preempt()` returns
+        // true or the stall detector triggers. For example::
+        // * If a task queue has a preempt timer of 100ms the the stall detector
+        // triggers if it doesn't yield after running for 110ms.
+        // * If a task queue has a preempt timer of 5ms the the stall detector
+        // triggers if it doesn't yield after running for 15ms.
+        Some(Duration::from_millis((max_expected_runtime.as_millis() as f64 * 0.1) as u64)
+            .max(Duration::from_millis(10)))
+    }
```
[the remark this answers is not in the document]
davidblewett: This is the impl for `DefaultStallDetectionHandler`?  I thought our customization method was going to be for someone to implement the `StallDetectionHandler` trait. I'm not a fan of knobs on knobs when it comes to configuration. It becomes increasingly difficult to figure out what exactly is going on. If there's a  `DefaultStallDetectionHandler`, we know exactly what values it's running with.

That being said, I'll bow to consensus if you and @glommer want to make it configurable another layer.

clippy.toml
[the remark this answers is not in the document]
glommer: Will this be addressed?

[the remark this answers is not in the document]
HippoBaro: Yeah this has to go. My work is merged so there is now a proper struct that avoids going over the argument limit.

glommio/src/executor/stall.rs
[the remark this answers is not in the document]
HippoBaro: I don't think calling this the "default" handler is a good idea. Stall detection will be disabled by default and users will have to provide this handler explicitly to enable it. 

I would call it a logging handler, personally since that's what it does. 

As for configuration maybe @glommer has a strong opinion on this. I feel like different applications will have different toleration levels for stalls and so it's not going to be possible for us to give good constants that will work for everyone. 

Instead, I would make it completely customizable and implement `Default`.

glommio/Cargo.toml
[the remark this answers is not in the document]
davidblewett: The version turned out to be `0.3.58`. I updated the pin to that minimum.

davidblewett:
> My one concern is that now that I am reading the code, I realize why I used TLS for the channels in my original PR. The reason is that signal handlers are global entities. Therefore I see two issues with this version of the code:
> 
> * First, you install a signal handler every time an executor is created. So either (1) it is replaced each time or (2) the library does something clever underneath. Let's make sure it's the former.

The library does (2). From the [docs](https://docs.rs/signal-hook/latest/signal_hook/low_level/fn.register.html): ```This makes sure there’s a signal handler for the given signal. It then adds the action to the ones called each time the signal is delivered. If multiple actions are set for the same signal, all are called, in the order of registration.```

> * Second, only a single executor will ever be signaled (the last one), and it will forward the traces to itself (because the handler captures the channels to send to). The only way for this to work is for the signal handler to fetch the channel to send to using TLS. Fortunately, the executor is hosted using TLS, so this should be trivial (Look at `LOCAL_EX`).

I added a unit test that exercises this behavior. The appropriate thread receives the signal. I also added checks to make sure that the signal handler runs in the expected thread, to handle the case where the process as a whole receives the signal and the OS picked our thread to handle it.

> I think the best way to install the signal handler is to use some global entirety with a lazy initialization (you need it to be lazy since we can configure what signal to use at run time.)

I'd like to stick with the current approach; I think the abstractions provided by `signal_hook` provide the functionality we need.

> One last thing... What happens if two executors specify different signals? That's a problem. I say panic in that case. If they are doing that they were looking for trouble.

I added a unit test exercising this. I don't think it's a problem with the implementation as is.

glommio/src/executor/stall.rs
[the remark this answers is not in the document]
davidblewett: Sounds good on renaming and adding the customization bits.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,543 @@
...
+        &self,
+        queue_handle: TaskQueueHandle,
+        max_expected_runtime: Duration,
+    ) -> Option<Duration>;
+
+    /// What signal number to use; see values in libc::SIG*
+    fn signal(&self) -> u8;
+
+    /// Handler called when a task exceeds its budget
+    fn stall(&self, detection: StallDetection<'_>);
```
HippoBaro, line 68: Let's add a comment to explain this callback is invoked from inside the reactor. We should ask users to not perform blocking syscalls, not to acquire locks, not to spawn tasks, etc.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,543 @@
...
+                if tx.is_full() || thread::current().id() != exec_thread {
+                    return;
+                }
+
+                backtrace::trace_unsynchronized(|frame| {
+                    tx.try_send(backtrace::BacktraceFrame::from(frame.clone()))
+                        .is_ok()
+                });
+            })?;
+            //.map_err::<io::Error, GlommioError<()>>(GlommioError::from)?;
```
HippoBaro, line 201: nit

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,543 @@
...
+        timer::sleep,
+        LocalExecutorBuilder,
+    };
+    use std::{
+        sync::{Arc, RwLock},
+        thread,
+        time::Duration,
+    };
+
+    #[allow(dead_code)]
```
HippoBaro, line 346: I don't think this is necessary?

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,543 @@
...
+                        .unwrap()
+                        .detections
+                        .pop()
+                        .is_some()
+                );
+
+                // no stall because < 50ms of un-cooperativeness
+                thread::sleep(Duration::from_millis(40));
+
+                assert!(stall_handler.inner.read().unwrap().detections.is_empty());
```
HippoBaro, line 438: This will never trigger. Not because you have not stalled but because the handler is called after the task queue has returned. Since you are executing in the queue, it will always be empty.

You need to move this assertion after the yielding point just below.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,543 @@
...
+            let handler = TestHandler::new(nix::libc::SIGUSR1 as u8);
+            let tname = format!("exec{}", i);
+            let builder = LocalExecutorBuilder::default()
+                .name(&tname)
+                .detect_stalls(Some(Box::new(handler.clone())))
+                .preempt_timer(Duration::from_millis(50));
+            build_handlers.push((handler, builder));
+        }
+        let mut handles = Vec::with_capacity(10);
+        for (handler, builder) in build_handlers.drain(..) {
```
HippoBaro, line 483: Not an issue but FYI you don't need to drain here, just iterate on `build_handlers` and that will consume it.

HippoBaro (changes requested):
Some very final comments but overall this is looking really good

glommio/src/executor/stall.rs
davidblewett: Not all of the fields are used by tests currently. I can remove the fields if it's a problem?

glommio/src/executor/mod.rs
```diff
@@ -680,6 +715,32 @@ pub struct LocalExecutorPoolBuilder {
...
+            "spin_before_park:{:?}, name:{}, io_memory:{}, ring_depth:{}, \
+             preempt_timer_duration:{:?}, placement:{:?}, record_io_latencies:{}, handler_gen:{:?}",
+            self.spin_before_park,
+            self.name,
+            self.io_memory,
+            self.ring_depth,
+            self.preempt_timer_duration,
+            self.placement,
+            self.record_io_latencies,
+            gen
```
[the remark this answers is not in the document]
HippoBaro: bump on this; it was marked resolved but it hasn't changed. Maybe there was a mixup :shrug:

glommio/src/executor/stall.rs
HippoBaro: If you plan to add tests that use this then let's wait on those; otherwise, let's remove it :)

glommio/src/executor/stall.rs
HippoBaro: In fact, I would be almost tempted to unset the LOCAL_EX TLS var just so that trying to enter the executor would panic the app.

glommio/src/executor/stall.rs
HippoBaro: Btw I think you can just return the directly rather than using `?` + `Ok(signal_id)`

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,534 @@
...
+    /// How far past the preemption timer should qualify as a stall
+    /// If None is returned, don't use the stall detector for this task queue.
+    fn high_water_mark(
+        &self,
+        queue_handle: TaskQueueHandle,
+        max_expected_runtime: Duration,
+    ) -> Option<Duration>;
+
+    /// What signal number to use; see values in libc::SIG*
+    fn signal(&self) -> u8;
```
glommer, line 65: why don't those have default implementations ?
I think it is best for a user to usually only specify the water mark, the rest I'd expect to be defaults most of the time.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,534 @@
...
+    }
+}
+
+/// Trait describing what signal to use to trigger stall detection,
+/// how far past expected execution time to trigger a stall,
+/// and how to handle a stall detection once triggered.
+pub trait StallDetectionHandler: std::fmt::Debug + Send + Sync {
+    /// How far past the preemption timer should qualify as a stall
+    /// If None is returned, don't use the stall detector for this task queue.
+    fn high_water_mark(
```
glommer, line 58: this name feels weird to me, maybe it is an appendage of earlier design?
Usually a high watermark implies a low watermark...

glommio/src/executor/stall.rs
HippoBaro: I didn't think about that, I agree it would be best to provide an overridable default implementation for that

glommio/src/executor/stall.rs
HippoBaro: I take the blame here, I suggested the name and I suck at names :laughing: Maybe `threshold` would work better for you?

glommer: something like `trigger_threshold` or `threshold` would indeed be better, but ultimately I don't have a problem with watermark. It's the `high` part that feels weird without a corresponding low

glommio/src/executor/stall.rs
davidblewett: I can see providing a default for `signal` (`SIGUSR1`), but I don't see what a default would be for the `stall` function? There could be any number of actions someone might want to take. The logging one just logs, but we're probably going to add one that emits Datadog events or something along those lines.

glommio/src/executor/stall.rs
davidblewett: I'll go with `threshold`.

glommio/src/executor/stall.rs
glommer: That just means that the one that emits Datadog events won't use the default =)
Logging is a great default.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,534 @@
...
+/// 10% of the expected run time. The default handler will log a stack trace of
+/// the currently executing task queue. The default signal number is
+/// [`nix::libc::SIGUSR1`].
+#[derive(Debug)]
+pub struct LoggingStallDetectionHandler {
+    /// What signal to use to trigger stall detection
+    pub signal: u8,
+    /// What percentage of the original expected runtime to allow before
+    /// triggering a stall
+    pub overage_ratio: f64,
```
glommer, line 86: I am still not a fan of the float ratio.
My reason for that is that it is hard to reason about.

I see where you're coming from: because there can't be different latency goals, you'd likely be okay with a 30ms overshoot on a 100ms latency goal and only okay with a 3ms overshoot on a 10ms goal.

But there are two big problems with that:
1) It is essentially impossible to predict what the goal will be. The latency threshold is not a property of a task queue: it is a property of a *set of runnable task queues*. If there is a (runnable) task queue with a 100ms threshold and another with 1ms, then the threshold is 1ms. This comes and goes and I think it complicates the design.

2) In an ideal world, nobody ever goes over the task quota, at any time, for no period of time. (or an absolute threshold == 0) Of course such world doesn't exist, so we log those events. But it doesn't sound okay to me to overshoot the quota for 30ms just because your task queue has a higher threshold. Ultimately, the stall detector is not about your task queue, but about the fact that you're preventing other task queues to run by clogging the pipes.

glommio/src/executor/stall.rs
```diff
@@ -0,0 +1,534 @@
...
+    /// Absolute minimum duration before triggering a stall
+    pub min_duration: Duration,
+}
+
+impl LoggingStallDetectionHandler {
+    fn new(signal: u8, overage_ratio: f64, min_duration: Duration) -> Self {
```
glommer, line 92: I think we should provide a concrete class with no settable properties, and allow extensions only through traits.

glommer (commented):
Very minor comments at this point. 
In the trait vs tunables, I prefer the traits and you seem too as well, so let's do that.

Let me know what you think about my argument against the percentage-based threshold.
I think we should provide a concrete class with, say, 10ms flat default and then allow overriding through traits.

glommio/src/executor/stall.rs
HippoBaro: I was pushing for a configurable class that provides `Default`

davidblewett:
> Let me know what you think about my argument against the percentage-based threshold. I think we should provide a concrete class with, say, 10ms flat default and then allow overriding through traits.

This seems reasonable to me. It's very easy to implement the exact behavior you want if this doesn't fit your use case.

davidblewett:
The test failures are due to newer lints in rust 1.58+, that seem unrelated to this PR.

glommio/src/executor/mod.rs
```diff
@@ -566,6 +578,28 @@ impl LocalExecutorBuilder {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use glommio::{DefaultStallDetectionHandler, LocalExecutorBuilder};
+    ///
+    /// let local_ex = LocalExecutorBuilder::default()
+    ///     .detect_stalls(Some(Box::new(DefaultStallDetectionHandler {})))
+    ///     .make()
+    ///     .unwrap();
+    /// ```
```
HippoBaro, line 593: Can you use the same doc as in there? https://github.com/DataDog/glommio/pull/499/files#diff-5b724ff10c950f45f86b8b247e7b6000f5f6cba737c7ca8e2ebe1ec5e85bc560R876-R899

glommio/src/executor/mod.rs
```diff
@@ -1060,7 +1155,31 @@ impl LocalExecutor {
...
+    pub fn detect_stalls(
+        &mut self,
+        handler: Option<Box<dyn stall::StallDetectionHandler + 'static>>,
+    ) -> Result<()> {
+        self.stall_detector = match handler {
+            Some(handler) => Some(StallDetector::new(self.id, handler)?),
+            None => None,
+        };
+        Ok(())
```
HippoBaro, line 1182: Can you add one final test that ensures this is safe for `pub` visibility? i.e., that we can safely disable it from outside the reactor? 

I'm asking because while user code is touching this, the stall detector is actively used, and a timer exists. I want to ensure that disabling it doesn't result in anything weird. 

Two little tests should do:
* A test with a cooperative task that disables the stall detector mid-course -> assert that we don't record anything;
* A test with an unyielding task that disables the stall detector mid-course -> assert that we don't produce a trace (or that we do one last time, depending on what behavior is easier to implement).

glommio/src/sys/uring.rs
```diff
@@ -1040,7 +1040,13 @@ impl SleepableRing {
...
             // it successfully to be able to safely sleep.
 
-            if self.ring.submit_sqes()? != 1 {
+            if self
+                .submit_sqes()
+                .or_else(Reactor::busy_ok)
+                .or_else(Reactor::again_ok)
+                .or_else(Reactor::intr_ok)?
+                != 1
+            {
```
HippoBaro, line 1049: This is in master so it needs to go, please drop the commit :)

glommio/src/sys/uring.rs
```diff
@@ -1049,13 +1055,22 @@ impl SleepableRing {
...
             } else {
                 // The rings are linked. Goodnight!
-                self.ring.cq().wait(1).map(|_| 1)
+                self.ring
+                    .cq()
+                    .wait(1)
+                    .map(|_| 1)
+                    .or_else(Reactor::busy_ok)
+                    .or_else(Reactor::again_ok)
+                    .or_else(Reactor::intr_ok)
```
HippoBaro, line 1064: ditto

glommio/src/sys/uring.rs
```diff
@@ -1049,13 +1055,22 @@ impl SleepableRing {
...
             }
         } else {
             // Can't link rings because we ran out of `CQE`s. Just can't sleep.
             // Submit what we have, once we're out of here we'll consume them
             // and at some point will be able to sleep again.
             self.submit_sqes()
+                .or_else(Reactor::busy_ok)
+                .or_else(Reactor::again_ok)
+                .or_else(Reactor::intr_ok)
```
HippoBaro, line 1073: ditto

glommio/src/sys/uring.rs
```diff
@@ -1548,9 +1563,9 @@ impl Reactor {
     /// delivery. This is fine: we just need to make sure that we
     /// don't sleep and that we don't failed rushed polls. So we just ignore
     /// this error
-    fn intr_ok(x: std::io::Error) -> io::Result<usize> {
+    fn intr_ok<T: Default>(x: std::io::Error) -> io::Result<T> {
         match x.raw_os_error() {
-            Some(libc::EINTR) => Ok(0),
+            Some(libc::EINTR) => Ok(Default::default()),
```
HippoBaro, line 1568: ditto

HippoBaro (changes requested):
Almost there!!

glommio/src/executor/mod.rs
davidblewett: But they're different? `LocalExecutorBuilder` only takes the stall handler directly. `LocalExecutorPoolBuilder` takes a handler factory function, and its doc uses `on_all_shards` which isn't on `LocalExecutorBuilder`?

glommio/src/executor/mod.rs
davidblewett: I removed the export for now.

HippoBaro (approved):

PR #500, opened 2022-01-14
do not panic when failing to create a `LocalExecutor`

HippoBaro:
Errors should propagate to the end users instead. Additionally,
allow returning a struct from the boot future. It is possible to do so
when using the pool builder but not when using the simple one, so this
brings them to parity.

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -52,7 +52,7 @@ async fn server(conns: usize) -> Result<()> {
         s.await.unwrap()?;
     }
 
-    client_handle.join().unwrap();
+    client_handle.join().unwrap().unwrap();
     Ok(())
 }
 
@@ -108,6 +108,6 @@ fn main() -> Result<()> {
     //
     // Now can you adapt it, so it uses multiple executors and all CPUs in your
     // system?
-    server_handle.join().unwrap();
+    server_handle.join().unwrap().unwrap();
     Ok(())
 }
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -385,6 +385,29 @@ impl ExecutorQueues {
     }
 }
 
+/// A wrapper around a [`std::thread::JoinHandle`]
+#[derive(Debug)]
+pub struct ExecutorJoinHandle<T: Send + 'static>(JoinHandle<Result<T>>);
+
+impl<T: Send + 'static> ExecutorJoinHandle<T> {
+    /// See [`std::thread::JoinHandle::thread()`]
+    #[must_use]
+    pub fn thread(&self) -> &std::thread::Thread {
+        self.0.thread()
+    }
+
+    /// See [`std::thread::JoinHandle::join()`]
+    pub fn join(self) -> Result<T> {
+        match self.0.join() {
+            Err(err) => Err(GlommioError::BuilderError(BuilderErrorKind::ThreadPanic(
+                err,
+            ))),
+            Ok(Err(err)) => Err(err),
+            Ok(Ok(res)) => Ok(res),
+        }
+    }
+}
+
 /// A factory that can be used to configure and create a [`LocalExecutor`].
 ///
 /// Methods can be chained on it in order to configure it.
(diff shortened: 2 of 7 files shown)


glommio/src/executor/mod.rs
```diff
@@ -592,10 +592,11 @@ impl LocalExecutorBuilder {
     /// [`LocalExecutor::run`]:struct.LocalExecutor.html#method.run
     #[must_use = "This spawns an executor on a thread, so you may need to call \
                   `JoinHandle::join()` to keep the main thread alive"]
-    pub fn spawn<G, F, T>(self, fut_gen: G) -> Result<JoinHandle<()>>
+    pub fn spawn<G, F, T>(self, fut_gen: G) -> io::Result<JoinHandle<io::Result<T>>>
```
bryandmc, line 595: Shouldn't these Results be glommio::Error variants to explain to the user what happened? Also shouldn't we just give the user the ability to return a result of their choosing from the main future?

glommio/src/channels/shared_channel.rs
```diff
@@ -524,8 +524,8 @@ mod test {
             })
             .unwrap();
 
-        ex1.join().unwrap();
-        ex2.join().unwrap();
+        ex1.join().unwrap().unwrap();
```
bryandmc, line 527: At some point it might make more sense to have the tests return Result<(), E> type thing so we can maybe get away with using `?` in a bunch of these unwraps.. obviously I haven't worked this idea full through, but it _might_ produce a nicer result and an easier pattern than unwrapping so many unwraps etc.. Thoughts?

bryandmc (commented):
Besides some stylistic ideas and the API thing regarding what type of result we return, the rest is pretty straight forward and all good.

glommio/src/channels/shared_channel.rs
HippoBaro: Agreed, these chains of `unwrap` are not nice. I am thinking of wrapping the `JoinHandle` into one of our making to flatten the two errors (from creating the thread and from creating the local executor inside the thread) into a single one.

HippoBaro:
I updated the PR to include a new type of `JoinHandle` that flattens the errors from creating the thread and from creating the local executor inside the thread. This removes all the spurious `.unwrap().unwrap()` from the last version.

examples/echo.rs
```diff
@@ -52,7 +52,7 @@ async fn server(conns: usize) -> Result<()> {
         s.await.unwrap()?;
     }
 
-    client_handle.join().unwrap();
+    client_handle.join().unwrap().unwrap();
```
HippoBaro, line 55: This one remains because we now return whatever the user returns from the bootstrap future. In the case of this example, we return an error (that we were dropping on the floor before this patchset).

examples/echo.rs
```diff
@@ -108,6 +108,6 @@ fn main() -> Result<()> {
     //
     // Now can you adapt it, so it uses multiple executors and all CPUs in your
     // system?
-    server_handle.join().unwrap();
+    server_handle.join().unwrap().unwrap();
```
HippoBaro, line 111: Same here

glommio/src/executor/mod.rs
HippoBaro: We do! The `T` in there can be anything the user wants. They can return a `Result` if they so wish. The two errors you see are respectively:
* The error thrown by `join()` if the `std::thread::Thread` failed to create or panicked
* The error thrown by `LocalExecutor::new` if we fail to create it

glommio/src/channels/shared_channel.rs
bryandmc: Yes! Love the flattening idea. I think that would clean up a lot of stuff and make the api cleaner.

examples/echo.rs
bryandmc: yes that's fine, of course

glommio/src/executor/mod.rs
bryandmc: It still seems like we need to return a glommio Error type (not an io::Error, from the io::Result) for functions that are public like this.

glommer (approved):

glommer:
Agree with Bryan on not using io::Result, but other than that looks good.

HippoBaro:
The latest version uses GlommioError all the way through. Users can now inspect exactly why the executor failed to create :+1:

PR #501, opened 2022-01-14
add `spawn_blocking` and use a blocking thread pool

HippoBaro:
Sometimes there is no way around doing a long, unyielding job. That's
the world we live in. These fall into two categories:
* Busy blocking jobs: usually, these are calls to external libraries
  that aren't cooperative and can't be easily broken into bite-size
  pieces. Your usual suspects are compression/decompression and
  marshaling/unmarshalling.
* Non-busy blocking jobs: stuff that acquires locks or syscalls waiting
  for events.

`spawn_blocking` works for both cases but is best suited for the latter
because they don't steal CPU cycles from the executor.

So far, we have had a single thread per executor used for blocking
operations. This has been fine to issue the odd syscall that couldn't go
into the ring, but now that we have `spawn_blocking,` we've opened
pandora's box. Users will submit all sorts of things in there, including
long-running jobs that could stall the entire system.

This commit introduces a static thread pool per executor instead. The
pool is static for now, meaning the threads are created once and live
for as long as the host executor does. Later on, we may want to be
smarter and automatically up/downscale the pool depending on usage.

This work is based #500.

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -464,6 +464,10 @@ pub struct LocalExecutorBuilder {
     preempt_timer_duration: Duration,
     /// Whether to record the latencies of individual IO requests
     record_io_latencies: bool,
+    /// The placement policy of the blocking thread pool
+    /// Defaults to one thread using the same placement strategy as the host
+    /// executor
+    blocking_thread_pool_placement: PoolPlacement,
 }
 
 impl LocalExecutorBuilder {
@@ -474,13 +478,14 @@ impl LocalExecutorBuilder {
     /// how many and which CPUs to use.
     pub fn new(placement: Placement) -> LocalExecutorBuilder {
         LocalExecutorBuilder {
-            placement,
+            placement: placement.clone(),
             spin_before_park: None,
             name: String::from(DEFAULT_EXECUTOR_NAME),
             io_memory: DEFAULT_IO_MEMORY,
             ring_depth: DEFAULT_RING_SUBMISSION_DEPTH,
             preempt_timer_duration: DEFAULT_PREEMPT_TIMER,
             record_io_latencies: false,
+            blocking_thread_pool_placement: PoolPlacement::from(placement),
         }
     }
 
@@ -549,6 +554,18 @@ impl LocalExecutorBuilder {
         self
     }
 
+    /// The placement policy of the blocking thread pool.
+    /// Defaults to one thread using the same placement strategy as the host
+    /// executor.
+    #[must_use = "The builder must be built to be useful"]
+    pub fn blocking_thread_pool_placement(
+        mut self,
+        placement: PoolPlacement,
+    ) -> LocalExecutorBuilder {
+        self.blocking_thread_pool_placement = placement;
+        self
+    }
+
     /// Make a new [`LocalExecutor`] by taking ownership of the Builder, and
     /// returns a [`Result`](crate::Result) to the executor.
     /// # Examples
@@ -563,12 +580,15 @@ impl LocalExecutorBuilder {
         let mut cpu_set_gen = placement::CpuSetGenerator::one(self.placement)?;
         let mut le = LocalExecutor::new(
             notifier,
-            self.io_memory,
-            self.ring_depth,
-            self.preempt_timer_duration,
-            self.record_io_latencies,
             cpu_set_gen.next().cpu_binding(),
-            self.spin_before_park,
+            LocalExecutorConfig {
+                io_memory: self.io_memory,
+                ring_depth: self.ring_depth,
+                preempt_timer: self.preempt_timer_duration,
+                record_io_latencies: self.record_io_latencies,
+                spin_before_park: self.spin_before_park,
+                thread_pool_placement: self.blocking_thread_pool_placement,
+            },
         )?;
         le.init();
         Ok(le)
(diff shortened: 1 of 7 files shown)


examples/echo.rs
```diff
@@ -52,7 +52,7 @@ async fn server(conns: usize) -> Result<()> {
         s.await.unwrap()?;
     }
 
-    client_handle.join().unwrap();
+    client_handle.join().unwrap().unwrap().unwrap();
```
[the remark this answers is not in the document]
HippoBaro: This is part of #500 (and it's been fixed since :smiley: )

glommio/src/executor/placement/mod.rs
```diff
@@ -184,6 +184,26 @@ impl PoolPlacement {
     pub fn generate_cpu_set(self) -> Result<CpuSetGenerator> {
         CpuSetGenerator::pool(self)
     }
+
+    /// Shrinks the pool placement policy to the first `len` placements.
+    ///
+    /// If `len` is greater than the number of placements in the pool, this has
+    /// no effect.
+    pub fn shrink_to(self, count: usize) -> Self {
```
glommer, line 192: why would this be public?
This looks confusing to me as an API

glommio/src/executor/mod.rs
```diff
@@ -2338,6 +2340,69 @@ impl ExecutorProxy {
...
+    /// performance of the overall system, not just to the performance of
+    /// the one stalling task.
+    ///
+    /// `spawn_blocking` is there as a last resort when a blocking task needs to
+    /// be executed and cannot be made cooperative. Examples are:
+    /// * Expensive syscalls that cannot use `io_uring`, such as `mmap`
+    ///   (especially with `MAP_POPULATE`)
+    /// * Calls to synchronous third-party code (compression, encoding, etc.)
+    ///
+    /// It is recommended to guard against calling this method if the blocking
```
glommer, line 2406: I don't agree, and we shouldn't be in the business of making this recommendation.

Let's remove this text

And if you're curious I can give you a lot of examples where even seemingly harmless things like "small mmap" can take 7+ seconds.

glommio/src/executor/mod.rs
```diff
@@ -2338,6 +2340,69 @@ impl ExecutorProxy {
...
+    ///         })
+    ///         .await;
+    /// });
+    /// ```
+    pub fn spawn_blocking<F, R>(&self, func: F) -> impl Future<Output = R>
+    where
+        F: FnOnce() -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        let result = Arc::new(Mutex::new(MaybeUninit::<R>::uninit()));
```
glommer, line 2436: this overall looks pretty good but I am not sure I like this mutex.

I suspect we could use `from_raw()` and `into_raw()` and pass the pointer. If not, ultimately it is fine.

glommio/src/sys/blocking.rs
```diff
@@ -47,12 +47,12 @@ macro_rules! c_str {
     };
 }
 
-#[derive(Clone, Debug)]
```
glommer, line 50: please keep debug implementations.
If you can't auto-derive them, provide one.

glommio/src/sys/uring.rs
```diff
@@ -1258,7 +1259,7 @@ impl Reactor {
             latency_ring: RefCell::new(latency_ring),
             poll_ring: RefCell::new(poll_ring),
             timeout_src: Cell::new(None),
-            syscall_thread: BlockingThread::new(notifier.clone()),
+            blocking_thread: BlockingThreadPool::new(PoolPlacement::Unbound(1), notifier.clone())?,
```
glommer: why 1?
I see no reason not to distribute threads into many cpus.

glommio/src/sys/blocking.rs
```diff
@@ -110,35 +112,99 @@ pub(super) struct BlockingThreadResp {
...
 }
 
-impl BlockingThread {
-    pub(super) fn push(&self, op: BlockingThreadOp, action: Box<dyn Fn(BlockingThreadResult)>) {
+impl BlockingThreadPool {
+    pub(crate) fn new(
+        placement: PoolPlacement,
+        sleep_notifier: Arc<SleepNotifier>,
+    ) -> io::Result<Self> {
+        let (in_tx, in_rx) = crossbeam::channel::unbounded();
```
glommer, line 160: can we use bounded channels?

glommer (approved):

glommio/src/sys/uring.rs
HippoBaro: You're looking at the commit that implements the pool itself. A way to configures the poll will come in a further commit :)

glommio/src/sys/blocking.rs
HippoBaro: Ideally I would like to find a channel that would have an async TX and a sync RX. 

I don't think it is okay to panic if the que is full like we did before. 

And it's also not okay to bubble up blocking to user code. This is meant to be async.

HippoBaro: So an ideal way to deal with this would be to provide async back pressure with an async TX. I haven't found one though. 

I think there is a way to use tokio channels that way. I'll look into it.

glommer: See if this does what you need: 
https://docs.rs/async-channel/latest/async_channel/

glommio/src/executor/placement/mod.rs
HippoBaro: It's now private

glommio/src/sys/blocking.rs
HippoBaro: It doesn't. Tokio channel however does have a `blocking_recv()`, but that would imply us depending on Tokio. Going to merge as-is but I agree this is something we'll need to keep an eye on.

PR #502, opened 2022-01-14
various reactor improvements and bug fixes

HippoBaro:
This PR contains various bug fixes to the reactor, mostly around whether
to enter the rings and whether it is okay to sleep.

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -341,7 +341,7 @@ struct ExecutorQueues {
     available_executors: AHashMap<usize, Rc<RefCell<TaskQueue>>>,
     active_executing: Option<Rc<RefCell<TaskQueue>>>,
     executor_index: usize,
-    last_vruntime: u64,
+    default_vruntime: u64,
     preempt_timer_duration: Duration,
     default_preempt_timer_duration: Duration,
     spin_before_park: Option<Duration>,
@@ -355,7 +355,7 @@ impl ExecutorQueues {
             available_executors: AHashMap::new(),
             active_executing: None,
             executor_index: 1, // 0 is the default
-            last_vruntime: 0,
+            default_vruntime: 0,
             preempt_timer_duration,
             default_preempt_timer_duration: preempt_timer_duration,
             spin_before_park,
@@ -378,7 +378,7 @@ impl ExecutorQueues {
     fn maybe_activate(&mut self, queue: Rc<RefCell<TaskQueue>>) {
         let mut state = queue.borrow_mut();
         if !state.is_active() {
-            state.vruntime = self.last_vruntime;
+            state.vruntime = self.default_vruntime + 1;
             state.active = true;
             drop(state);
             self.active_executors.push(queue);
@@ -1235,7 +1235,7 @@ impl LocalExecutor {
 
                 let runtime = time.elapsed();
 
-                let (need_repush, last_vruntime) = {
+                let (need_repush, vruntime) = {
                     let mut state = queue.borrow_mut();
                     let last_vruntime = state.account_vruntime(runtime);
                     (state.is_active(), last_vruntime)
@@ -1245,8 +1245,7 @@ impl LocalExecutor {
                 tq.active_executing = None;
                 tq.stats.executor_runtime += runtime;
                 tq.stats.tasks_executed += tasks_executed_this_loop;
-
-                tq.last_vruntime = match last_vruntime {
+                let vruntime = match vruntime {
                     Some(x) => x,
                     None => {
                         for queue in tq.available_executors.values() {
@@ -1262,6 +1261,17 @@ impl LocalExecutor {
                 } else {
                     tq.reevaluate_preempt_timer();
                 }
+
+                // Compute the smallest vruntime out of all the active task queues
+                // This value is used to set the vruntime of deactivated task queues when they
+                // are woken up.
+                tq.default_vruntime = tq
+                    .active_executors
+                    .iter()
+                    .map(|x| x.borrow().vruntime)
+                    .min()
+                    .unwrap_or(vruntime);
+
                 true
             }
             None => false,
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -1317,8 +1317,14 @@ impl LocalExecutor {
                 // requests that are latency sensitive we want them out of the
                 // ring ASAP (before we run the task queues). We will also use
                 // the opportunity to install the timer.
-                self.parker.poll_io(|| Some(self.preempt_timer_duration()));
+                self.parker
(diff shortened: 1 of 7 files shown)


glommio/src/sys/uring.rs
```diff
@@ -777,7 +777,9 @@ impl PollRing {
         let ring = iou::IoUring::new_with_flags(
             size as _,
             iou::SetupFlags::IOPOLL,
-            iou::SetupFeatures::empty(),
+            iou::SetupFeatures::FAST_POLL
```
glommer, line 780: I don't think this does what you think it does.

Read the kernel source. This is the kernel telling us which features are available, not us picking which features we want to use.

glommio/src/sys/uring.rs
```diff
@@ -1630,37 +1630,29 @@ impl Reactor {
         let mut main_ring = self.main_ring.borrow_mut();
         let mut lat_ring = self.latency_ring.borrow_mut();
 
-        // we need to make sure we consume the latency ring before setting the
-        // preemption timer because consuming this ring may wake up some previously
-        // asleep task queues.
-        consume_rings!(into &mut woke; lat_ring);
```
glommer, line 1591: How "tiny" does it have to be for it to matter?
In reality the kernel won't be able to deliver anything with less than 250us reliably anyway

glommio/src/sys/uring.rs
HippoBaro: This sets the following flag when calling `io_uring_enter`:
```
IORING_FEAT_FAST_POLL

If this flag is set, then io_uring supports using an internal poll 
mechanism to drive data/space readiness. This means that requests that 
cannot read or write data to a file no longer need to be punted to an 
async thread for handling, instead they will begin operation when the 
file is ready. This is similar to doing poll + read/write in userspace, 
but eliminates the need to do so. If this flag is set, requests waiting 
on space/data consume a lot less resources doing so as they are not 
blocking a thread. Available since kernel 5.7.
```

glommio/src/sys/uring.rs
```diff
@@ -1210,7 +1210,7 @@ macro_rules! flush_rings {
     ($( $ring:expr ),+ ) => {{
         let mut ret = 0;
         $(
-            ret += $ring.consume_submission_queue()?;
+            ret += $ring.consume_submission_queue().or_else(Reactor::busy_ok)?;
```
glommer, line 1221: isn't EAGAIN fine too ?

glommio/src/sys/uring.rs
```diff
@@ -1667,7 +1667,10 @@ impl Reactor {
             // is a timer set, we need to make sure we wake up to handle it.
             if let Some(dur) = user_timer {
                 self.timeout_src.set(Some(lat_ring.arm_timer(dur)));
-                flush_rings!(lat_ring)?;
+                if flush_rings!(lat_ring)? != 1 {
```
glommer, line 1670: agree that it isn't safe to sleep upon failures.
How reproucible is this ? A test would be great, but it may be hard to craft one

glommio/src/sys/uring.rs
```diff
@@ -1554,6 +1554,7 @@ impl Reactor {
 
     fn simple_poll(ring: &RefCell<dyn UringCommon>, woke: &mut usize) -> io::Result<()> {
         let mut ring = ring.borrow_mut();
+        ring.consume_cancellation_queue().or_else(Self::busy_ok)?;
```
glommer, line 1565: good catch!

glommio/src/sys/uring.rs
HippoBaro: That and `EINTR` I have yet to see. This version panics if it sees an unexpected error code, and I haven't seen any so far. That doesn't mean they don't happen, of course... 

I'll probably add them eventually. I just want to make our error handling is sound.

glommio/src/sys/uring.rs
glommer: Read again: If this flag is set, *THEN IO_URING SUPPORTS* . This is the kernel telling you what it supports, not you asking for the feature.

glommio/src/sys/uring.rs
HippoBaro: This is me being paranoiac. I can't reproduce this but reading the man page there's no guarantee the kernel will consume what we give it. Maybe today it always does, but who knows about tomorrow :shrug: 

I'm especially concerned about cases where some internal data structure in the kernel is full, and it decides to apply back-pressure by denying new submissions for a while.

glommio/src/executor/mod.rs
```diff
@@ -202,6 +202,10 @@ impl TaskQueue {
         let vruntime = self.vruntime.checked_add(delta_scaled);
         if let Some(x) = vruntime {
             self.vruntime = x;
+        } else if self.stats.queue_selected == 1 {
```
glommer, line 205: this is not a bug. A task queue that woke up and was previously sleeping should have its vruntime brought to the same ballpark as the others. Resetting everybody is actually a bug, because now if you had 10 task queues that were runnable and keeping their runtime fraction among themselves, a single wake will destroy all of that.

You shouldn't be seeing starvation because even if the newly woken TQ is ran right away, it will run for at most one period. If you are, maybe there's some ordering issue with the task_queues. We could set it to lowest_vruntime + 1, for example.

glommio/src/sys/uring.rs
glommer: ok, but "busy_ok" only tests for EBUSY. Your code will end up in an `except` in those cases now, no ?

glommio/src/sys/uring.rs
HippoBaro: oh lolilol

HippoBaro: That's embarrassing. Will remove!

glommio/src/executor/mod.rs
HippoBaro: Ah gotcha. I see why we'd reset the runtime upon activation. That makes sense! I'll rework!

glommio/src/sys/uring.rs
HippoBaro: Yeah, I need to change `busy_ok` to include the other, or add new fn and chain them to handle all cases.

glommio/src/executor/mod.rs
HippoBaro: This should be fixed in https://github.com/DataDog/glommio/pull/502/commits/9b3930e1e6706461cb983c25bfd8b7d298006e19

glommio/src/sys/uring.rs
HippoBaro: Fixed in 167a03139c744a68b27dbc6dafc4a9470c9248c5

PR #505, opened 2022-01-15
don't panic when trying to yield outside a Glommio context

HippoBaro:
Calls to `glommio::yield_if_needed` and friends are everywhere in
Glommio apps and calling this function panics outside a Glommio
context. Sometimes, it is helpful to move some Glommio code to a Tokio
context (in unit tests, for instance). This commit makes the panic go
away.

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -1826,14 +1826,19 @@ impl ExecutorProxy {
     where
         F: FnOnce(&LocalExecutor) -> bool,
     {
-        let need_yield = LOCAL_EX.with(|local_ex| {
-            if cond(local_ex) {
-                local_ex.mark_me_for_yield();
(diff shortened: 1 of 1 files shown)


glommio/src/executor/mod.rs
```diff
@@ -1826,14 +1826,19 @@ impl ExecutorProxy {
...
+                if cond(local_ex) {
+                    local_ex.mark_me_for_yield();
+                    true
+                } else {
+                    false
+                }
+            })
+        } else {
+            // We are not in a glommio context
+            true
```
glommer, line 1840: that will kill performance, no?
If you returned from this function right now it would not be that bad, but right below this line we use this result to `yield_now()`

HippoBaro: Good point

PR #508, opened 2022-01-17
limit test concurrency

HippoBaro:
We have a fair few tests that rely on timing and are otherwise highly
CPU intensive (`test_spin,` I am looking at you!). They have proven to
be quite flaky in CI. Ideally, we would have them run serially while
keeping the other ones running in parallel.
Rust lacks this functionality, however, so in the meantime, run
everything with a single thread in CI.


glommer:
I sadly agree to this in principle, but likely we could try a higher concurrency first?

HippoBaro:
Trying out with "--test-threads=`nproc`"

PR #509, opened 2022-01-18
add throughput-based preemption 

HippoBaro:
```
Preemption is difficult to get right. Implementations need to strike a
delicate balance between preempting too little at the risk of starving
the reactor or preempting too much at the expense of CPU performance
(d/i caches, branch prediction, etc.).

Glommio today uses two related mechanisms for preemption:
* An event-based preemption mechanism that preempts every time the
  latency ring has at least one CQE ready;
* A timer-based approach to preemption: every task queue in the system
  either defines or inherits a preemption duration representing the
  maximum amount of time we can spend outside the reactor. A timer is
  inserted in the latency ring based on this value. When it fires,
  condition (1) triggers, and we preempt.

This system is effective when a user submits some IO requests and values
the latency of those individual requests more than the system's
concurrency as a whole. In other words, the user trades throughput for
latency.

However, for workloads only interested in throughput, this system
doesn't work as well. Although it is intuitive to think a higher
preemption timer would favor throughput, it is the opposite for IO. If
the preemption timer is set to 100ms (the default), we will only submit
and collect requests once every 100ms. Although the CPU will deliver
very high throughput, the IO system will be barely utilized. Therefore,
users are encouraged to use lower and lower preemption timers to the
detriment of CPU throughput.

In this case, a good system would maximize both IO and CPU performance
by preempting exactly at the starvation tipping point: just before it
runs out of work. Such a system would maximize throughput because if the
system can process one thousand concurrent requests, then preemption
would trigger only once every 1k or so requests processed.

This is what this commit introduces: throughput-based preemption. We now
insert an additional timer in the main ring that only fires when a
minimum number of CQEs are available to collect. It does so rather
cleverly by defining a timer and linking it to a second SQE that writes
to the latency ring's eventfd. Because the latency ring always reads
this eventfd, we can trigger a preemption from an event on the main
ring.

I collected some numbers using the `competing_io` benchmark that I made 
to evaluate this use case:
```

| Preemption →   Shares↓ | 10ms                | 10ms + Tput          | 500us               | 500us + Tput        |
|-------------------------|---------------------|----------------------|---------------------|---------------------|
| IO:100  CPU:100        | IOPS: 39k CPU: 157M | IOPS: 297k CPU: 55M  | IOPS: 413k CPU: 28M | IOPS: 398k CPU: 31M |
| IO:100  CPU:10         | IOPS: 39k CPU: 159M | IOPS: 308k CPU: 54M  | IOPS: 471k CPU: 6M  | IOPS: 462k CPU: 5M  |

### DIFF
--- a/glommio/src/reactor.rs
+++ b/glommio/src/reactor.rs
@@ -45,7 +45,6 @@ use crate::{
     },
     IoRequirements,
     IoStats,
-    Latency,
     PoolPlacement,
     TaskQueueHandle,
 };
@@ -760,8 +759,7 @@ impl Reactor {
     }
 
     pub(crate) fn rush_dispatch(&self, source: &Source) -> io::Result<()> {
-        self.sys.rush_dispatch(Some(source.latency_req()), &mut 0)?;
-        Ok(())
+        self.sys.rush_dispatch(source, &mut 0)
     }
 
     /// Polls for I/O, but does not change any timer registration.
@@ -769,15 +767,9 @@ impl Reactor {
     /// This doesn't ever sleep, and does not touch the preemption timer.
     pub(crate) fn spin_poll_io(&self) -> io::Result<bool> {
         let mut woke = 0;
-        // any duration, just so we land in the latency ring
-        self.sys
-            .rush_dispatch(Some(Latency::Matters(Duration::from_secs(1))), &mut woke)?;
-        self.sys
-            .rush_dispatch(Some(Latency::NotImportant), &mut woke)?;
-        self.sys.rush_dispatch(None, &mut woke)?;
+        self.sys.poll_io(&mut woke)?;
         woke += self.process_timers().1;
         woke += self.process_shared_channels();
-        woke += self.sys.flush_syscall_thread();
 
         Ok(woke > 0)
     }
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -942,18 +942,18 @@ impl SleepableRing {
         self.ring.raw().ring_fd
     }
 
-    fn arm_timer(&mut self, d: Duration, min_events: u32) -> Source {
+    /// This function prepares a timer that fires unconditionally after a
+    /// certain duration. The timer is added at the front of the queue such
+    /// that it will be the first SQE submitted the next time we enter the
+    /// latency ring
+    fn prepare_latency_preemption_timer(&mut self, d: Duration) -> Source {
         let source = Source::new(
             IoRequirements::default(),
             -1,
-            SourceType::Timeout(TimeSpec64::from(d), min_events),
+            SourceType::Timeout(TimeSpec64::from(d), 0),
             None,
             None,
         );
-        let new_id = self
-            .source_map
-            .borrow_mut()
-            .add_source(&source, self.submission_queue.clone());
         let op = match &*source.source_type() {
             SourceType::Timeout(ts, events) => {
                 UringOpDescriptor::Timeout(&ts.raw as *const _, *events)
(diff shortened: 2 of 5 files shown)


HippoBaro:
Upon request, I disabled the IO deduplication and reran the benchmarks. IO dedup made the numbers larger and less reliable.

### 10ms preemption:

Before: latency preemption alone
After: latency+throughput preemption

| Shares          | Before              | After                |
|-----------------|---------------------|----------------------|
| IO:100  CPU:100 | IOPS: 25k CPU: 157M | IOPS: 118k CPU: 111M |
| IO:100  CPU:10  | IOPS: 24k CPU: 159M | IOPS: 150k CPU: 92M  |

### 500us preemption:

Before: latency preemption alone
After: latency+throughput preemption

| Shares          | Before              | After               |
|-----------------|---------------------|---------------------|
| IO:100  CPU:100 | IOPS: 312k CPU: 15M | IOPS: 312k CPU: 15M |
| IO:100  CPU:10  | IOPS: 324k CPU: 9M  | IOPS: 336k CPU: 1M  |

HippoBaro:
I forgot to mention that an essential knob of this new preemption is how many CQEs ready in the main ring trigger a preemption. The numbers I report above use a conservative value: the ring depth. Meaning that we only preempt if 128 CQEs are available for a ring with 128 SQE slots.

If we reduce this value, we get better numbers at the cost of more frequent preemptions.

Here are the same benchmark as above, but with this value set to 32 instead of 128:

### 10ms preemption:

Before: latency preemption alone
After: latency+throughput preemption (32 min CQEs available)

| Shares          | Before              | After                |
|-----------------|---------------------|----------------------|
| IO:100  CPU:100 | IOPS: 25k CPU: 157M | IOPS: 158k CPU: 90M |
| IO:100  CPU:10  | IOPS: 24k CPU: 159M | IOPS: 313k CPU: 7M  |

### 500us preemption:

Before: latency preemption alone
After: latency+throughput preemption (32 min CQEs available)

| Shares          | Before              | After               |
|-----------------|---------------------|---------------------|
| IO:100  CPU:100 | IOPS: 312k CPU: 15M | IOPS: 310k CPU: 13M |
| IO:100  CPU:10  | IOPS: 324k CPU: 9M  | IOPS: 324k CPU: 1M  |

glommio/src/reactor.rs
```diff
@@ -45,7 +45,6 @@ use crate::{
     },
     IoRequirements,
     IoStats,
-    Latency,
```
glommer, line 48: I don't think your heuristic (as described in the commit message) makes sense.
Why send "other tasks" on the latency ring? Other tasks will include stuff like grabbing the file size from metadata (statx). If you don't think reading a file should be in the latency ring, why do you think stat should be?

I am sympathetic towards moving stuff out of the latency ring (although I think there is a case to be made for rx networking), but then stuff that goes into it should be explicitly marked, and not "the rest".

glommio/src/sys/uring.rs
```diff
@@ -92,7 +92,7 @@ enum UringOpDescriptor {
 }
 
 #[derive(Debug)]
-struct UringDescriptor {
+pub(crate) struct UringDescriptor {
```
glommer, line 95: I don't see where this is being used, but can we make do with `pub(super)`?

glommio/src/sys/uring.rs
```diff
@@ -978,6 +978,49 @@ impl SleepableRing {
         source
     }
 
+    /// This function prepares a timer that fires only after a given number of
```
glommer, line 981: please update the commit message benchmark when you push this again.

glommer (changes requested):

glommio/src/reactor.rs
HippoBaro: That makes sense, I changed the behavior to what you suggested. I also added `connect/accept` to the latency ring as well, as those are also important for network operations.

glommio/src/sys/uring.rs
HippoBaro: `UringDescriptor` is now part of the `UringCommin` trait and since that one is `pub(crate)`, this one needs to be as well. 
At some poi not it would be nice to break that trait apart

Issue #482, opened 2021-12-21
RwLock panics inside implementation of Deref/DerefMut traits


laa:
According to the contract of Deref trait - "For similar reasons, this trait should never fail. Failure during dereferencing can be extremely confusing when Deref is invoked implicitly." https://doc.rust-lang.org/std/ops/trait.Deref.html
Because RwLockGuards fail inside of Deref implementation that probably should be fixed.

glommer:
Thanks for pointing this out @laa!

Are you planning on providing a fix ? You don't have to, but that'd be cool. Just let us know for planning

laa:
Hi @glommer .
Sure will provide a fix in a couple of weeks.

P.S. Glad to interact with you after a while  :-)

laa:
Hi @glommer .
I have checked the code, I suppose we should panic on close. It will panic if readers/writers are still active. WDYT ?
I suppose better would be to return a result on close instead, but it will change the API of the library. IMHO this solution is better. 
WDYT ?

glommer:
Either panic or close, or maybe return an error at close.
Users will then be forced to check it due to `must_check` rules and can retry, etc.

bryandmc:
This is fixed by #510

PR #510, opened 2022-01-18
 Method RwLock#close returns error if some fibres still hold locks.

laa:
### What does this PR do?

 Method RwLock#close returns error if some fibres still hold locks.

### Motivation

This change is done to remove panic from Deref implementations of RwLockRead(Write)Guards if RwLock is already closed.

### Related issues

Issue #482

### Checklist

### DIFF
--- a/glommio/src/error.rs
+++ b/glommio/src/error.rs
@@ -218,6 +218,10 @@ pub enum GlommioError<T> {
     /// variant needs to return the actual item sent into the channel.
     Closed(ResourceType<T>),
 
+    /// The resource can not be closed because certain conditions are not
+    /// satisfied
+    CanNotBeClosed(ResourceType<T>, &'static str),
+
     /// Error encapsulating the `WouldBlock` error for types that don't have
     /// errors originating in the standard library. Glommio also has
     /// nonblocking types that need to indicate if they are blocking or not.
@@ -254,7 +258,7 @@ impl<T> fmt::Display for GlommioError<T> {
                 source, op, path, fd
             ),
             GlommioError::ExecutorError(err) => write!(f, "Executor error: {}", err),
-            GlommioError::BuilderError(err) => write!(f, "Executror builder error: {}", err),
+            GlommioError::BuilderError(err) => write!(f, "Executor builder error: {}", err),
             GlommioError::Closed(rt) => match rt {
                 ResourceType::Semaphore {
                     requested,
@@ -270,6 +274,11 @@ impl<T> fmt::Display for GlommioError<T> {
                 ResourceType::File(msg) => write!(f, "File is closed ({})", msg),
                 ResourceType::Gate => write!(f, "Gate is closed"),
             },
+            GlommioError::CanNotBeClosed(_, s) => write!(
+                f,
+                "Can not be closed because certain conditions are not satisfied. {}",
+                *s
+            ),
             GlommioError::WouldBlock(rt) => match rt {
                 ResourceType::Semaphore {
                     requested,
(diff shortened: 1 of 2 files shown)


glommer:
Hey @laa this looks good to me, but I'd like @bryandmc to take a look (mostly for the errors) if he can.

closing returning an error is a good way to do it.

Also, I think tests are not running on this branch because it is out of sync with the master branch. In which case, you'll have to rebase it.

laa:
Thank you @glommer , looking forward to checking the results of the review. I regret a bit that I paused participation in such an interesting project. But after all, I am back again :-)

bryandmc:
Glad you are back @laa ! I'll take a look this evening PST.

glommio/src/error.rs
```diff
@@ -254,7 +258,7 @@ impl<T> fmt::Display for GlommioError<T> {
                 source, op, path, fd
             ),
             GlommioError::ExecutorError(err) => write!(f, "Executor error: {}", err),
-            GlommioError::BuilderError(err) => write!(f, "Executror builder error: {}", err),
+            GlommioError::BuilderError(err) => write!(f, "Executor builder error: {}", err),
```
bryandmc, line 261: nice catch

glommio/src/sync/rwlock.rs
```diff
@@ -763,17 +757,23 @@ impl<T> RwLock<T> {
...
         let mut state = self.state.borrow_mut();
         if state.closed {
-            return;
+            return Ok(());
+        }
+
+        if state.writers > 0 || state.readers > 0 {
+            return Err(GlommioError::CanNotBeClosed(
+                ResourceType::RwLock,
+                "Lock is still hold by fibers",
```
bryandmc, line 769: *"held" and also maybe have it say "fiber(s)" so that it's clear it also might be a single one

glommio/src/sync/rwlock.rs
```diff
@@ -917,7 +917,7 @@ impl<T: Default> Default for RwLock<T> {
 
 impl<T> Drop for RwLock<T> {
     fn drop(&mut self) {
-        self.close();
+        self.close().unwrap();
```
bryandmc, line 920: Isn't this problematic?

glommio/src/error.rs
```diff
@@ -473,6 +498,9 @@ impl<T> From<GlommioError<T>> for io::Error {
             GlommioError::IoError(io_err) => io_err,
             GlommioError::WouldBlock(_) => io::Error::new(io::ErrorKind::WouldBlock, display_err),
             GlommioError::Closed(_) => io::Error::new(io::ErrorKind::BrokenPipe, display_err),
+            GlommioError::CanNotBeClosed(_, _) => {
+                io::Error::new(io::ErrorKind::BrokenPipe, display_err)
```
bryandmc, line 502: I don't know if this is the correct ErrorKind but I also don't have a better one in mind.. The thing about broken pipe is it seems to usually be an unfixable situation. While if we can't close this lock now, it might be close-able later. But maybe this is the best we can do.. I'll have to look at all the variants later.

bryandmc (commented):
Looks good for the most part. I think there are a few small issues, but they are easy. Also this is reminding me that we might need to either rework some of these error types or make a macro to simplify things. That's not your problem, though haha

glommio/src/sync/rwlock.rs
laa: No, Guard can not outlive the lock itself (lifetime annotation prohibits that), so at this moment all locks are bound to be freed. I will add a comment to this line.

laa:
@bryandmc I have pushed fixes for observations and waiting for your suggestion about the type of io::Error that better fits for conversion.

glommio/src/error.rs
bryandmc: What do you think about using "ResourceBusy" for this? I think it makes some sense..

laa: Hi, @bryandmc ResourceBusy is part of an unstable build. Do not think it could be used. Maybe you have other options in your list?

bryandmc:
@laa you see my comment on the error type? it was in the code comment.

laa:
> @laa you see my comment on the error type? it was in the code comment.

@bryandmc yes and I answered you there.

glommio/src/error.rs
bryandmc: I can't believe I didn't see this! I feel silly haha ok after looking again, not in my editor on nightly, I think this might be our only decent choice actually. Ok I think we are stuck with this until better ones get stabilized.

bryandmc (approved):

bryandmc:
Looks like a formatting issue.. I can also fix it if you don't see this before I'm off work later today. So whoever gets to it faster ;-p

laa:
@bryandmc there is a problem in the doc of the close method. I fix it alongside with fix of formatting.

PR #511, opened 2022-01-18
TcpListener: support FromRawFd

jmesmon:
### What does this PR do?

UdpSocket, TcpStream, UnixStream, etc, provide FromRawFd already, but
TcpListener lacks it. Add FromRawFd to TcpListener to allow custom
sources for TcpListeners to be used (for example, systemd-style socket
activation and fds sent over unix sockets).

### Motivation

I wanted to have systemd socket activation work with a program using glommio

### Additional Notes

 - I've chosen not to set the listen queue or any flags on the fd here under the theory that the user, if they are passing a raw fd, has likely already configured them (for example, when using systemd's socket activation, I configure the listen queue in the .socket unit file)
 - It doesn't look like there are any existing unit tests for any FromRawFd impls. I'm not sure exactly what the pattern should be for unit testing this.
 - Separately, it could also be useful to have `IntoRawFd` and `AsRawFd` to allow users to get fds out of glommio if needed for some reason
 - I've exposed only FromRawFd in this PR, but it might make sense to provide `From` impls for `std::net::*` and/or `socket2` types as they would allow applications needed these conversions to avoid using unsafe code (calling `from_raw_fd()` is unsafe).

### Checklist

### DIFF
--- a/glommio/src/net/tcp_socket.rs
+++ b/glommio/src/net/tcp_socket.rs
@@ -77,6 +77,19 @@ pub struct TcpListener {
     listener: net::TcpListener,
 }
 
+impl FromRawFd for TcpListener {
+    /// Convert an already bound and listening RawFd into a TcpListener
+    unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        let sk = Socket::from_raw_fd(fd);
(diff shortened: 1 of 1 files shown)


glommer (approved):
Thanks!

PR #512, opened 2022-01-19
force submission of important IO requests

HippoBaro:
So far, we have been pretty lax when submitting stuff in
the rings. We push a request in there, flush, and if it fails, we keep
going.

This works for user requests that can wait until the next visit to the
reactor but can cause problems for those requests we depend on. One such
request is the preemption timer.

This commit adds a force flush logic that only returns if it has
successfully submitted a specific number of events to the kernel and
will even block as a last resort if it has to.

This replaces the old cancellation flushing logic that would poll the
kernel for one million times and panic afterward.

Additionally, this fixes a bug that would leave cancellations in the
queue, unsubmitted, if the number of cancellations exceeds the SQE ring
capacity (which is pretty small at ~128 typically).

### DIFF
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -670,6 +670,7 @@ pub(crate) trait UringCommon {
     /// Return `None` if no event is completed, `Some(true)` for a task is woken
     /// up and `Some(false)` for not.
     fn consume_one_event(&mut self) -> Option<bool>;
+    fn wait_for_events(&mut self, events: u32) -> io::Result<()>;
     fn name(&self) -> &'static str;
     fn io_stats_mut(&mut self) -> &mut RingIoStats;
     fn io_stats_for_task_queue_mut(&mut self, handle: TaskQueueHandle) -> &mut RingIoStats;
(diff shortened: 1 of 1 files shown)


HippoBaro:
For reference, this PR is motivated by the fact that I have an app that consistently panics at the following assertion: https://github.com/DataDog/glommio/blob/master/glommio/src/sys/uring.rs#L1721

glommio/src/sys/uring.rs
```diff
@@ -731,29 +732,56 @@ trait UringCommon {
...
+                        );
+                        self.wait_for_events((min_requests - submitted) as u32)
+                            .expect("failed to wait for CQE. Game over");
+                        consumed = false;
+                    }
+                    self.consume_completion_queue(woke);
+                }
+                Ok(x) => submitted += x,
             }
-            self.consume_completion_queue(woke);
```
glommer, line 747: I feel that this still deserves at least a `warn!()`. If you have executed this loop 1MM times... at the very least you have a giant performance problem on your hands and I bet you'd love to know where it came from

glommio/src/sys/uring.rs
```diff
@@ -676,6 +676,7 @@ trait UringCommon {
     /// Return `None` if no event is completed, `Some(true)` for a task is woken
     /// up and `Some(false)` for not.
     fn consume_one_event(&mut self) -> Option<bool>;
+    fn wait_for_events(&mut self, events: u32) -> io::Result<()>;
```
glommer, line 679: if we had a bug about cancellations not being sent, it would be good if this came with an accompanying test.

glommio/src/sys/uring.rs
HippoBaro: See the warning here https://github.com/DataDog/glommio/pull/512/files#diff-d8d2d815ed457d48361ae92c5976d152059d3f04d59dea4ad148d1ea8a75f21aR761-R768

glommio/src/sys/uring.rs
HippoBaro: I've tried to write a test that reliably (and cheaply) reproduces the bug for a few hours, but I've not been successful. 

The kernel may apply back-pressure and refuse our SQE submissions if it fails to allocate or if the CQE ring is full. If that happens before we submit the preemption timer, we panic because of a failed assertion (before this patch). 

Because I don't have control over the kernel, it's difficult for me to recreate those conditions.

PR #513, opened 2022-01-21
Fix of Clippy observations about unsound code generated by intrusive collections macro

laa:
### What does this PR do?
A new macro for implementation of `Adapter` trait required by intrusive collections instances was introduced. 
Unlike the standard macro , its `Adapter` implementation does not implement `Send` and `Sync` traits, that fixes clippy 
warnings and makes code sound.

### Motivation

Fix of generated unsound code.

### Checklist

### DIFF
--- a/glommio/src/intrusive_collections_ext.rs
+++ b/glommio/src/intrusive_collections_ext.rs
@@ -1,0 +1,165 @@
+/// Macro to generate an implementation of `Adapter` from
+/// `intrusive_collections` crate for a given set of types.
+/// In particular this will automatically generate implementations of the
+/// `get_value` and `get_link` methods for a given named field in a struct.
+///
+/// Unlike the macro provided by `intrusive_collections`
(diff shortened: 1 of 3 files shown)


glommer (approved):

laa:
If my PR is approved may someone merge it into the branch?

glommer:
Hi @laa I'd like to understand first if you plan to push this to their project. If you do, unless this is very urgent for some reason it is best to wait a couple of days to see if it gets merged. So for example, if the name changes slightly we don't have to patch this later in glommio.

HippoBaro:
I agree, let's pursue upstreaming this change. It seems like they could implement `?Send` instead of `Send` and it would just work. Same principle where a `Box<T>` is `Send+Sync` only if `T:  Send + Sync`

laa:
Hi, @HippoBaro that is a good idea. Will do that.

PR #514, opened 2022-01-21
Update `OpenOptions::append` to return `&mut Self`

r3v2d0g:
### What does this PR do?

It updates `OpenOptions::append` to return `&mut Self`.

### Motivation

N/A

### Related issues

N/A


glommer (approved):
thanks!

PR #515, opened 2022-01-25
gracefully handle `SIGINT` signals while sleeping

HippoBaro:
A thread may be sleeping while other threads in the system deal with
signals. If that happens, gracefully wake up and loop back to the
reactor.

Co-authored-by: David Blewett <david.blewett@datadoghq.com>


davidblewett (approved):
LGTM!

HippoBaro:
Tests are passing modulo unrelated Clippy warnings that we need to address (they are new from 1.58).

PR #519, opened 2022-01-30
use an asynchronous queue to submit blocking operations

HippoBaro:
We now use a bounded channel that yields to the scheduler when it is
full. This is desirable because we want to apply backpressure to
user-code in those cases rather than to allocate or block.


HippoBaro:
Of note is that this back-pressure mechanism doesn't exist for regular requests (i.e., the one going through the ring). A user can create many requests, which will pile up in the reactor queues. 

It would be nice to have the same back-pressure there as well.

bryandmc (approved):
Looks good. I had to look into the flume crate and was amazed that it's substantially better than crossbeam with a much simpler design. Good find. Also much easier that it's just a bounded channel and already works async.

HippoBaro:
> Looks good. I had to look into the flume crate and was amazed that it's substantially better than crossbeam with a much simpler design. Good find. Also much easier that it's just a bounded channel and already works async.

Totally! I'm considering replacing crossbeam with this one to avoid having two dependencies doing the same thing.

PR #520, opened 2022-01-30
small reactor improvements

HippoBaro:
See individual commits

### DIFF
--- a/glommio/src/free_list.rs
+++ b/glommio/src/free_list.rs
@@ -47,7 +47,7 @@ impl<T> Default for FreeList<T> {
     fn default() -> Self {
         FreeList {
             first_free: None,
-            slots: Vec::new(),
+            slots: Vec::with_capacity(8 << 10),
         }
     }
 }
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -662,6 +662,7 @@ pub(crate) trait UringCommon {
     fn submission_queue(&mut self) -> ReactorQueue;
     fn submit_sqes(&mut self) -> io::Result<usize>;
     fn waiting_kernel_submission(&self) -> usize;
+    fn in_kernel(&self) -> usize;
     fn waiting_kernel_collection(&self) -> usize;
     fn needs_kernel_enter(&self) -> bool;
     fn can_sleep(&self) -> bool;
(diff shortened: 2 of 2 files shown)


glommio/src/free_list.rs
```diff
@@ -47,7 +47,7 @@ impl<T> Default for FreeList<T> {
     fn default() -> Self {
         FreeList {
             first_free: None,
-            slots: Vec::new(),
+            slots: Vec::with_capacity(8 << 10),
```
duarten, line 50: Why not `1 << 13`?

duarten (approved):

glommio/src/free_list.rs
glommer: I personally find `8 << 10` way more readable. I am a bit dumb so I have to do math in my head to remember how much 13 shift is. It's not *that* hard, because from my Linux days I had "12" memorized (page size), but still.
10 is a kB, so "8 << 10" immediately evokes 8kB. Of course this is all subjective, because I had to memorize that 10 is a kB as well, and I can say all day this is "more natural", but at the end of the day that obviously means nothing.

glommio/src/sys/source.rs
```diff
@@ -323,6 +323,7 @@ impl Source {
 impl Drop for Source {
     fn drop(&mut self) {
         let mut inner = self.inner.borrow_mut();
+        inner.wakers.waiters.clear();
```
glommer, line 326: that's the kind of things bugs are made of. "No one is interested" until they are, and dropping the source was an accident.

How often did you find pending sources at this point ? I'd be interested in understanding why they happen in the first place.

glommio/src/free_list.rs
HippoBaro: Ahaha I'm the same as @glommer and this was going to be my response. I'm fine changing this to `1 << 13` if you feel strongly about it. Eventually, I'll remember what value it ends up being :laughing:

glommio/src/sys/source.rs
HippoBaro: Btw I don't feel strongly about this and I'd be happy to drop that commit if you're concerned about it. My point, though, is that if this is causing trouble, then it would be indicative of a (probably worse) underlying issue.

glommio/src/free_list.rs
duarten: I don't feel strongly about it at all, I was just asking out of curiosity :)

glommio/src/sys/source.rs
glommer: I agree that this is indicative of a bigger problem, the main issue is how things manifest.
Things silently not executing is certainly not fun to debug - as you know.

My suggested strategy with this would be:
1. Print the existing wakers, if any, first.
2. Figure out why those wakers exist and try, if possible, to patch the callers not to have them.
3. Assert that there are no more wakers at drop, at least in debug mode.

PR #521, opened 2022-01-30
add `read_many_unordered`

HippoBaro:
Sometimes order doesn't matter, and an application can consume the IO
results in random order. This allows the stream to drive more IO
concurrency than the ordered variant.

In a pathological case, the ordered stream may have completed all IO
requests except the next one to yield. In that situation, the IO
concurrency of the stream drops to 1 because we need the first IO to
come back before we can yield anything and make room to schedule more IO
requests.

This may seem like it shouldn't happen but ywo factors play against us:
 * The kernel doesn't order anything and,
 * Devices don't order anything and,
 * Request coalescing may produce requests of varying sizes. Large IO 
 requests take much longer to complete.

An unordered stream will yield the IO results as soon as they arrive so
that the IO concurrency can remain maximal at all times.

### DIFF
--- a/glommio/src/io/bulk_io.rs
+++ b/glommio/src/io/bulk_io.rs
@@ -298,6 +298,36 @@ impl<V: IoVec + Unpin, S: Stream<Item = V> + Unpin> Stream for CoalescedReads<V,
     }
 }
 
+/// A trait that limits the number of IO concurrency of [`BulkIo`] streams
+///
+/// Multiple concurrency limits are available and exist in parallel: whichever
+/// limit triggers first wins.
+pub trait IoConcurrencyLimiter {
+    /// Sets the maximum concurrency of the stream in terms of number of
+    /// inflight IO requests. A request is inflight if it is pending IO
+    /// completion or if it is waiting for the user to consume the stream.
+    fn set_concurrency_limit(&mut self, limit: usize);
+
+    /// Sets the maximum concurrency of the stream in terms of memory
+    /// consumption. Usage correspond to the amount of IO buffers the
+    /// stream has a reference to.
+    fn set_memory_limit(&mut self, limit: Option<usize>);
+
+    /// Whether this stream has reached maximum concurrency
+    ///
+    /// If true, this stream will not schedule any more IO until some IO results
+    /// have been successfully consumed
+    fn is_full(&self) -> bool;
+}
+
+/// A trait for streams that perform many IO requests concurrently
+pub trait BulkIo<U: IoVec + Unpin>:
+    Stream<Item = (ScheduledSource, U)> + Unpin + IoConcurrencyLimiter
+{
+    /// The file behind this stream
+    fn file(&self) -> &Rc<DmaFile>;
+}
+
 #[derive(Debug)]
 pub(crate) struct OrderedBulkIo<U: IoVec + Unpin, S: Stream<Item = (ScheduledSource, U)> + Unpin> {
     file: Rc<DmaFile>,
@@ -323,8 +353,12 @@ impl<U: IoVec + Unpin, S: Stream<Item = (ScheduledSource, U)> + Unpin> OrderedBu
             terminated: false,
         }
     }
+}
 
-    pub(crate) fn set_concurrency_limit(&mut self, limit: usize) {
+impl<U: IoVec + Unpin, S: Stream<Item = (ScheduledSource, U)> + Unpin> IoConcurrencyLimiter
+    for OrderedBulkIo<U, S>
+{
+    fn set_concurrency_limit(&mut self, limit: usize) {
         assert!(limit > 0);
         assert!(
             !self.terminated && self.inflight.is_empty(),
@@ -334,7 +368,7 @@ impl<U: IoVec + Unpin, S: Stream<Item = (ScheduledSource, U)> + Unpin> OrderedBu
         self.concurrency_cap = limit
     }
 
-    pub(crate) fn set_memory_limit(&mut self, limit: Option<usize>) {
+    fn set_memory_limit(&mut self, limit: Option<usize>) {
         assert!(limit.unwrap_or(usize::MAX) > 0);
         assert!(
             !self.terminated && self.inflight.is_empty(),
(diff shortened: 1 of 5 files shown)


dignifiedquire:
Hey, I wanted to test this PR out, but ran into an issue, I can't pass around a `ReadManyResult` because `ReadManyArgs` seems to be a private trait, and as such, I can't write the actual trait bounds out. Any chance that could be fixed, or am I missing something?

HippoBaro:
Hey @dignifiedquire! 

Can you post an example I can compile on my end to reproduce?

dignifiedquire:
sure, sth like this

```rust
use futures::stream;
use glommio::io::{ReadManyResult, BulkIo, DmaFile, ReadAmplificationLimit, MergedBufferLimit};

async fn read_things(
	offsets: &[u64],
    sizes: &[usize],
    file: &Rc<DmaFile>,
) -> ReadManyResult<(u64, usize), impl BulkIo<XX>> { // << the XX is the part I can't fill in to get it to compile
    let iovs = stream::iter(offsets.iter().zip(sizes.iter()).map(|(o, s)| (*o, *s)));
    file.read_many(previous_layer.read_many(
        iovs,
        MergedBufferLimit::DeviceMaxSingleRequest,
        ReadAmplificationLimit::NoAmplification,
    ).await
```

HippoBaro:
Oh, I see what's going on. The issue is that this symbol you don't have access to itself wraps the stream you provide us. 

Instead, you should treat it like an unnamable entity. This is the same principle that applies to your `iovs` stream: try to return it and you won't be able to for the same reason. Fortunately, there is a way around this:

```rust
use futures::stream;
use glommio::io::{ReadManyResult, BulkIo, DmaFile, ReadAmplificationLimit, MergedBufferLimit};

async fn read_things(
	offsets: &[u64],
    sizes: &[usize],
    file: &Rc<DmaFile>,
) -> impl Stream<Item = ((u64, usize), ReadResult)> { // << we are now returning an impl Trait 
    let iovs = stream::iter(offsets.iter().zip(sizes.iter()).map(|(o, s)| (*o, *s)));
    file.read_many(previous_layer.read_many(
        iovs,
        MergedBufferLimit::DeviceMaxSingleRequest,
        ReadAmplificationLimit::NoAmplification,
    ).await
```

Let me know if that helps!

dignifiedquire:
On a positive note, I have tested this, and have seen speedups of up to 90% between `read_many` and `read_many_unordered` for small reads.

HippoBaro:
> On a positive note, I have tested this, and have seen speedups of up to 90% between read_many and read_many_unordered for small reads.

That's awesome; thanks for the feedback! 

> Small side note, this does mean I won't be able to put a concrete type anywhere if I end up storing this in another structure because I need to wrap it potentially.

The issue is that, because we take inputs as a stream (although the issue would be the same with an iterator), we have the same problem in Glommio. We can't provide a concrete type unless we use `dyn` internally, which we won't because of the performance implication. 

Therefore, if you need to store the `read_many` stream on your end, you'll have to either use generics or `dyn`. AFAIK there is no other way, but let me know if I'm missing something.

dignifiedquire:
> AFAIK there is no other way, but let me know if I'm missing something.

There is, it is just a little bit more lines of code, so not always worth the tradeoffs: https://github.com/dignifiedquire/glommio/commit/e014bc5f677c78611c9af275d592c96d7fbf1802

HippoBaro:
I looked at the code and I'm not sure what that buys us (the new type you add is also templated with the input stream). 
Can you write a small example that would work with the example you posted earlier? Seems to me that, because the `iovs` stream cannot be named, you'll end up in the same situation.

dignifiedquire:
It is not pretty, but it does compile

```rust
struct MyReadResult {
    res: ReadManyResult<
        (u64, usize),
        UnorderedBulkIo<
            ReadManyArgs<(u64, usize)>,
            CoalescedReadsMapped<
                (u64, usize),
                stream::Iter<std::array::IntoIter<(u64, usize), 1>>,
            >,
        >,
    >,
}
async fn test_read_struct() {
    let file = Rc::new(DmaFile::open("foo").await.unwrap());

    let res = file.read_many_unordered(
        stream::iter([(0, 10)]),
        MergedBufferLimit::NoMerging,
        ReadAmplificationLimit::NoAmplification,
    );

    let x: MyReadResult = MyReadResult { res };
}
```

The nice thing about this, while being very verbose, is that the end consumer of the library does get the ability to fully name all involved types when they want/need to.

HippoBaro:
Right, I see what you mean. Here is what I understand so far:
* The original version of the code takes the input stream and refers to it using `impl Stream<....>`, making it de facto unnamelable.
* Your version doesn't, so if the user provides a named Stream, they can use the ReadMany stream as you did above.

In a realistic case, most of the streams people will give us won't be namable (because it would make the code somewhat fragile to any change). Nonetheless, your solution is superior.  

@dignifiedquire Would you be interested in upstreaming your code? We may need to rework it a little bit, though. Let me know if I can help!

dignifiedquire:
sure, I can make a PR

Max-Meldrum:
I'm very interested in this feature. Currently using it on a fork together with  #529. 

Would love to help get it merged if there is something left remaining.

Also, wondering whether the inflight data structure can be a HashMap instead of a BTreeMap as things are not ordered?
```rust
inflight: BTreeMap<u64, (ScheduledSource, U)>
```

glommer:
@HippoBaro I see you updated this. Thanks. But tests are failing.

glommio/src/io/bulk_io.rs
```diff
@@ -417,6 +522,145 @@ impl<U: IoVec + Unpin, S: Stream<Item = (ScheduledSource, U)> + Unpin> Stream
...
+        this.waker.update(cx.waker().clone());
+
+        // Poll the underlying stream and insert the resulting source, if any, in the
+        // local buffer
+        let poll_inner = |this: &mut Self, cx: &mut Context<'_>| match this.iovs.poll_next(cx) {
+            Poll::Ready(Some(res)) => {
+                this.inflight_memory += res.1.size();
+                let (id, proxy_waker) = this.waker.proxy();
+                res.0.add_waiter_many(proxy_waker);
+                this.inflight.insert(id, res);
```
Max-Meldrum, line 608: Running into some issues where unordered version only returns 1 ready item.
Might be related to the failing test.  

The waker id that is returned in ``poll_inner`` is always the same,  and the inflight map will keep having only 1 entry inside it and then return ``Poll::Ready(None)`` at a later point.

Currently trying to debug it more.

HippoBaro: Thanks, @Max-Meldrum! I rebased it this weekend and planned to give it a look today. Let us know here if you find the culprit.

Max-Meldrum: Have not gotten around to implementing any fix yet, but the main issue is that were using Rc::as_ptr() as id when there is only a single ProxyWaker instance. that is, always the same pointer address used as id.

```rust
// won't change as its referring to a single instance of Rc<ProxyWaker>
let id = Rc::as_ptr(&rc) as u64;
```

PR #527, opened 2022-02-12
offer the option to use native TLS LLVM support

HippoBaro:
I have been waiting for this to land in stable forever, and my patience
is running out. So we'll start supporting this behind a feature gate
that's disabled by default (native TLS requires a nightly compiler).

Additionally, this PR gets rid of `cond_yield` in favor of manually
inlining the preemption code. I tried to get the closure passed to
`cond_yield` to be inlined, but the compiler won't cooperate. This is
such a hot code path that it makes sense to inline it by hand.

And the results are great:

### With std-tls (default):
Before (no inlining):
```
cost of checking for need_preempt:   2ns (ran 1000000 times)
cost of calling yield_if_needed:     5ns (ran 1000000 times)
```
After (inlined):
```
cost of checking for need_preempt:   1ns (ran 1000000 times)
cost of calling yield_if_needed:     3ns (ran 1000000 times)
```

### With native-tls (behind the feature gate):
Before (no inlining):
```
cost of checking for need_preempt:   1ns (ran 1000000 times)
cost of calling yield_if_needed:     3ns (ran 1000000 times)
```
After  (inlined):
```
cost of checking for need_preempt:   0ns (ran 1000000 times)
cost of calling yield_if_needed:     2ns (ran 1000000 times)
```

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -86,11 +86,11 @@ pub(crate) const DEFAULT_RING_SUBMISSION_DEPTH: usize = 128;
 /// the second type parameter.
 type Result<T> = crate::Result<T, ()>;
 
-#[cfg(feature = "native-tls")]
+#[cfg(feature = "nightly")]
 #[thread_local]
 static mut LOCAL_EX: *const LocalExecutor = std::ptr::null();
 
-#[cfg(not(feature = "native-tls"))]
+#[cfg(not(feature = "nightly"))]
 scoped_tls::scoped_thread_local!(static LOCAL_EX: LocalExecutor);
 
 /// Returns a proxy struct to the [`LocalExecutor`]
@@ -100,7 +100,7 @@ pub fn executor() -> ExecutorProxy {
 }
 
 pub(crate) fn executor_id() -> Option<usize> {
-    #[cfg(not(feature = "native-tls"))]
+    #[cfg(not(feature = "nightly"))]
     {
         if LOCAL_EX.is_set() {
             Some(LOCAL_EX.with(|ex| ex.id))
@@ -109,7 +109,7 @@ pub(crate) fn executor_id() -> Option<usize> {
         }
     }
 
-    #[cfg(feature = "native-tls")]
+    #[cfg(feature = "nightly")]
     unsafe {
         LOCAL_EX.as_ref().map(|ex| ex.id)
     }
@@ -1064,13 +1064,13 @@ impl<T> PoolThreadHandles<T> {
 }
 
 pub(crate) fn maybe_activate(tq: Rc<RefCell<TaskQueue>>) {
-    #[cfg(not(feature = "native-tls"))]
+    #[cfg(not(feature = "nightly"))]
     LOCAL_EX.with(|local_ex| {
         let mut queues = local_ex.queues.borrow_mut();
         queues.maybe_activate(tq);
     });
 
-    #[cfg(feature = "native-tls")]
+    #[cfg(feature = "nightly")]
     unsafe {
         let mut queues = LOCAL_EX
             .as_ref()
@@ -1513,7 +1513,7 @@ impl LocalExecutor {
             }
         };
 
-        #[cfg(not(feature = "native-tls"))]
+        #[cfg(not(feature = "nightly"))]
         {
             assert!(
                 !LOCAL_EX.is_set(),
@@ -1522,7 +1522,7 @@ impl LocalExecutor {
             LOCAL_EX.set(self, || run(self))
         }
 
-        #[cfg(feature = "native-tls")]
+        #[cfg(feature = "nightly")]
         unsafe {
             assert!(
                 LOCAL_EX.is_null(),
@@ -2111,10 +2111,10 @@ impl ExecutorProxy {
     /// [`RefMut`]: https://doc.rust-lang.org/std/cell/struct.RefMut.html
     #[inline(always)]
     pub fn need_preempt(&self) -> bool {
-        #[cfg(not(feature = "native-tls"))]
+        #[cfg(not(feature = "nightly"))]
         return LOCAL_EX.with(|local_ex| local_ex.need_preempt());
 
-        #[cfg(feature = "native-tls")]
+        #[cfg(feature = "nightly")]
         return unsafe {
             LOCAL_EX
                 .as_ref()
@@ -2140,7 +2140,7 @@ impl ExecutorProxy {
     /// [`ExecutorProxy::yield_task_queue_now`].
     #[inline(always)]
     pub async fn yield_if_needed(&self) {
-        #[cfg(not(feature = "native-tls"))]
+        #[cfg(not(feature = "nightly"))]
         {
             let need_yield = if LOCAL_EX.is_set() {
                 LOCAL_EX.with(|local_ex| {
@@ -2161,7 +2161,7 @@ impl ExecutorProxy {
             }
         }
 
-        #[cfg(feature = "native-tls")]
+        #[cfg(feature = "nightly")]
         unsafe {
             if self.need_preempt() {
                 (*LOCAL_EX).mark_me_for_yield();
@@ -2189,7 +2189,7 @@ impl ExecutorProxy {
     /// [`ExecutorProxy::yield_if_needed`] instead is the better choice.
     #[inline(always)]
     pub async fn yield_task_queue_now(&self) {
-        #[cfg(not(feature = "native-tls"))]
+        #[cfg(not(feature = "nightly"))]
         {
             if LOCAL_EX.is_set() {
                 LOCAL_EX.with(|local_ex| {
@@ -2199,7 +2199,7 @@ impl ExecutorProxy {
             futures_lite::future::yield_now().await;
         }
 
-        #[cfg(feature = "native-tls")]
+        #[cfg(feature = "nightly")]
         {
             if let Some(local_ex) = unsafe { LOCAL_EX.as_ref() } {
                 local_ex.mark_me_for_yield();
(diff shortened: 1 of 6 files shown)


HippoBaro:
@glommer This is something we've been meaning to look into for a while now. Let me know if you are happy with the implementation. 

One thing that's problematic ATM is that our Clippy in CI runs with `--all features`. Because Clippy isn't on nightly, Clippy fails.
I don't like the fact that this PR makes the `--all-features`  cargo flag broken on `stable`. I'll try to think of something :shrug:

glommio/src/executor/mod.rs
```diff
@@ -2091,30 +2110,17 @@ impl ExecutorProxy {
...
     pub fn need_preempt(&self) -> bool {
-        LOCAL_EX.with(|local_ex| local_ex.need_preempt())
+        #[cfg(not(feature = "native-tls"))]
+        return LOCAL_EX.with(|local_ex| local_ex.need_preempt());
+
+        #[cfg(feature = "native-tls")]
+        return unsafe {
+            LOCAL_EX
+                .as_ref()
+                .expect("this thread doesn't have a LocalExecutor running")
```
glommer, line 2121: I'd be interested in seeing what kind of code we generate.
There's no need to panic here if the pointer is null, just return false.
That could likely lead to a more efficient implementation since there's no need to inline the panic handler

glommio/src/executor/mod.rs
```diff
@@ -2132,9 +2138,38 @@ impl ExecutorProxy {
...
+            };
+
+            if need_yield {
+                futures_lite::future::yield_now().await;
+            }
+        }
+
+        #[cfg(feature = "native-tls")]
+        unsafe {
+            if !LOCAL_EX.is_null() {
```
glommer, line 2166: and if you tackle the above to just return false when the executor is null, then you don't need this test here

glommio/src/executor/mod.rs
```diff
@@ -1490,7 +1511,28 @@ impl LocalExecutor {
...
+        }
+
+        #[cfg(feature = "native-tls")]
+        unsafe {
+            assert!(
+                LOCAL_EX.is_null(),
+                "There is already an LocalExecutor running on this thread"
+            );
+
+            defer!(LOCAL_EX = std::ptr::null());
```
glommer, line 1532: I am concerned with a destruction race here.
When we call drop on the executor, we will then call drop on its task queues, which then call drop on tasks, etc.

Some of that code in drop may want to reference the reactor to issue async cancellations and the like, and now that will be null.

So I think the safest way would be for us to destroy the executor manually, and only then make it null.

If you think that couldn't happen, we'd have to prove that this is the case and write a big comment. It should be relatively easy to test too

glommer:
By "behind a feature gate" do you mean that people using this would need use nightly ?

I obviously don't mind allowing people to use nightly.
the code looks uglier now, but that's ultimately okay, with cond features there's no much of a choice.

The code looks okay, it's just a lot of busy work to convert the call sites. There are one place where I have a safety concern, and one where I am pointing out a minor performance improvement.

About CI, I looked a while ago at something similar and it seems to be possible to add conditions on the compiler through `cfg`, and then you can write a composite statement like `if nightly and feature_x` . But I don't remember how

glommer:
If this is standard in nightly and not gated, we can just depend on nightly for that feature, not the specific feature flag.

glommio/src/executor/mod.rs
HippoBaro: This code is written to exactly match the behavior of the other implementation (set the TLS var here and unset as soon as we return). The defer ensures we unset the tls var if we panic.

This was done this way because the std-tls are scoped and can't be freely accessed. This is not a limitation that I have with the native TLS, but I mirrored the behavior to avoid creating subtle differences in behavior between the two implementations.

So long story short this is not a problem and I couldn't implement your solution using the current TLS support on stable.

glommio/src/executor/mod.rs
HippoBaro: OIC; let me do that. 

Using std-tls:
```
glommio::executor::ExecutorProxy::need_preempt:
 push    rax
 lea     rdi, [rip, +, _ZN7glommio8executor8LOCAL_EX3FOO7__getit5__KEY17h9afbb68dc0c48f46E@TLSLD]
 call    __tls_get_addr
 cmp     dword, ptr, [rax, +, _ZN7glommio8executor8LOCAL_EX3FOO7__getit5__KEY17h9afbb68dc0c48f46E@DTPOFF], 1
 jne     .LBB640_2
 lea     rax, [rax, +, _ZN7glommio8executor8LOCAL_EX3FOO7__getit5__KEY17h9afbb68dc0c48f46E@DTPOFF+8]
 jmp     .LBB640_3
.LBB640_2:
 call    std::thread::local::fast::Key<T>::try_initialize
 test    rax, rax
 je      .LBB640_5
.LBB640_3:
 mov     rax, qword, ptr, [rax]
 test    rax, rax
 je      .LBB640_6
 mov     rax, qword, ptr, [rax, +, 24]
 mov     rcx, qword, ptr, [rax, +, 2600]
 mov     rax, qword, ptr, [rax, +, 2608]
 mov     eax, dword, ptr, [rax]
 cmp     eax, dword, ptr, [rcx]
 setne   al
 pop     rcx
 ret
.LBB640_6:
 lea     rdi, [rip, +, .L__unnamed_92]
 lea     rdx, [rip, +, .L__unnamed_93]
 mov     esi, 72
 call    std::panicking::begin_panic
 ud2
.LBB640_5:
 lea     rdi, [rip, +, .L__unnamed_98]
 lea     rcx, [rip, +, .L__unnamed_99]
 lea     r8, [rip, +, .L__unnamed_100]
 mov     rdx, rsp
 mov     esi, 70
 call    qword, ptr, [rip, +, _ZN4core6result13unwrap_failed17hf836a820f86fa4a9E@GOTPCREL]
 ud2
```

Using native-tls (as in the PR):
```
glommio::executor::ExecutorProxy::need_preempt:
 push    rax
 data16
 lea     rdi, [rip, +, _ZN7glommio8executor8LOCAL_EX17ha0f5da69d3f777eaE@TLSGD]
 data16
 data16
 rex64
 call    __tls_get_addr
 mov     rax, qword, ptr, [rax]
 test    rax, rax
 je      .LBB640_1
 mov     rax, qword, ptr, [rax, +, 24]
 mov     rcx, qword, ptr, [rax, +, 2600]
 mov     rax, qword, ptr, [rax, +, 2608]
 mov     eax, dword, ptr, [rax]
 cmp     eax, dword, ptr, [rcx]
 setne   al
 pop     rcx
 ret
.LBB640_1:
 lea     rdi, [rip, +, .L__unnamed_92]
 lea     rdx, [rip, +, .L__unnamed_639]
 mov     esi, 48
 call    qword, ptr, [rip, +, _ZN4core6option13expect_failed17h7e5fdafed144cc7fE@GOTPCREL]
 ud2
```

Using native-tls and your suggestion to return `false` instead of panicking
```
glommio::executor::ExecutorProxy::need_preempt:
 push    rax
 data16
 lea     rdi, [rip, +, _ZN7glommio8executor8LOCAL_EX17ha0f5da69d3f777eaE@TLSGD]
 data16
 data16
 rex64
 call    __tls_get_addr
 mov     rax, qword, ptr, [rax]
 test    rax, rax
 je      .LBB640_1
 mov     rax, qword, ptr, [rax, +, 24]
 mov     rcx, qword, ptr, [rax, +, 2600]
 mov     rax, qword, ptr, [rax, +, 2608]
 mov     eax, dword, ptr, [rax]
 cmp     eax, dword, ptr, [rcx]
 setne   cl
 jmp     .LBB640_3
.LBB640_1:
 mov     cl, 2
.LBB640_3:
 cmp     cl, 2
 setne   al
 and     al, cl
 pop     rcx
 ret
```

glommio/src/executor/mod.rs
HippoBaro: I did that too. I am not seeing any measurable improvement but it makes the code shorter so I'll take it.

HippoBaro:
> By "behind a feature gate" do you mean that people using this would need use nightly ?

No, I mean that you have to invoke cargo like so `cargo build --features=native-tls` to enable this. If you're on the stable channel, this will fail. 

Likewise, a downstream consumer of glommio can opt-in using 
```TOML
[dependencies]
glommio = {version = "0.7", features = [ "native-tls" ]}
```

About CI, I looked a while ago at something similar and it seems to be possible to add conditions on the compiler through cfg, and then you can write a composite statement like if nightly and feature_x . But I don't remember how

Right, I already do that. But if you invoke CI with the feature flag enabled and on the stable channel, then it won't work. 

> If this is standard in nightly and not gated, we can just depend on nightly for that feature, not the specific feature flag.

AFAIK there is no way in rust to detect at compile-time whether you are on nightly/beta/stable. This is on purpose because they don't want us to do that :stuck_out_tongue:

bryandmc:
> Ok, so it doesn't look like there is any reliable (and not hacky) way to automatically enable this code when building on nightly. Instead, I added a new `nightly` feature. We can now use this to use nightly-only feature in Glommio (less is more, please).
> 
> Note though that you still have to compile on nightly and enable the feature. Doing the latter and not the former will result in compilation errors.
> 
> I also have fixed CI on my branch but I don't think GH will use the newer CI files before they are merged so we are in a little bit of a chicken-and-egg problem here.

Yeah looks like its just not recommended:
https://stackoverflow.com/questions/59542378/conditional-compilation-for-nightly-vs-stable-rust-or-compiler-version

Something I have seen is to call the feature "unstable" or similar. If we foresee further nightly-only features we could make individual feature flags and then the meta-flag of "unstable". Either way I think that makes more sense than doing what the stack overflow suggests, which is basically a build.rs script.

bryandmc:
I think it matches the version feature strategy we discussed in the chat so I think it makes sense. If it's too much of a pain I think just leaving it all under the "nightly" feature could be fine too but just less flexible. Either of those I would be fine with, personally.

HippoBaro:
> I also was curious if you saw performance gains on higher level tests

I did! In my "competing io" benchmark I get this:

Before:
```
performed 100k read IO at 102k IOPS (took 0.98s)
[measured] CPU task ran at 190 Mops/s
```

After:
```
performed 100k read IO at 102k IOPS (took 0.98s)
[measured] CPU task ran at 345 Mops/s
```

This test creates two task queues: one performs IO, and the other does CPU computations. The latter task will try to yield all the time to maximize IO throughput. With native TLS, I get almost double CPU op/s.

HippoBaro:
@bryandmc The native TLS stuff is now gated behind the `native-tls` feature. I also added a `nightly` as a meta feature as we discussed.

Issue #531, opened 2022-02-20
Implement poll_based API for net


dignifiedquire:
I am running into  issues when trying to work with TCP and UDP transports, because they don’t implement poll_ methods. Especially wrapping them into streams becomes very complex and sometimes impossible, because it requires a poll_next implementation.

PR #533, opened 2022-02-21
net/tcp: Add poll_accept, poll_shared_accept and poll_shutdown

dignifiedquire:
### What does this PR do?

Adds 
- `TcpListener::poll_accept`,
- `TcpListener::poll_shared_accept` and
- `TcpStream::poll_shutdown` 

and internally adds `poll_collect_rw`.

### Motivation

The main methods I need right now for #531.


glommio/src/net/tcp_socket.rs
```diff
@@ -208,6 +233,19 @@ impl TcpListener {
         Ok(a.bind_to_executor())
     }
```
HippoBaro, line 234: I think we should use `poll_fn` here like elsewhere for the sake of consistency

HippoBaro (approved):
Some small nits, otherwise I am okay to merge

pbdeuchler:
Is this PR dead? I'm in need of a non-blocking way to accept connections on a `TcpListener` and I believe this PR solves my problem

vlovich:
@glommer this seems fine to merge (I don't need it but just working through the backlog). There's an outstanding nit where the poll_shutdown isn't as absolutely concise as it could be (neither is `poll_accept` in my estimation), but either we can merge it or I can push a rebased PR with those nits fixed to get this closed out.

PR #534, opened 2022-02-23
fix(io): handle missing source in stdin

dignifiedquire:
Currently `io::stdin()` when used with `lines().fuse()` from
`futures`, will panic on the first poll. This adds a test for this
case, as well as a potential fix.

### DIFF
--- a/glommio/src/io/buffered_file_stream.rs
+++ b/glommio/src/io/buffered_file_stream.rs
@@ -654,7 +654,12 @@ impl AsyncBufRead for Stdin {
     ) -> Poll<io::Result<&'a [u8]>> {
         match self.source.take() {
             Some(source) => {
-                let res = source.result().unwrap();
+                let res = match source.result() {
+                    Some(res) => res,
+                    None => {
+                        return Poll::Pending;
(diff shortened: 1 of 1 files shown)


glommio/src/io/buffered_file_stream.rs
```diff
@@ -654,7 +654,12 @@ impl AsyncBufRead for Stdin {
     ) -> Poll<io::Result<&'a [u8]>> {
         match self.source.take() {
             Some(source) => {
-                let res = source.result().unwrap();
+                let res = match source.result() {
+                    Some(res) => res,
+                    None => {
+                        return Poll::Pending;
```
dignifiedquire, line 660: I don't know if this is the correct fix, but it does fix both the test case and when using `stdin` in my program to read from it

dignifiedquire:
it seems the missing stdin on CI is failing the tests, any suggestions on how to adjust such that it actually tests the right thing?

dignifiedquire:
it doesn’t actually rely on timing, just the fact that nothing gets written to stdin and that it is available

PR #535, opened 2022-02-23
pass proper max TimeSpec64 to throughput preemption timer source

khodzha:
### What does this PR do?

I suppose the idea was to pass max (positive) value timespec can represent, but due to the cast from u64 to i64 Duration::MAX becomes timespec with negative tv_sec field.

### Motivation

I've noticed EINVALs (-22) being returned from kernel and that should probably not happen

### Additional Notes

Maybe seconds and nanoseconds constraints should be encoded in TimeSpec64 type?
I'm not sure about that so I left it as it is right now.

### DIFF
--- a/glommio/src/sys/uring.rs
+++ b/glommio/src/sys/uring.rs
@@ -62,6 +62,7 @@ use nix::sys::{
 };
 use smallvec::SmallVec;
 
+const MAX_I64_DURATION: Duration = Duration::new(i64::MAX as u64, 999_999_999);
 const MSG_ZEROCOPY: i32 = 0x4000000;
 
 #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


HippoBaro:
Thank you for pointing this out @khodzha 🙇 

I wonder if the `EINVAL` comes from this, though? If the kernel refuses negative `timespec`, I assume it would do so consistently, and the preemption timer wouldn't work. Does the error come from submitting the timer SQE, or from a CQE? 

I think a proper fix is to remove `impl From<Duration>` and only implement `impl TryFrom<Duration>` instead. We can then make sure we do not overflow when converting the value.

bryandmc:
> I think a proper fix is to remove `impl From<Duration>` and only implement `impl TryFrom<Duration>` instead. We can then make sure we do not overflow when converting the value.

I agree with this.

HippoBaro:
Fixed in #536

khodzha:
@HippoBaro i think the error comes from a CQE

PR #536, opened 2022-02-27
fix the conversion from `Duration` to `TimeSpec64`

HippoBaro:
`std::time::Duration` uses `u64` to represent seconds while the kernel
uses signed integers. Our implementation of `From` would cast directly,
resulting in two's complement wrapping.

This commit removes `From` in favor of `TryFrom`. This new
implementation allows the user to know whether the returned `TimeSpec64`
maps the original duration or if it was capped to the maximum
value a `TimeSpec64` can represent.

It also makes it easy for a user to ignore the saturation (presumably
this will only ever trigger when trying to convert `Duration::MAX`
values and the likes).

This was originally reported by @khodzha in #535 time and the fix is
co-authored by them. Credit where credit is due! :bow:


bryandmc (approved):
Besides the failing test, I think this all looks good to me.

PR #539, opened 2022-03-16
limit the number of consecutive tasks scheduling per queue in a quantum

HippoBaro:
At the start of a quantum, the Glommio scheduler selects the task queue
with the lowest vruntime and runs it until one of two things happens:
* The queue is exhausted (there are no more active tasks to run), or,
* The scheduler is preempted.

This mode of operation means that if a task queue contains many tasks,
the scheduler will try to execute them all, even if, at some point
during execution, the vruntime of the current queue is no longer the
lowest of all queues.

This commit limits the number of consecutive tasks the Glommio scheduler
runs in any given queue before falling back to selecting the next queue
with the lowest vruntime (possibly the same queue).

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -80,6 +80,10 @@ pub(crate) const DEFAULT_PREEMPT_TIMER: Duration = Duration::from_millis(100);
 pub(crate) const DEFAULT_IO_MEMORY: usize = 10 << 20;
 pub(crate) const DEFAULT_RING_SUBMISSION_DEPTH: usize = 128;
 
+/// The maximum number of tasks the scheduler runs in a given queue before
+/// considering running another queue.
+const QUANTUM_MAX_TASKS_PER_QUEUE: i32 = 8;
+
 /// Result type alias that removes the need to specify a type parameter
 /// that's only valid in the channel variants of the error. Otherwise, it
 /// might be confused with the error (`E`) that a result usually has in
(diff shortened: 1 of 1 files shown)


glommer:
this violates the preempt timer agreement. If there is quantum, why would we switch task queues ?
Notice that switching task queues too often reduces throughput. Even if you switch to the same task queue, as you now have to run the scheduler code and potentially dirty your cache lines

HippoBaro:
> this violates the preempt timer agreement

How does it? The agreement encompasses two things on a best-effort basis:
* We will schedule IO on that (moving) interval, and
* We will schedule active tasks within that duration.

This commit doesn't violate either and improves the #2 item on that list. I guess the question is whether it is best to fulfill the latency requirement of one queue at the expense of another or whether we should try to be fair.
If we have two queues with the same latency requirements, is it okay to run T1 for the full quantum? That means that we are missing T2's requirement completely. Or does it make sense to share run time between the two (based on vruntime)? 

I argue the latter makes more sense. As for performance, I can try to write a benchmark to quantify the impact, wdyt?

glommer:
The preempt timer is meant to reflect that. If we have a latency requirement of 1ms, that means we run each task queue for 1ms. If there are not enough tasks to fill 1ms that's fine, but it doesn't make sense to keep switching task queues within that 1ms.

Every time you pick a new task queue to run, there is a cold start. The caches are cold, and there is overhead to pick a new task queue. If you incur that overhead too often, you are hurting throughput. If you were to benchmark that, I bet you would see a decrease in throughput in a scenario where you have many classes that are runnable and force frequent switches. Might I bet, I expect that drop to be around 20%.

If you need to switch more often than that, that likely just means you need a lower preempt timer. What, in practice, can you achieve with an arbitrary limit like you propose, that you cannot obtain with a lower preempt timer?

Issue #540, opened 2022-04-06
Tests in timer/timer_impl.rs are time sensitive and can be flaky


vmingchen:
These tests cover the Timer API, so they are sensitive to timing and can be flaky if the dev environment changes. We need to make them more robust.

glommer:
Fully Agree.

Usually the way to make it work is to have a parametrized clock implementation that defaults to wallclock, but for tests you can pass your own that you can manually control.

PR #541, opened 2022-04-06
Make all tests pass on ARM VM

vmingchen:
### What does this PR do?

Make existing unit tests work on a ARM VM.

### Motivation

Make tests more robust.

### Related issues

https://github.com/DataDog/glommio/issues/477
https://github.com/DataDog/glommio/issues/540

### Additional Notes

I don't have a working x86 dev environment yet so I cannot test it over there. It would be great if the reviewer can do so. Thanks!

### Checklist

### DIFF
--- a/glommio/src/timer/timer_impl.rs
+++ b/glommio/src/timer/timer_impl.rs
@@ -1085,7 +1085,12 @@ mod test {
             action.join().await;
             // When action completes we are halfway through the count
             assert_ne!(*(exec2.borrow()), 11);
-            Timer::new(Duration::from_millis(100)).await;
+
+            // TODO(issue#540): Ideally waiting 60ms (10 + 10*10 - 50) should
+            // be enough, but we need as large as 200ms for this test to pass
+            // in an ARM VM. It might be worth looking into the root cause and
+            // a fix.
+            Timer::new(Duration::from_millis(200)).await;
 
             // But because it is detached then it completes the count
             assert_eq!(*(exec2.borrow()), 11);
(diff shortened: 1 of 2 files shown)


glommio/src/timer/timer_impl.rs
```diff
@@ -1127,11 +1132,10 @@ mod test {
                     }
                 }
             });
-            Timer::new(Duration::from_millis(100)).await;
```
glommer, line 1130: removing this line and not replacing by anything is not great.
I understand the timing issue, but the goal of that line was to make sure that we're counting to 10 over time, through the repeat action and not, for example, that the timers are firing all at once.

So we should at least capture a timestamp before and after and make sure that enough time has passed.

vmingchen: Good point. I added an assertion about elapsed time.

vmingchen: Hmmm, the test failed because of that in the github test (passed on my local VM). I need to look more into it.

vmingchen: The failed test is fixed (the minimum time should be 45ms instead of 50).

The nightly doc is still failing but should not be a blocker.

 Please take another look. Thanks!

glommer: @HippoBaro any idea what's going on with the doctest ?

HippoBaro: Rust nightly is panicking. We should submit bug report to the rust folks

vmingchen: Thank you! I filed https://github.com/rust-lang/rust/issues/95793 for the rust nightly bug.

PR #543, opened 2022-04-12
improve error message on bad source

glommer:
Recently a user trying Glommio on WSL got an error that we don't
understand. The problem is really hard to debug, because the only
message is:

```
  Custom { kind: InvalidData, error: "Incorrect source type!" },
```

(which source? which data?)

So in this PR I am improving the message so it includes more
information.

### DIFF
--- a/glommio/src/error.rs
+++ b/glommio/src/error.rs
@@ -57,10 +57,10 @@ pub enum QueueErrorKind {
 }
 
 /// Errors coming from the reactor.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ReactorErrorKind {
     /// Indicates an incorrect source type.
-    IncorrectSourceType,
+    IncorrectSourceType(String),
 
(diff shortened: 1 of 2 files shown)


glommer:
Should wait for user to test!

ChillFish8:
This has improved the error however, I did have to add another section which I think's been missed which is the `From<GlommioError<T>> for io::Error` impl which the error seems to get cast as.

I just hacked it temporarily for now :)

```rust
let msg = match e {
  ReactorErrorKind::IncorrectSourceType(x) => {
       format!("ReactorError {{ kind: IncorrectSourceType {:?} }}", x)
  }
  ReactorErrorKind::MemLockLimit(a, b) => {
       format!("ReactorError {{ kind: '{}' }} {{ kind: MemLockLimit({:?}/{:?}) }}", e, a, b)
  }
 };

io::Error::new(io::ErrorKind::Other, msg)
```

glommer:
@HippoBaro I updated this PR, can you take a look ?

HippoBaro (approved):
LGTM!

PR #549, opened 2022-04-18
chore: fix clippy lints

dignifiedquire:
Clippy is failing currently on nightly CI runs, this should fix it

### DIFF
--- a/glommio/src/io/immutable_file.rs
+++ b/glommio/src/io/immutable_file.rs
@@ -497,7 +497,8 @@ mod test {
     immutable_file_test!(seal_and_stream, path, {
         let fname = path.join("testfile");
         let mut immutable = ImmutableFileBuilder::new(fname).build_sink().await.unwrap();
-        immutable.write(&[0, 1, 2, 3, 4, 5]).await.unwrap();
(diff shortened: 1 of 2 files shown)


dignifiedquire:
``` 
---- executor::test::test_spin stdout ----
thread 'executor::test::test_spin' panicked at 'expected user time on LE is much greater than 50 millisecond', glommio/src/executor/mod.rs:3217:13
```

This seems unrelated, not sure what to do about it.

glommer:
Pulling @vmingchen to take a look as he touched this recently.

@HippoBaro in general, I think CI has been a bit unstable for a while, and that's not great. One of the things I suggest is to pin the nightly version so we're in more control over when those new clippy things will be introduced.

The poor guy's initial patch was all good to begin with =)

vmingchen:
My change was to the `test_no_spin` test, and the failing test in this run is `test_spin`. I just triggered the test to re-run in case it is just flaky.

vmingchen (approved):
Thanks!

vmingchen:
FYI: The test passed in the re-run.

Issue #548, opened 2022-04-15
PollRing and SleepRing have exactly the same submit_one_event implementation


vmingchen:
See [PollRing::submit_one_event](https://github.com/DataDog/glommio/blob/master/glommio/src/sys/uring.rs#L881) and [SleepableRing::submit_one_event](https://github.com/DataDog/glommio/blob/master/glommio/src/sys/uring.rs#L1213).

We should remove this duplication and check other parts as well.

PR #551, opened 2022-04-18
Merge common impls of `submit_one_event` into `submit_event_chain`

vmingchen:
### What does this PR do?

Address https://github.com/DataDog/glommio/issues/548

### Motivation

Refactor code to reduce code duplication.

### Related issues

[A list of issues either fixed, containing architectural discussions, otherwise relevant
for this Pull Request.](https://github.com/DataDog/glommio/issues/548)

### Additional Notes

Pure refactoring; no behavior change.

### Checklist


glommio/src/sys/uring.rs
```diff
@@ -879,32 +917,13 @@ impl UringCommon for PollRing {
...
-        Some(false)
+        submit_event_chain(
+            &mut *self.source_map.borrow_mut(),
+            &mut self.ring,
+            self.allocator.clone(),
+            queue,
+            self.size,
+        )
     }
 }
```
HippoBaro, line 928: Looks like we could go one step further and remove `submit_one_event` from the trait and avoid the indirection of the vtable

vmingchen: How do we do that? `submit_one_event` is used inside `submit_one_event` which is used in the default trait implementation where we cannot refer concrete fields like `self.ring`.

bryandmc:
Hey  @vmingchen  -- In the future, can you either fill out the pull request template or at least delete it (if it's not relevant, though that's also debatable..)? It was surprising to come read this and realize it was just the template unchanged.

glommer:
anything missing for merging this ?

vlovich:
@glommer this seems OK to merge.

PR #552, opened 2022-04-18
sleep by `poll`ing an eventfd

HippoBaro:
We no longer link the rings at all. Instead, we register an eventfd on
the rings (that is written to by the kernel each time a CQE is posted)
and we poll on that instead.

It simplifies the sleeping procedure quite a bit at the expense of
adding one (nonblocking) `read` syscall per reactor trip.

### DIFF
--- a/glommio/src/sys/mod.rs
+++ b/glommio/src/sys/mod.rs
@@ -74,14 +74,21 @@ pub(crate) fn duplicate_file(fd: RawFd) -> io::Result<RawFd> {
     syscall!(dup(fd))
 }
 
-pub(crate) fn create_eventfd() -> io::Result<RawFd> {
-    syscall!(eventfd(0, libc::O_CLOEXEC))
+pub(crate) fn create_eventfd(flags: OFlag) -> io::Result<RawFd> {
+    syscall!(eventfd(0, flags.bits()))
 }
 
 pub(crate) fn write_eventfd(eventfd: RawFd) {
     let buf = [1u64; 1];
     let ret = syscall!(write(eventfd, &buf as *const u64 as _, 8)).unwrap();
     assert_eq!(ret, 8);
+}
+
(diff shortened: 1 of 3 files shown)


HippoBaro:
I've confirmed with @vmingchen that this fixes the hangs he sees when running unit tests. I suspect this will also resolve the occasional hangs in CI we have experienced. If so, I am delighted.

glommio/src/sys/mod.rs
```diff
@@ -84,6 +84,13 @@ pub(crate) fn write_eventfd(eventfd: RawFd) {
     assert_eq!(ret, 8);
 }
 
+pub(crate) fn read_eventfd(eventfd: RawFd) -> io::Result<u64> {
+    let mut buf = [0u8; 8];
+    let ret = syscall!(read(eventfd, buf.as_mut_ptr() as *mut libc::c_void, 8))?;
```
vmingchen, line 89: In `Reactor::new()`, we call `create_eventfd` with `OFlag::O_NONBLOCK`. That means `read()` here may return with `-EAGAIN` right? Then, the following assert may be inappropriate.

glommio/src/sys/uring.rs
```diff
@@ -1736,24 +1671,30 @@ impl Reactor {
...
-        );
-        ring.sleep(&mut link_rings)
-            .or_else(Self::busy_ok)
-            .map(|_| {})
+    /// Sleep until we get an event or for the specified duration (whichever
+    /// comes first)
+    fn sleep(&self, timeout: Option<Duration>) -> io::Result<()> {
+        defer!({
+            // Flush the eventfd!
+            let _ = read_eventfd(self.cqe_events_fd);
```
vmingchen, line 1679: Is this the correct thing to do if the poll call returned because of an event from the remote notifier?

vmingchen (commented):
Thanks a lot for the fixes!

Issue #546, opened 2022-04-15
Inconsistency in  `Sharded::send` and `Sharded::send_to`


dignifiedquire:
Currently `Sharded::send` takes `&self` and `Sharded::send_to` takes as `&mut self`, but there seems to be no need for the latter. Is there a reason for this, or is this just an oversight?

glommer:
definitely an oversight.

PR #562, opened 2022-08-14
Remove redundant mut in Sharded::send_to

LizardWizzard:
### What does this PR do?

Removes redundant mut in receiver of Sharded::send_to

### Motivation

Resolve https://github.com/DataDog/glommio/issues/546

### Related issues

Resolves https://github.com/DataDog/glommio/issues/546

### Additional Notes

It is a change in public API but I believe it is not a major one. Such a change is not listed on [this](https://doc.rust-lang.org/cargo/reference/semver.html) cargo book page 

I tried to check it with cargo-semver but hit [this](https://github.com/rust-lang/rust-semverver/issues/296) bug and failed to get a report. 

Mutable references are turned into immutable ones with autoderef, so there is no need to change code. Existing test code shows that

### Checklist

- [ ] I have added unit tests to the code I am submitting
- [ ] My unit tests cover both failure and success scenarios
- [ ] If applicable, I have discussed my architecture

### DIFF
--- a/glommio/src/channels/sharding.rs
+++ b/glommio/src/channels/sharding.rs
@@ -115,7 +115,7 @@ impl<T: Send + 'static, H: Handler<T> + 'static> Sharded<T, H> {
     ///
     /// [`GlommioError::Closed`]: crate::GlommioError::Closed
     /// [`InvalidInput`]: std::io::ErrorKind::InvalidInput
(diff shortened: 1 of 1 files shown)


glommer:
this is fine, because existing code passing `&mut` will still work. The other way around would require a semver change.

glommer (approved):

PR #571, opened 2022-10-13
Reset current IoVec upon `IOVecMerger::flush` call

vmingchen:
### What does this PR do?

Fix a bug in `IOVecMerger::flush()`.

### Checklist

### DIFF
--- a/glommio/src/io/bulk_io.rs
+++ b/glommio/src/io/bulk_io.rs
@@ -186,7 +186,7 @@ impl<V: IoVec> IOVecMerger<V> {
     }
 
     pub(super) fn flush(&mut self) -> Option<MergedIOVecs<V>> {
-        self.current.map(|x| MergedIOVecs {
+        self.current.take().map(|x| MergedIOVecs {
             coalesced_user_iovecs: self.merged.drain(..).collect(),
             pos: x.pos(),
             size: x.size(),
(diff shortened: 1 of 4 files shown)


glommio/src/io/bulk_io.rs
```diff
@@ -186,7 +186,7 @@ impl<V: IoVec> IOVecMerger<V> {
     }
 
     pub(super) fn flush(&mut self) -> Option<MergedIOVecs<V>> {
-        self.current.map(|x| MergedIOVecs {
+        self.current.take().map(|x| MergedIOVecs {
```
HippoBaro, line 189: Can you explain the context of this? What was the bug, what were you observing, and what's the expected behavior? I wrote this code a long time ago, and I admit I forgot most of the details 😅

vmingchen: Sure. `MergedIOVecs` is used to merge (coalesce) I/Os inside `CoalescedReads` as its `merger`. `CoalescedReads` may [flush its internal merger](https://github.com/DataDog/glommio/blob/6d4aa8fb00b90fa9b6faa31d71a6d054ad043908/glommio/src/io/bulk_io.rs#L254) upon Pending and keep using the same `merger` afterwards. However, the flush method didn't reset the current I/O, and this means the current I/O will still be considered after it is flushed.

The added unit test demonstrates the problematic usage. Without this fix, the test failed. And this fix let the test pass.

HippoBaro: That makes sense to me, thanks!

HippoBaro (approved):

Issue #576, opened 2022-10-24
`pre_allocate` implemented incorrectly?


vlovich:
`pre_allocate` uses the mode of `FALLOC_FL_ZERO_RANGE`. However, the "Allocating disk space" section of https://man7.org/linux/man-pages/man2/fallocate.2.html indicates that you should probably use a value of 0 for the flags if you want the range `offset + length` to not be overwritten with zeroes (i.e. only new space is zeroed out). Considering `pre_allocate` doesn't take a starting offset to let the user make a selection here (which makes sense), I think `pre_allocate` shouldn't be using any flags.

It may be also worth letting the user provide something to control `FALLOC_FL_KEEP_SIZE` in case they want to preallocate for append purposes without actually causing new data to appear in the file.

PR #577, opened 2022-10-25
Fix `pre_allocate` implementation

vlovich:
### What does this PR do?

`pre_allocate` will no longer zero out existing content and will only zero-fill past EOF if requested with the new parameter.

### Motivation

I invoked `pre_allocate` thinking it worked similarly to `posix_fallocate` or how `fallocate` documents preallocating space. Had a surprising source of memory corruption using this.

Making this change also makes `pre_allocate` work on `tmpfs`.

### Related issues

#576 

### Additional Notes

None.

### Checklist

### DIFF
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -192,9 +192,16 @@ impl BufferedFile {
     }
 
     /// pre-allocates space in the filesystem to hold a file at least as big as
-    /// the size argument.
-    pub async fn pre_allocate(&self, size: u64) -> Result<()> {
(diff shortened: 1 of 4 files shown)


vlovich:
It looks like it's failing on some kind of formatting linter but `CONTIBUTING.md` doesn't have a description of how to run it and cargo format seems to change more than what's been modified.

glommer:
I think the issue with the linter is preexisting, because some nightly tests were failing before, and we ended up ignoring some errors - so this slipped.

I fixed and will merge this.

glommer (approved):

Issue #574, opened 2022-10-21
`append` changes `write_at` in a weird way that's not well-documented


vlovich:
The test below will fail if you uncomment the "append" line. The reason is that the position is ignored for all writes. It may be nice to put a clearer warning in the docs for `append` that it changes `write_to` in this way + a callout in `write_to` that `pos` is ignored when the file is opened `append`.

Overall, I'm kind of curious what the use-case `append` solves is. Is it to avoid some kind of TOCTTOU issue around vs just reading the size of the file and using that as the EOF position?

```
    #[test]
    fn glommio_write_with_gaps() {
        test_executor!(async move {
            use tempfile::Builder;

            let temp_dir = Builder::new()
                .prefix("glommio_write_with_gaps")
                .tempdir()
                .unwrap();
            let path = PathBuf::from(temp_dir.path()).join("my_test.bin");

            let dma_file = glommio::io::OpenOptions::new()
                .create(true)
                .read(true)
                // .append(true)
                .write(true)
                .dma_open(path)
                .await
                .unwrap();
            let write_length = dma_file.align_up(1) as usize;
            let mut buffer = dma_file.alloc_dma_buffer(write_length);
            buffer.as_bytes_mut()[0] = 5;
            buffer.as_bytes_mut()[1..].fill(0);
            dma_file.write_at(buffer, 0).await.unwrap();
            assert_eq!(dma_file.file_size().await.unwrap(), write_length as u64);

            let mut buffer = dma_file.alloc_dma_buffer(write_length);
            buffer.as_bytes_mut()[0] = 6;
            buffer.as_bytes_mut()[1..].fill(0);
            dma_file.write_at(buffer, 2 * 1024 * 1024).await.unwrap();
            assert_eq!(dma_file.file_size().await.unwrap(), 2 * 1024 * 1024 + write_length as u64);

            let buffer = dma_file.read_at_aligned(0, write_length).await.unwrap();
            assert_eq!(buffer.len(), write_length);
            assert_eq!(buffer[0], 5);
            assert_eq!(buffer[1], 0);

            let buffer = dma_file.read_at_aligned(2 * 1024 * 1024, write_length).await.unwrap();
            assert_eq!(buffer.len(), write_length);
            assert_eq!(buffer[0], 6);
            assert_eq!(buffer[1], 0);
        });
```

glommer:
This is very surprising to me.

but in a nutshell, just shows how complex and full of edge cases things are =)

I checked the man page, which says this:

```
[...]

PR #578, opened 2022-10-25
Add a warning note for O_APPEND corner case

vlovich:
### What does this PR do?

O_APPEND causes a bit surprising behavior in that the position passed to `write_at` is ignored. Document this footgun.

### Motivation

I had accidentally opened the file with `O_APPEND` and my writes were not ending up where I was setting the position
to.

### Related issues

#574 

### Additional Notes

One alternate consideration might be to prevent opening/creating a `DmaFile` with `append` set in the first place.

### Checklist

[] I have added unit tests to the code I am submitting (N/A)
[] My unit tests cover both failure and success scenarios (N/A)

### DIFF
--- a/glommio/src/io/dma_file.rs
+++ b/glommio/src/io/dma_file.rs
@@ -222,6 +222,12 @@ impl DmaFile {
     /// is the situation, for example, when the device runs out of space
     /// (See the man page for write(2) for details)
     ///
+    /// <p style="background:rgba(255,181,77,0.16);padding:0.75em;">
+    /// <strong>Warning:</strong> If the file has been opened with `append`,
+    /// then the position will get ignored and the buffer will be written at
+    /// the current end of file. See the [man page] for `O_APPEND`
+    /// </p>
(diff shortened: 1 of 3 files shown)


vlovich:
Having trouble getting the test case to pass:

A) `cargo +nightly fmt --all` doesn't modify anything
B) The failure mentioned is wrong. The comment below (in fact the word "probes") only appears in README.md not `lib.rs`
```
Diff in /home/runner/work/glommio/glommio/glommio/src/lib.rs at line 230:
 //! 512
 //! ```
 //!
-//! Glommio also requires a kernel with a recent enough `io_uring` support, at least recent enough to run discovery probes.
-//! The minimum version at this time is 5.8
+//! Glommio also requires a kernel with a recent enough `io_uring` support, at
+//! least recent enough to run discovery probes. The minimum version at this
+//! time is 5.8
```
C) I haven't modified any of the relevant code...

glommer (approved):

Issue #575, opened 2022-10-22
Extend read ownership of `DmaBuffer` past `write_to`


vlovich:
I had this thought that it would be really convenient to be able to continue doing some kind of background processing of the data buffer in a background thread (e.g. converting it into a more optimal data structure) while I/O is ongoing without requiring a copy. I'm not too familiar with how io_uring works so maybe this isn't possible, but would it make sense to try to offer the ability to extend the read lifetime of a `DmaBuffer` by creating a `ReadReference` which would preclude that buffer being reused for allocation until it's dropped? Happy to try to work on this if you think it's possible & would welcome a contribution along these lines.

glommer:
I think it is possible, but would require a different API.

This needs more thought to make sure the API is sane, we wouldn't be able to do this by just adding stuff on top of `write`. But it is something that other people had requested in the past, and I think it is reasonable.

This seems to me mostly a rust issue, not an io_uring issue. io_uring will write whatever pointer you give it. The biggest question is how to derive an API in which buffers are safely shared on the rust side. The buffer needs to be writeable up to a point, but then after you pass it down to write, it needs to be read only. It also has to be safe enough so it is not dropped until the kernel returns, and handle cancellations.

IOW, totally possible, just need a really sane API.

vlovich:
I figured it was going to be a Rust thing. The only concern I had was that submitting an io_uring buffer passed off ownership of the buffer to the kernel. My perhaps naiive idea (I'm very new to Rust), is to basically just have:

```
impl DmaBuffer {
  fn seal(self) -> SealedDmaBuffer { ... }
}
```

Then all `write` methods can be changed to require taking a `SealedDmaBuffer`. A `SealedDmaBuffer` is clonable and acts like a `ReadResult`. Each `SealedDmaBuffer` holds a non-atomic reference counter that prevents that buffer from getting reused.

PR #579, opened 2022-10-25
`SealedDmaBuffer` that provides a read-only view that can survive being submitted for writing.

vlovich:
### What does this PR do?

Introduces a `SealedDmaBuffer` which is a read-only view that's submitted to file writes. This new struct is clonable and maintains
a Rc to the underlying owned buffer.

### Motivation

Today submitting a `DmaBuffer` causes you to lose the ability to do anything else with that buffer. However, it's desirable to be
able to post-process that buffer in the background without losing it (e.g. perhaps you want to maintain that buffer in-memory
as a cache of what you wrote or you want to post-process the data in some way).

### Related issues

#575 

### Additional Notes

I wasn't sure if I should add `write_sealed_at` or break compat by changing `write_at` (callers would need to call `.into` or `.seal` on the buffer they were previously supplying). I went with the option of not breaking back compat but technically since this is
a `0.x` library it probably would be OK?

I also wasn't really sure what kind of tests would constitute test coverage here. Putting up the code as a starting discussion point of what I was thinking of.

### Checklist

### DIFF
--- a/glommio/src/io/dma_file.rs
+++ b/glommio/src/io/dma_file.rs
@@ -19,7 +19,7 @@ use crate::{
         read_result::ReadResult,
         ScheduledSource,
     },
-    sys::{self, sysfs, DirectIo, DmaBuffer, PollableStatus},
+    sys::{self, sysfs, DirectIo, DmaBuffer, PollableStatus, DmaSource},
 };
 use futures_lite::{Stream, StreamExt};
 use nix::sys::statfs::*;
@@ -250,7 +250,7 @@ impl DmaFile {
     pub async fn write_at(&self, buf: DmaBuffer, pos: u64) -> Result<usize> {
         let source = self.file.reactor.upgrade().unwrap().write_dma(
             self.as_raw_fd(),
-            buf,
+            DmaSource::Owned(buf),
             pos,
             self.pollable,
         );
--- a/glommio/src/io/dma_file.rs
+++ b/glommio/src/io/dma_file.rs
@@ -19,7 +19,7 @@ use crate::{
         read_result::ReadResult,
         ScheduledSource,
     },
-    sys::{self, sysfs, DirectIo, DmaBuffer, PollableStatus, DmaSource},
+    sys::{self, sysfs, DirectIo, DmaBuffer, DmaSource, PollableStatus},
 };
 use futures_lite::{Stream, StreamExt};
 use nix::sys::statfs::*;
@@ -29,6 +29,7 @@ use std::{
     os::unix::io::{AsRawFd, RawFd},
     path::Path,
     rc::Rc,
+    sync::Arc,
 };
 
 use super::Stat;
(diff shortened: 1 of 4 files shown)


glommer:
Hi @vlovich ! I merged your other two commits, so there's a conflict here now. (thanks!)

I'll review this with more care tomorrow, but from your description alone, the separate file is warranted if nothing else, so that the user not always have to worry about sealing buffers.

glommer:
In which situations do you expect the user to keep using a `DmaBuffer` ?

glommer:
I think a potential alternative approach for us to consider is do something similar to what mutexes will commonly do: have a variant that takes &T, and another variant that takes Rc<T> as an argument. 

The main problem of internally converting `DmaBuffer` to its Sealed version, is that we're adding an allocation for the `Rc` for no reason.

So I'd consider letting `write_at` as is: it is safe, because we take ownership of the buffer, which cannot be written to.

A different signature, `write_static_at`, takes an `Rc<DmaBuffer>` (static comes from nomenclature usually used for `T` vs `Rc<T>`)

vlovich:
Oh very good points. I forgot that `Rc` would have to be a heap-allocated struct for the counter

> In which situations do you expect the user to keep using a DmaBuffer ?

Only to fill it up. Normally I'd expect them to not use it.

My goal is to make `SealedDmaBuffer` "0-overhead". Do you think that's possible by moving the counter down into the underlying buffer which is presumably already heap-allocated?

glommer:
It is possible, but now you have to allocate a bigger buffer, and this can be quite wasteful and have bad results with the allocator.

for example, if you want 4kB buffers, you need at least 4kB+512b buffers allocated (for alignment reasons) that can create fragmentation.

You can then have an (atomic) counter somewhere and essentially open code an RC (Arc, if you go atomic)

vlovich:
That's a fair point. I think I'll go with the simpler version where the API can accept `Rc<T>` or `T` vs trying to hand-roll an RC myself.

vlovich:
I realized I probably might need `Arc` as well so I added both `Rc` and `Arc` variants. I'm not familiar with the nomenclature to use so for now I went `write_shared_at` and `write_send_at`. Please let me know what the final names should actually be.

glommer:
BTW - tests are failing.
The failure are related to an nvme environment variable, but since all other I/O tests are passing there must be something wrong with the added code.

About naming, some prior art:

https://docs.rs/async-mutex/latest/async_mutex/struct.Mutex.html#method.lock_arc

Async mutex will call it "lock_arc"

glommer (approved):

glommer:
I am 100% ok now with the implementation - thanks!

Let's just figure out the naming and why tests are failing and I can press the button

vlovich:
Sorry for the delay. Other things came up at work and I've also come down with something. Will try to get this PR cleaned up and ready to merge.

vlovich:
Another thought I had is that actually you don't need to pad by 512 bytes if you put the counter at the end of the allocation, right?

glommer:
Sorry @vlovich, I am not sure which counter you are talking about.

In any case, the best answer to questions like that come in the form of unit tests!

vlovich:
We had discussed how we could store the reference count for the io_uring buffer inline with the allocation. Your concern was that it brings up padding issues to maintain alignment, but I think that's only true if you put the counter at the end. If you put it at the beginning, you don't need any extra padding I think - you can just tack on the extra space for a size_t.

glommer:
Okay, that makes sense now. But I don't think it changes anything materially.

Let's say you have a 160kB pool of memory. That can fit 10 16kB buffers. But if you now add an in-buffer counter to that, you will be able to allocate only 9 buffers out of that pool.

The problem is made worse by alignment issues, and that may indeed be made less prominent by adding the reference counter to the end, but you still have the memory fragmentation issue.

vlovich:
> Let's say you have a 160kB pool of memory. That can fit 10 16kB buffers. But if you now add an in-buffer counter to that, you will be able to allocate only 9 buffers out of that pool.

Sure. Is that maybe an acceptable tradeoff? We're talking about ~0.1% overhead with 4kib buffers and 1/4 of that for 16 kib buffers while adding some nice flexibility without forcing any heap allocations onto the caller. I guess I don't have context to know how to evaluate whether or not the 160kib use-case can maybe allocate 160kib + 40 bytes to account for the overhead of the counter? Or are you saying you have a contiguous 160kib and you're wrapping a DmaBuffer around each consecutive chunk & that each one of those then needs to be 4kib (512 byte?) aligned for direct I/O?

glommer:
the buffers will have to be aligned at some level, so even if you add this to the end, as you suggested, you then have to pad the next allocation, so the overhead is at least 512b.

I am saying that it is actually likely more than that, since you will want to align a 4kB buffer to a 4kB address. So now you have 4kB+512, but if the next buffer wants to be 4kB aligned, that's a lot of overhead.

vlovich:
That's a good point. I'm mostly dealing with relatively large buffers being allocated (i.e. 8 MiB or more at a time) so the extra overhead is minimal for my use-case and I hadn't considered the use-case that you're allocating the minimum amount at all times. I think for cases where the intrusive RC would work (i.e. allocating large buffers), we'd have to then create a separate API call for when users want the intrusive RC and know they're working with larger buffers such that it would make sense. Thanks for pointing out the bug in my thinking.

vlovich:
Tests were missing a `.join("testfile")` when opening the path.

vlovich:
@glommer the test failures here seem unrelated to my change. Did something change in the test environment?

vlovich:
#586 should fix the CI failures.

glommer:
There were some conflicts, they seemed easy enough, and I resolved them. Tests are not passing - not sure if preexisting or due to the resolution. I will take a look soon.

It looks great other than that.

vlovich:
I think the merge conflicts were resolved incorrectly. It was just a conflict that I added 1 test in the other branches in a similar spot to 2 other tests so there wasn't any merging needed vs just telling kdiff3 the alignment points.

glommer:
@vlovich FYI : I still see CI failures.

vlovich:
Yeah I saw. I've been in and out of meetings today. Will try to get this fixed. Not sure how to repro it locally so iteration times are a PITA.

glommer:
It reproduces for me.
I'll take a look

glommer:
The problem is in the doc test for `write_rc_at`. The issue is that even though the code doesn't run, it does have to compile, which it doesn't.

It doesn't compile because `index()` does not return a `Future`, and `join!` expects both arms to be the same.

We need to make `index()` async, and in general it has to return something similar to `write_rc_at` as well.

vlovich:
Yeah, I just finished it. One thing I wanted to highlight is that the `arc` implementation may be broken design-wise (I copied the implementation mostly from `rc`). Here's the example code that works for me:

```
    /// ex.run(async {
    ///     let file = DmaFile::create("test.txt").await.unwrap();
    ///
    ///     let mut buf = file.alloc_dma_buffer(4096);
    ///     // Write some data into the buffer.
    ///     populate(&mut buf);
    ///
    ///     // Seal the buffer by moving ownership to a non-threaded reference
    ///     // counter on the heap.
    ///     let write_buf = Arc::new(buf);
    ///     let read_buf = write_buf.clone();
    ///     let buf_view = unsafe { std::mem::transmute(read_buf.as_bytes()) };
    ///
    ///     let (written, transformed) =
    ///         join!(
    ///             async { file.write_arc_at(write_buf, 0).await.unwrap() },
    ///             async { () }
    ///         );
    ///     let transformed = executor().spawn_blocking(move || async move { transform(buf_view) });
    ///     assert_eq!(written, 4096);
    ///     file.close().await.unwrap();
    ///
    ///     // transformed AND buf can still be used even though the buffer got
    ///     // written. Note that there may be performance issues if buf is large
    ///     // and you remain hanging onto it.
    /// });
```

Note the existence of `unsafe` because the lifetime of the `buf_view` is required to be `static`. Similarly, I can't move the actual `Arc<DmaBuffer>` into the `spawn_blocking` closure because `DmaBuffer` isn't Send... Not sure how to fix it here. Should I actually require passing in `Arc<RwMutex<DmaBuffer>>`? I can't implement the `Send` trait for `Arc<DmaBuffer>` (even though in principle it's safe?) because `Arc` is external to this crate and `Send` on `DmaBuffer` is incorrect.

I also added a new test `mirror_buffer_to_two_files` but it's not passing for me for some reason (not sure about CI) - the read fails but I'm not sure if that's an issue on my machine. I'm not noticing anything wrong in the code and the returned error code 95 EOPNOTSUPP makes no sense and isn't documented for `read`. I've tried various variants of this including one where I only write the buffer directly (no `write_rc_at`) and then read the file directly and get the same problem (yet `file_simple_readwrite` passes for me... :confused:) 

I'm noticing one thing that's a bit weird API-wise is that I can create a DmaBuffer in one file and then pass it into a second file (perhaps on a different filesystem with totally different alignment requirements), but I can't allocate the DmaBuffer outside of a file. It's a bit weird and not sure a) if that might cause issues b) whether there should be an ability to just call `DmaBuffer::new` that carries a documentation warning "it's your responsibility to match the alignment of the file".

glommer:
Do you have a strong use case for the Arc version? It would be better to merge just the rc and let it cook for a while.

glommer:
It's more in character for us to keep just the RC, and also seems safer in this case, so I'll advocate for that. It gives us time to mature options as well.

glommer:
@vlovich heads up, I removed the arc variant, and merged this.

PR #581, opened 2022-11-13
RFC: Initial vsock stream support

gkelly:
### What does this PR do?

Adds initial `vsock` stream socket support to `glommio`. This is just to start a discussion of the approach, and is _NOT_ ready to land.

### Motivation

I am working on high-throughput `vsock` applications running inside of hypervisors and am evaluating `glommio` as a potential IO framework.

This implementation is based heavily off of the [`tcp_socket`](https://github.com/DataDog/glommio/blob/master/glommio/src/net/tcp_socket.rs) implementation.

### Additional Notes

This is a draft PR, it is not yet ready to merge. It is missing the following:
- [ ] Decide on how to treat `SOCK_DGRAM` sockets. I'm open to feedback here, but given how similar a `STREAM` and `DGRAM` socket behave under `vsock` I'm tempted to genericize the pieces across the `SOCK_` type.
- [ ] Make the `vsock` example more complete, it's currently just an echo server to prove it works.
- [ ] **Tests.**
- [ ] **Benchmarks.**

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -59,3 +59,7 @@ path = "hyper.rs"
 [[example]]
 name = "gate"
 path = "gate.rs"
+
+[[example]]
+name = "vsock"
+path = "vsock.rs"
--- a/examples/vsock.rs
+++ b/examples/vsock.rs
@@ -1,0 +1,22 @@
+use futures_lite::io::{AsyncReadExt, AsyncWriteExt};
+use glommio::{
+    net::{VsockListener, VsockStream},
+    prelude::*,
+};
+use std::io::Result;
+
+fn main() -> Result<()> {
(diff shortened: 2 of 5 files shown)


glommer:
Thanks @gkelly !

I am taking a look at this now

glommer:
I don't have a lot of comments, to be honest - aside from the lack of documentations on public functions that I am sure you will fix eventually.

The implementation is quite straightforward, and I think the abstraction is the right one (having a generic socket type would be overkill IMHO, so a new class for VSock makes sense).

The example application is neat and simple. The only thing that I'd like to see more is testing and benchmarks, but I realize this can get a bit non trivial with vsock and its setup.

What are the milestones you have in mind before considering ready to merge?

PR #583, opened 2022-12-15
Add fsync interface to DmaFile.

vlovich:
### What does this PR do?

Exposes the full fsync API via io_uring.

### Motivation

Sometimes you need a full fsync for database correctness. For example, making sure any newly written data is actually published and will be guaranteed visible before deleting other data to avoid losing information in the face of unexpected termination. My understanding is fdatasync guarantees that the blocks are written to the disk but does not guarantee that subsequent reads will see those blocks. I believe the set of extents backing the file are tracked in the metadata portion of the file.

### Related issues

N/A

### Additional Notes

Wasn't sure about bool vs enum for fsync. Went with bool because vscode makes the variable name visible but may not be very good in other dev environments.

### Checklist

It doesn't look like fdatasync really has any meaningful test coverage. Not sure what test would best add coverage here but I just randomly picked between fdatasync vs fsync for existing tests.

### DIFF
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -188,7 +188,14 @@ impl BufferedFile {
     /// all writes to the device, providing durability even if the system
     /// crashes or is rebooted.
     pub async fn fdatasync(&self) -> Result<()> {
-        self.file.fdatasync().await.map_err(Into::into)
+        self.file.fsync(true).await.map_err(Into::into)
+    }
+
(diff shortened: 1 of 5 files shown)


vlovich:
I may be wrong though. https://news.ycombinator.com/item?id=25171572 and https://man.archlinux.org/man/fdatasync.2.en indicates that fdatasync forces the flush of any metadata required for a subsequent read (file size + extents). So the value add here may be minimal. Leaving this open just in case it's interesting but I could also see a philosophical reason to not provide it and tell people to use fdatasync.

PR #584, opened 2022-12-26
Expose a `.stat` method for files

vlovich:
Obtaining the file_size is convenient but sometimes you need the cluster size and/or the size allocated on-disk.
Add an exposed `.stat` call that provides this. Rather than exporting `libc::statx`, I figured I would write a friendlier
`Stat` function (I also tried to make them friendlier to use than st_blksize / st_blocks). Hope that's OK.

### What does this PR do?

Expose a call to `stat` an open file.

### Motivation

My program needs a way to get the filesystem cluster size to make sure it's doing I/O optimally on the filesystem it's running on + the actually allocated size for various reporting / statistics.

### Related issues

N/A

### Additional Notes

### Checklist

[/] My unit tests cover both failure and success scenarios
Partially. I've tested success scenarios but not sure how to test a failure scenario for stat.

### DIFF
--- a/glommio/src/io/buffered_file.rs
+++ b/glommio/src/io/buffered_file.rs
@@ -14,6 +14,8 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use super::Stat;
+
 type Result<T> = crate::Result<T, ()>;
 
 /// An asynchronously accessed file backed by the OS page cache.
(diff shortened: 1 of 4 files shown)


vlovich:
Test failure appears to be a breakage on master.

vlovich:
#586 should fix the CI failures.

PR #600, opened 2023-08-14
Improve string formatting

nyurik:
### What does this PR do?
* improve performance of the `glommio/src/error.rs` to avoid allocating strings only to print them
* Use clippy to inline all format args, and fix them by hand afterwards:
* remove a few unneeded `&` in `format!("{}", &str_val)` -- this actually has a minor performance hit (about 6%) due to dynamic dispatch that rustc cannot inline.

```
❯ cargo clippy --workspace --fix --benches --tests --bins -- -A clippy::all -W clippy::uninlined_format_args
```

Note that there are many cases of `panic!` and `assert!` that also have inlinable variables that cannot be inlined due to 2018 edition. In 2021, their behavior is made consistent, but that requires migrating to 2021 first. `assert_eq!` doesn't have this issue.

### Motivation

Cleaner, easier to read code, and a few minor performance optimizations


nyurik:
it seems, suprisingly, that the clippy & fmt tasks are implemented as unit tests - and moreover they are using unstable nightly, thus failing.  I believe this should be done as a separate PR that possibly require some more in-depth code knowledge than i have - any maintainers out there? :)

glommer:
I've been very annoyed with the nightly thing. I bought the reasoning at first, but it's been the most consistent source of failures and headaches in CI.

I'm down to drop it. Want me to do it, and then you can rebase your stuff on top of it ?

You're welcome to patch the CI file as well if it makes your rebase easier. I kind of like that calling fmt and clippy are part of the unit tests, it's the nightly thing that I think must go away.

bryandmc:
Drop nightly, it's been too much of a headache for sure.

nyurik:
Dropped, and made a few more minor fixes.  Note that overall CI can now be significantly simplified - I don't think you need build cache because the new SPARSE cargo support (since v1.70 I think?) is standard, thus all loads are pretty fast. The only exception is for building MSRV verison.  In a separate PR, a much easier path would be to have 3 build jobs: stable, nightly, msrv; don't use cache for anything except msrv for now; in each job perform all the needed steps (build, test, doc build).  Lastly, all this can be easily done with a single job and a param matrix... just my 0.02 :heavy_dollar_sign:

nyurik:
turns out doc builds were failing all along.  In a separate PR, doc build needs to be fixed.  I made a CI change but commented it out for now to keep this PR smaller

nyurik:
Heh, not so simple. That's why i don't think fmt/clippy should be part of the tests themselves -- they are executed using the **current** build environment, which means you can either not build nightly nor msrv at all (because they may have different settings for fmt and clippy), or have some magical features to exclude those specific tests from all but the stable build, e.g. `cargo test --features test-stable`, which is no different than simply to add a `cargo fmt --all -- --check`.

Instead, I found `just` to be a far better alternative to all this, and auto-configuring a git-push hook. I use this pattern in all of my crates, e.g. https://github.com/nyurik/sqlite-hashes.

Basic idea:
* all CI tests are added to [a pre-push hook](https://github.com/nyurik/sqlite-hashes/blob/main/.cargo-husky/hooks/pre-push)
* pre-hook is autoinstalled using [cargo-husky](https://github.com/nyurik/sqlite-hashes/blob/main/Cargo.toml#L30)
* all local development is done with simple [justfile targets](https://github.com/nyurik/sqlite-hashes/blob/main/justfile) (similar to make, but far cleaner)

nyurik:
I think fixing CI should be outside of the scope for this PR, or else it may get too messy

glommer:
I just merged #601 , that should make things a bit better. It's by no means the end goal here, I tried to do as little as possible to just make your life easier. Then we can continue to iterate.

glommer:
Thanks!

PR #602, opened 2023-08-25
Make `GlommioError` implement `Send` and `Sync`

c410-f3r:
### What does this PR do?

Makes `GlommioError: Send + Sync`

### Motivation

Third-party requires `T: Send + Sync`

### Additional Notes

I am not aware of any thread panic error that is `!Sync`


c410-f3r:
NVM, `std::thread::JoinHandle::join_handle` returns `Result<T, Box<dyn Any + Send + 'static>>`

PR #607, opened 2023-10-04
Fix Unbound background thread pool affinity

vlovich:
### What does this PR do?

Creating a local executor with a non-Unbound placement but with a background thread of  `Unbound` placement now correctly has those background threads free-floating across any CPU instead of pinned to the CPU affinity assigned to the executor.

This is accomplished by creating the background thread pool before we bind the executor's CPU affinity.

### Motivation

Seems like buggy behavior for the executor having an affinity breaking `Unbound` placement for background threads. Violated principle of least surprise when my pinned executor was doing blocking dispatch to 16 threads but all that work was still getting scheduled on just 1 thread.

### Additional Notes

If a non-unbound placement for background threads is provided, this PR has no effect.

### Checklist

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -4149,6 +4149,51 @@ mod test {
     }
 
     #[test]
+    fn blocking_function_placement_independent_of_executor_placement() {
+        let affinity = nix::sched::sched_getaffinity(nix::unistd::Pid::from_raw(0)).unwrap();
+        let num_cpus_accessible_by_default = (0..nix::sched::CpuSet::count())
+            .map(|cpu| affinity.is_set(cpu).unwrap() as usize)
+            .sum::<usize>();
+        if num_cpus_accessible_by_default < 2 {
+            eprintln!(
+                "Insufficient CPUs available to test blocking_function_placement_independent_of_executor_placement (affinity only allows for {})",
+                num_cpus_accessible_by_default,
+            );
+            return;
+        }
+
+        let num_schedulable_cpus = LocalExecutorBuilder::new(Placement::Fixed(0))
+            .blocking_thread_pool_placement(PoolPlacement::Unbound(2))
+            .spawn(|| async {
+                executor()
+                    .spawn_blocking(move || {
+                        let pid = nix::unistd::Pid::from_raw(0);
+                        let affinity =
+                            nix::sched::sched_getaffinity(pid).expect("Failed to get affinity");
+                        (0..nix::sched::CpuSet::count())
+                            .map(|cpu| {
+                                affinity
+                                    .is_set(cpu)
+                                    .expect("Failed to check if cpu affinity is set")
+                                    as usize
+                            })
+                            .sum::<usize>()
+                    })
+                    .await
+            })
+            .unwrap()
+            .join()
+            .unwrap();
+
+        assert!(
+            num_schedulable_cpus >= num_cpus_accessible_by_default,
+            "num schedulable {}, num cpus accessible {}",
+            num_schedulable_cpus,
+            num_cpus_accessible_by_default,
+        );
+    }
+
+    #[test]
     fn blocking_pool_invalid_placement() {
         let ret = LocalExecutorBuilder::new(Placement::Unbound)
             .blocking_thread_pool_placement(PoolPlacement::Unbound(0))
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -36,7 +36,8 @@ use crate::{
     error::BuilderErrorKind,
     executor::stall::StallDetector,
     io::DmaBuffer,
-    parking, reactor, sys,
+    parking, reactor,
+    sys::{self, blocking::BlockingThreadPool},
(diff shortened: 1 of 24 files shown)


glommio/src/executor/mod.rs
```diff
@@ -1140,6 +1141,9 @@ impl LocalExecutor {
         cpu_binding: Option<impl IntoIterator<Item = usize>>,
         mut config: LocalExecutorConfig,
     ) -> Result<LocalExecutor> {
+        let blocking_thread =
+            BlockingThreadPool::new(config.thread_pool_placement, notifier.clone())?;
```
vlovich, line 1145: So technically this now maybe has problems with the NUMA logic below? Not 100% sure about the impact.

The other way to do this maybe is to retrieve the CPU affinity prior to binding & passing that into `BlockingThreadPool` so that it resets the CPU affinity when the thread starts?

vlovich:
I don't think the build failure in the tests has to do with any of my changes does it? I'm unable to repro locally but I have a newer Rust version (1.72 vs 1.58). Maybe the minimum Rust version needs to be bumped?

When I try to run 1.58 locally I get
```
cargo +1.58 check --lib
    Updating crates.io index
error: failed to select a version for the requirement `memchr = "^2.3.3"`
candidate versions found which didn't match: 2.5.0, 2.4.1, 2.4.0, ...
location searched: crates.io index
required by package `futures-lite v1.13.0`
    ... which satisfies dependency `futures-lite = "^1.11.1"` (locked to 1.13.0) of package `examples v0.0.0 (glommio/examples)`
```

When I exclude `examples` from the workspace, the next failure is:
```
cargo +1.58 check --lib
    Updating crates.io index
error: failed to select a version for the requirement `once_cell = "^1.8"`
candidate versions found which didn't match: 1.17.2, 1.17.1, 1.17.0, ...
location searched: crates.io index
required by package `ahash v0.7.6`
    ... which satisfies dependency `ahash = "^0.7"` (locked to 0.7.6) of package `glommio v0.8.0 (glommio/glommio)`
```

glommio/src/executor/mod.rs
```diff
@@ -4148,6 +4152,40 @@ mod test {
             .unwrap();
     }
 
+    #[test]
+    fn blocking_function_placement_independent_of_executor_placement() {
+        let num_schedulable_cpus = LocalExecutorBuilder::new(Placement::Fixed(0))
+            .blocking_thread_pool_placement(PoolPlacement::Unbound(4))
```
glommer, line 4158: I am afraid this test will break when executed in machines with less than 4 CPUs.

Should we skip it in that scenario? It's still a valuable test

glommer:
@vlovich I'm okay bumping. 1.58 is very old, and if I try, I get a different set of errors altogether, meaning this won't be easy to track down.

glommio/src/executor/mod.rs
vlovich: This doesn't actually need 4 CPUs. Can be reduced to 2 and I'll adjust to skip the test if there's only 1 online CPU found.

glommer: that'd be better

glommer:
I opened https://github.com/DataDog/glommio/actions/runs/6412940182/job/17411069816?pr=608, but there's now a flurry of clippy stuff.

I'll take a look tomorrow.

vlovich:
I bumped it to 1.65 here as it seems like the root cause (at least for me locally) was backtrace acquiring a new dep on addr2line which required 1.65. Maybe other packages had updated dependencies too, but that was the obvious one. Not sure about the policy about how to pick the minimum Rust version.

vlovich:
@glommer one thing that nightly clippy warns me about that is the single warning not resolved by this PR:

```
warning: usage of an `Arc` that is not `Send` or `Sync`
   --> glommio/src/channels/channel_mesh.rs:311:20
    |
311 |             peers: Arc::new(RwLock::new(Vec::new())),
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the trait `Sync` is not implemented for `RwLock<Vec<Peer>>`
    = note: required for `Arc<RwLock<Vec<Peer>>>` to implement `Send` and `Sync`
    = help: consider using an `Rc` instead or wrapping the inner type with a `Mutex`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
    = note: `#[warn(clippy::arc_with_non_send_sync)]` on by default
```

The warning seems accurate since this is glommio's RwLock which is single-thread only, so potentially allowing for sharing that between threads via Rc seems dangerous. Certainly changing it to Rc seems to work fine, but I didn't dig into it more deeply than that if there are risks with that conversion.

glommer:
We should be fine with an RC here.

glommio/src/executor/mod.rs
glommer: I think this should be fine. The executor is still allocated in the node where it wants to run, and the BlockingThreadPool is way less performance sensitive.

glommer (approved):

PR #608, opened 2023-10-04
bump minimum version to 1.72

glommer:
Increasingly difficult to run 1.58, many dependency issues when compiling.


vlovich:
Didn't bump all the way to 1.72 in https://github.com/DataDog/glommio/pull/607 but instead did 1.65 which is the minimum version that works. All clippy issues resolved. I appreciate the impetus to own the CI state as the maintainer. Hope you don't mind me doing it for you.

glommer:
@vlovich I hate CI =p

You're doing me a big favor. Don't worry about that.

Issue #610, opened 2023-10-13
futures::try_join on DmaStreamWriter can cause panic


tontinton:
Calling `poll_write` after it returned `Pending` results in crash. This in itself might be ok, but `futures::try_join!` triggers this behavior, so we should either fix the bug or at least document that.

This program triggers the crash:

```rust
use futures::{try_join, AsyncWriteExt};
use glommio::{
    io::{DmaFile, DmaStreamWriterBuilder},
    LocalExecutorBuilder, Placement,
};

fn main() {
    let builder = LocalExecutorBuilder::new(Placement::Fixed(1));
    builder
        .spawn(|| async move {
            let file1 = DmaFile::create("/tmp/bug1").await.unwrap();
            let mut writer1 = DmaStreamWriterBuilder::new(file1)
                .with_write_behind(1)
                .with_buffer_size(4096)
                .build();
            let file2 = DmaFile::create("/tmp/bug2").await.unwrap();
            let mut writer2 = DmaStreamWriterBuilder::new(file2)
                .with_write_behind(1)
                .with_buffer_size(4096)
                .build();
            let buf = [b'A'; 4096];
            println!("Writing...");
            for _ in 0..100000 {
                try_join!(writer1.write_all(&buf), writer2.write_all(&buf)).unwrap();  // Crash here
            }

        })
        .unwrap()
        .join()
        .unwrap();

    println!("No bug!");
}
```

With the following output:

```
Writing...
thread 'unnamed-1' panicked at /home/tony/.cargo/registry/src/index.crates.io-6f17d22bba15001f/glommio-0.8.0/src/io/dma_file_stream.rs:832:9:
assertion failed: self.waker.is_none()
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   1: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:117:5
   3: glommio::io::dma_file_stream::DmaStreamWriterState::add_waker
             at /home/tony/.cargo/registry/src/index.crates.io-6f17d22bba15001f/glommio-0.8.0/src/io/dma_file_stream.rs:832:9
[...]

PR #611, opened 2023-10-13
Allow DmaFileStream to be polled multiple times

tontinton:
Fixes https://github.com/DataDog/glommio/issues/610 and https://github.com/DataDog/glommio/issues/612

### DIFF
--- a/glommio/src/io/dma_file_stream.rs
+++ b/glommio/src/io/dma_file_stream.rs
@@ -827,8 +827,6 @@ macro_rules! ensure_not_closed {
 
 impl DmaStreamWriterState {
     fn add_waker(&mut self, waker: Waker) {
-        // Linear file stream, not supposed to have parallel writers!!
-        assert!(self.waker.is_none());
         self.waker = Some(waker);
     }
 
@@ -839,13 +837,7 @@ impl DmaStreamWriterState {
         self.flushed_pos() > self.aligned_pos
     }
 
-    fn initiate_close(
-        &mut self,
-        waker: Waker,
-        state: Rc<RefCell<Self>>,
-        file: Rc<DmaFile>,
-        do_close: bool,
-    ) {
+    fn initiate_close(&mut self, state: Rc<RefCell<Self>>, file: Rc<DmaFile>, do_close: bool) {
         self.file_status = FileStatus::Closing;
         let final_pos = self.current_pos();
         self.flush_padded(state.clone(), file.clone());
@@ -854,7 +846,11 @@ impl DmaStreamWriterState {
             futures_lite::future::yield_now().await;
 
             defer! {
-                waker.wake();
+                let mut state = state.borrow_mut();
+                if let Some(waker) = state.waker.take() {
+                    drop(state);
+                    waker.wake();
+                }
             }
 
             for flush in pending.drain(..) {
@@ -960,9 +956,14 @@ impl DmaStreamWriterState {
             if !collect_error!(state, res) {
                 state.flush_state.on_complete(flush_pos);
             }
-            if let Some(waker) = state.waker.take() {
-                drop(state);
-                waker.wake();
+
+            // When the file is closing registers a waker, we don't want to
+            // wake it now.
+            if let FileStatus::Open = state.file_status {
+                if let Some(waker) = state.waker.take() {
+                    drop(state);
+                    waker.wake();
+                }
             }
         })
         .detach();
@@ -1272,13 +1273,24 @@ impl DmaStreamWriter {
         cx: &Context<'_>,
     ) -> Poll<io::Result<DmaStreamReaderBuilder>> {
         let mut state = self.state.borrow_mut();
+        let previous_waker = state.waker.take();
         match state.file_status {
             FileStatus::Open => {
+                assert!(
+                    previous_waker.is_none(),
+                    "Cannot seal while flushing / writing"
+                );
                 let file = ensure_not_closed!(self.file.clone());
-                state.initiate_close(cx.waker().clone(), self.state.clone(), file, false);
+                state.add_waker(cx.waker().clone());
+                state.initiate_close(self.state.clone(), file, false);
                 Poll::Pending
             }
             FileStatus::Closing => {
+                if previous_waker.is_some() {
+                    state.add_waker(cx.waker().clone());
+                    return Poll::Pending;
+                }
+
                 state.file_status = FileStatus::Closed;
                 let file = ensure_not_closed!(self.file);
 
@@ -1325,8 +1337,15 @@ impl<'a> AsyncWrite for &'a DmaStreamWriter {
         buf: &[u8],
     ) -> Poll<io::Result<usize>> {
         let mut state = self.state.borrow_mut();
+        let previous_waker = state.waker.take();
+
         if let Some(err) = current_error!(state) {
             return Poll::Ready(err);
+        }
+
+        if previous_waker.is_some() {
+            state.add_waker(cx.waker().clone());
+            return Poll::Pending;
         }
 
         let mut written = 0;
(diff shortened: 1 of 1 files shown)


glommio/src/io/dma_file_stream.rs
```diff
@@ -1325,6 +1325,10 @@ impl<'a> AsyncWrite for &'a DmaStreamWriter {
         buf: &[u8],
     ) -> Poll<io::Result<usize>> {
         let mut state = self.state.borrow_mut();
+        if state.waker.is_some() {
```
vlovich, line 1328: This doesn't seem right in two ways. First, I believe you're supposed to always grab the waker from the provided Context if you're pending. If you just grabbed the first one, it's a recipe for never seeing the wake event. Similarly, once this gets into the polling state once, I think it remains stuck that way because the poll will always see a waker.

vlovich: Also, I would recommend adding a test to repro the failure in the issue so that we know it's fixed.

vlovich: I think the correct fix is closer to changing the code in `add_waker` to not assert that the waker `is_none` since that's not guaranteed (you may enter the poll method multiple times AFAIK).

tontinton: Maybe we need to store a `Vec` of `Waker`s? although I'm afraid it will blow up when polled a lot...

I'm not familiar enough with async rust yet, I think I'll take the time to learn it a bit more, and then think of a better solution.

tontinton: Ok I think I got to the conclusion that it is not too interesting to check for parallel writes, but I would like to hear your opinion of course.

Anyway, I updated the PR to simply remove the assert for now.

vlovich: I think the first thing is to come up with a minimal repro as a test that obviously should work and doesn’t and explain the specific sequence of events that leads to a problem (ie is this a bug in the code due to receiving a spurious wake or something else).

vlovich: Also, you mentioned keeping a a vec of Wakers. FWIW, that's not something you need to do unless you're actually managing a collection of futures. The Waker you get on a poll supersedes & invalidates all previous wakers (i.e. once you get polled, you MUST stop using the previous waker unless it's the same wake event). This confused me as well as the documentation here isn't quite so clear.

There's a `will_wake` method you can use to check if it's the same underlying Waker which can be a fast-path to avoid cloning the Waker. That's not particularly important for Glommio since the Waker's it's dealing with are single-threaded reference counts which are essentially free to clone / unref so replacing the waker blindly is fine (i.e. 2 extra integer additions are nothing in the scheme of handling the poll event). The `will_wake` optimization is more valuable if you're polling a huge number of Wakers in a runtime like tokio where it's using atomic reference counts.

As for how to track the parallel writer thing, I might explore something simple like setting a dirty bit in the write entrypoint and then clearing it when Poll::Ready is returned (& if the write entrypoint encounters the dirty bit, return an error). All that being said, I'm not the maintainer of this library, not at all familiar with DmaFileStream, & haven't really investigated how the code works / what it's intent is.

glommer: I think it's a bit dangerous to just remove the assert, without a reproducer. If you get polled twice, you can consume the buffer twice and end up with a corrupted file.

A panic is better than a corrupted file, so we should at least understand why this is happening with a bit more certainty.

tontinton: You were, right, I tried this change in the program I'm working on, and some writes were simply gone (which makes sense looking at the code).

I've uploaded a better fix, which after trying out again with my program, no writes are gone and the issue https://github.com/DataDog/glommio/issues/610 program is no longer crashing.

tontinton: Ok I thought of a better solution, where it will also fix https://github.com/DataDog/glommio/pull/612 in one go!

vlovich:
I'm still not seeing any test cases to repro the issues you mention. At first glance, it's weird to me that there's now two wakers given that the contract with poll is that only the most recently provided Waker should be remembered (i.e. if it can't be collapsed into 1 waker in the struct, that indicates there's a misuse of the Waker).

tontinton:
> If you want to add a test case that reproduces #610 and #612 (which is a prereq for any PR anyway) I can take a stab at a comprehensive fix.

Added test cases.

`join_write_multiple_writers` fails 100% of the times I ran it without my fix.

`join_close_multiple_writers` fails 50% of the times I ran it without my fix.

The reason for using 3 writers instead of just 2 is that it makes the bug more reproducible.

tontinton:
Just thought about it, maybe there is no need for the 2 wakers, I'll play with it again once I have more time.

tontinton:
> Just thought about it, maybe there is no need for the 2 wakers, I'll play with it again once I have more time.

Ok it can be done, uploaded a fix :)

vlovich:
Digging into the source code a bit, I'm wondering if there aren't additional problems with this API. One example is that the `write` is always a truncated write. i.e. if you even have 1 byte of buffer space available, it'll return Ready(1) and only write out that single byte & nothing else. Not sure if this is intended and the documentation doesn't provide clarity, but it's certainly surprising to me as it feels like `write` is returning prematurely. If not intended, there's a couple of ways to fix this:

* Change `poll_write` to return Ready only if written == buf.len().
* Change the `Write` struct that `write` returns to adjust the buffer that's passed until the full buffer is supplied.

I suspect option 2 is the better approach so that `write` correctly returns the length of the buffer once complete. With the first option it would return `Ready(buffer_size)` which isn't really helpful information.

@glommer what do you think?

All that being said, from what I've read I think the assert within `add_waker` is 100% correct and removing it allows for doing 2 writes simultaneously and them being weirdly interleaved.

I also wonder if the truncated write problem maybe is a root cause behind all the other issues you observed but I didn't spend long enough to determine that.

Here's some additional tests I've written:

* Make sure that closing prematurely results in an error instead of panicking:
```
    file_stream_write_test!(close_with_pending_write, path, _k, filename, file, {
        let mut writer = DmaStreamWriterBuilder::new(file)
            .with_buffer_size(4096)
            .with_write_behind(1)
            .build();

        let buf = [0u8; 4096];
        let mut cx = std::task::Context::from_waker(noop_waker_ref());

        assert!(matches!(Pin::new(&mut writer).poll_write(&mut cx, &buf), Poll::Ready(Ok(4096))));
        assert!(matches!(Pin::new(&mut writer).poll_write(&mut cx, &buf), Poll::Pending));

        assert_eq!(writer.close().await.expect_err("Unexpected successful closed with pending writes").to_string(), "Cannot close while flushing / writing");
    });
```

* Incomplete test (due to above description of truncated write behavior) but make sure that once we get a pending once, we still eventually complete writes.
```
    file_stream_write_test!(pending_writes_complete_eventually, path, _k, filename, file, {
        let mut writer = DmaStreamWriterBuilder::new(file)
            .with_buffer_size(4096)
            .with_write_behind(1)
            .build();

        let buf = vec![0u8; 128 * 1024 * 1024];
        let mut cx = std::task::Context::from_waker(noop_waker_ref());

        assert_eq!(buf.len(), writer.write(&buf).await.expect("Unexpected failure writing"));

        assert!(matches!(Pin::new(&mut writer).poll_write(&mut cx, &buf), Poll::Pending));
        assert!(matches!(Pin::new(&mut writer).poll_write(&mut cx, &buf), Poll::Pending));

        let r = timeout(Duration::from_millis(100), async {
            let r = writer.write(&[]).await;
            match r {
                Ok(_) => Ok(()),
                Err(e) => Err(crate::GlommioError::IoError(e)),
            }
        }).await;

        assert!(r.is_ok(), "Failed to complete writes within 100ms: {:?}", r);

        writer.close().await.unwrap();
    });
```

tontinton:
I'm currently maintaining a fork with this fix, just so I can continue working.

I think that fixing `write` should be in another PR.

The only degradation in this PR is the removal of the `add_waker` `assert` , but I feel like the benefits of allowing multiple polls overcome that.

Tell me if you want me to think about dealing with the degradation or something else, I will gladly do it, I love using `glommio` and would love to contribute more time into improving it.

vlovich:
I'm currently focused on other things at the moment. Hopefully once Glauber provides his feedback I'll be able to help guide you more.

glommer:
Directionally, dropping the existing waker and registering a new one sounds like the right thing to do.
Let me play with it a bit more, and I will comment in more details

glommer (approved):

glommer:
Merging now because we're all in agreement that we shouldn't panic in this case, and I think that dropping the previous waiter is an improvement, and the right way to deal with this problem in isolation.

@vlovich you're right about the `Ready(1)` thing. But I am honestly not sure what is the best solution is.
Unless we're going to spawn a background thread (which we shouldn't), it doesn't make that much sense to keep returning Pending, since unless you call the future again, some writes will never complete.

PR #612, opened 2023-10-13
Add a `with_truncate_on_close` to `DmaStreamWriterBuilder`

tontinton:
### What does this PR do?

Adds a `with_truncate_on_close` to `DmaStreamWriterBuilder`, to remove any padding that might have been added by glommio.

### Motivation

I wanted to get the file size of a file by reading its metadata using `file_size()`, instead of saving it to another file for example, and it was important to me to calculate how many items I have written to the file by a simple calculation instead of parsing it.

Any padding added would have ruined the calculation.

### DIFF
--- a/glommio/src/io/dma_file_stream.rs
+++ b/glommio/src/io/dma_file_stream.rs
@@ -620,6 +620,7 @@ pub struct DmaStreamWriterBuilder {
     buffer_size: usize,
     write_behind: usize,
     sync_on_close: bool,
+    truncate_on_close: bool,
     file: Rc<DmaFile>,
 }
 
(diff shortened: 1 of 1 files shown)


vlovich:
I'd love to see a unit test that shows the motivating example. I don't understand the impact this flag has because `close` consumes the `DmaStreamWriterState`. And if the flag is true, then we avoid the logic in `must_truncate` to force a truncation, but truncation should only have any effect if and only if `flushed pos > aligned pos` (i.e. `must_truncate`).

tontinton:
There's a check though in `must_truncate`:

```rust
        if let FileStatus::Closed = self.file_status {
            return false;
        }
```

Which is true when you call close on the file.

vlovich:
I see. If that's the case, I would posit that there's a higher level logical bug where that state is being set prematurely and truncation isn't getting applied correctly (assuming this is happening in a very simple happy path).

tontinton:
Closing as fixed in https://github.com/DataDog/glommio/pull/611

PR #617, opened 2023-10-15
Trimmable buffer + expose some more DmaFile internals

vlovich:
### What does this PR do?

1 small change to expose the inode / dev_major / dev_minor.
1 small change to expose DmaBuffer::trim_to_size.

### Motivation

For the dev_major / dev_minor stuff, I want to monitor NVME health (i.e. write amplification). Doing so means figuring out the physical disk that owns a given filesystem and the major / minor are starting points for that.

For the DmaBuffer, I frequently have a conservative overestimate of how much I'll end up writing. I want to make sure after I serialize that the data I'm writing to disk is trimmed to the exact bounds & the offset adjustment by the conservative estimate will simply leave gaps in the file. 

### Related issues

N/A

### Additional Notes

N/A

### Checklist

Not sure what a negative test would look like.


vlovich:
@glommer what's the distinction between alignment and fs_cluster_size? I'm seeing (at least for tmpfs) alignment is computed as 512 bytes but fs cluster is 4096. Should alignment be the larger of logical_block_size and fs cluster size?

glommer:
`fs_cluster_size` is essentially `stat`'s blksize, which is defined as 

> This [st_blksize] field gives the "preferred" block size for efficient filesystem I/O.

Filesystems have their own minimum blocks internally. You can usually write 512b, but the filesystem internally will allocate at least a page to back it, and therefore the minimum size of the file will always be 4096. I'd expect a similar thing on storage-backed filesystems, not only tmpfs.

glommer (approved):

PR #619, opened 2023-10-17
Make it more straightforward to send files across threads safely

vlovich:
### What does this PR do?

Introduce an OwnedDmaFile struct that a DmaFile can be converted into and out of. This ensures that the reactor reference that's obtained is always the one that the DmaFile is running on. This is useful as a way to dupe a file descriptor, send it to a background thread for some kind of batch processing. Similarly, the batch processing may create a new file that contains the result of the batch processing and return that file handle back to the main thread.

### Motivation

I want to send a handle to a file + various metadata to a background thread for processing (e.g. LSM compaction) and then generate a new file containing the compacted result which I then send back to the main thread for ingestion. 

### Additional Notes

I believe io_uring generally supports registering FDs although I don't believe that's used here AFAICT. If the FDs are registered, then shifting it around between io_urings would require a lot more work (or at least converting it to a sendable form would be a failable operation if the FD turned out to be a registered one).

### Checklist


glommer:
last time I tried to deal with registered API, I found it very difficult to consume, so I never bothered with it. Will look at the code today.

glommer (approved):
this is really good, thanks

PR #625, opened 2023-11-18
Peek active executors min heap instead of iterating

tontinton:
O(1) instead of O(n).


glommer:
@tontinton sorry for taking a while to get to this. I had an extremely busy couple of weeks, but with the holidays, will have some time now to go through your work.

glommer (approved):

PR #626, opened 2023-11-23
don't test for file size on resize_dma_buffer

glommer:
We already test that the write is successful, and we already test that we read what we write.

In my local machine on an ext4 filesystem, the file size never changes after we write this. This is a metadata operation and the filesystem is free to do this optimization AFAIR, which it likely does, especially since this is a tmpfile that was not closed. We'd have to have a named file and fsync() the directory to make sure the metadata always gets updated, but I don't see the value in testing that.

### What does this PR do?

### Motivation

### Related issues

for this Pull Request.

### Additional Notes

### Checklist

### DIFF
--- a/glommio/src/io/dma_file.rs
+++ b/glommio/src/io/dma_file.rs
@@ -1627,9 +1627,6 @@ pub(crate) mod test {
         let read = file.read_at_aligned(0, 2 * alignment).await.unwrap();
         assert_eq!(read.len(), alignment);
         assert!(read.iter().all(|&b| b == 1));
-
-        let stat = file.stat().await.unwrap();
-        assert_eq!(stat.file_size, alignment as u64, "{:?}", stat);
     });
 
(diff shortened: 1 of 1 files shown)


glommer:
@vlovich FYI

vlovich:
I believe you but my impression has always been that fsync is only a synchronization barrier for durability and otherwise has no effect on any operations unless there’s a kernel reboot involved.  I believe the size in particular does not need a directory fsync as it’s considered “data” metadata associated with the file itself (eg the man page says the file size is synchronized by fdatasync which makes sense for many reasons). Does adding an fdatasync before the file retrieval fix your issue?

glommer:
I'd have to look into ext4 is doing to comment properly. I think most likely this is due to the file being a temporary file. A test I could do is use a named file, then reopen it. At worst, this is a kernel bug on my particular version. But one way or another, I don't think the assert adds a lot of value, so removing is the way to go.

PR #627, opened 2023-11-23
fix test for file send

glommer:
The test is broken because we can't really guarantee that the old fd will fail to open. It is entirely possible that by now, some other test (remember tests are threaded) have opened a new file for itself that ended up with the same fd as original_fd.

This test was always failing for me when I ran "cargo test" but always succeeding if I ran the test in isolation, which confirms that this is the likely cause.


glommer:
@vlovich FYI

vlovich:
Uggh. Terrible oversight on my part. I’m usually good about stuff like
that. Thanks for tracking it down and will be more mindful going forward.

PR #628, opened 2023-11-24
Fix stall detector to be correct threshold

tontinton:
The final expected runtime, is supposed to be whatever we get from the `threshold` call.

If we add the `max_expected_time` again, it will be wrong.

Whoever CRs this, I have a question: Do we want the `threshold` method to return the time to add to `max_expected_time` or do we want the user to completely control the amount, like I did in this PR?

I think it is better to let the user always control, he can decide whether to use `max_expected_time` or not.

### DIFF
--- a/glommio/src/executor/stall.rs
+++ b/glommio/src/executor/stall.rs
@@ -60,9 +60,9 @@ pub trait StallDetectionHandler: std::fmt::Debug + Send + Sync {
     fn threshold(
         &self,
         _queue_handle: TaskQueueHandle,
-        max_expected_runtime: Duration,
+        _max_expected_runtime: Duration,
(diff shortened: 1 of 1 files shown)


tontinton:
Seems like a test broke, I'll check it tomorrow.

glommer:
`max_expected_runtime` is passed by the reactor, because the time may be different for different task queues.

The `threshold()` default implementation on line 65 seems to be the thing that's wrong, and it should be returning just 10ms.

glommer:
No, if the task queue has a quota to run for x ms, running for less than that is not a stall, technically speaking.

tontinton:
Got it, fixed

Issue #633, opened 2024-01-02
process gets stuck in D state if there is a big read (about 6 MB)


ozgrakkurt:
This in on kernel version 5.15 (Ubuntu) and the executor is created like this:
```rust
let executor = LocalExecutorBuilder::new(Placement::Unbound)
    .blocking_thread_pool_placement(PoolPlacement::Unbound(512))
    .io_memory(128 * 1024 * 1024)
    .ring_depth(8 * 1024)
    .detect_stalls(Some(Box::new(DefaultStallDetectionHandler {})))
    .make()
    .unwrap();
```

I am creating an `ImmutableFile` and then calling `read_many` on it. When I have a iovec with size more than 6MB it puts the process into D state and I can't kill it even with `kill -9` command. It is stuck using 100% cpu until I reboot the system.

I am trying to split my iovecs before calling `read_many` to see if this fixes the problem (not sure if `read_many` al ready splits big ones internally).

EDIT: verified it is fixed if I limit each read to max 8KB and call read_many like this:
```rust
let reads = file
        .read_many(
            iovs,
            MergedBufferLimit::NoMerging,
            ReadAmplificationLimit::NoAmplification,
        )
        .with_concurrency(16)
        .with_memory_limit(None);
```

ozgrakkurt:
Also might be unrelated but I am constantly getting `File dropped while still active` warning in logs even though I tried dropping file with .close in every place I can.

ozgrakkurt:
Minimal reproducer:
```rust
use glommio::{LocalExecutorBuilder, Placement, io::ImmutableFileBuilder};

fn main() {
    let executor = LocalExecutorBuilder::new(Placement::Unbound)
        .make()
        .unwrap();

    executor.run(async move {
        let file = ImmutableFileBuilder::new("foo")
            .build_existing()
            .await
            .unwrap();

        let data = file.read_at(0, 6 * 1024 * 1024).await.unwrap();
        println!("{}", data.len());
    });
}

```

ozgrakkurt:
Also with `--release` flag it hangs forever in D state but without the `--release` flag it just hangs forever and can be stopped by doing ctrl+c

ozgrakkurt:
Ok I found that this happends if the read size is bigger than half of what I get from this command
```
cat /sys/block/.../queue/max_sectors_kb
```

So this should be cought inside the library (I saw this is what is used for `DeviceMaxSingleRequest `)or maybe the read should be split to multiple reads internally?

[...]

PR #635, opened 2024-01-05
feat: handle `DmaFile::read_at` called with a size larger than `max_sectors_size`

ozgrakkurt:
### What does this PR do?

This PR tries to prevent the case of executing `DmaFile::read_at` with a larger size than `DmaFile::max_sectors_size`. Because this causes the process to hang forever in D state in some kernels (e.g. 5.15)and just hang forever in interruptible state in other kernels. (e.g. 6.1)

### Motivation

I have this case where I read pages of parquet files from disk in parallel and this problem causes my program to hang. I am able to mitigate this issue by just splitting my reads before calling read_at/read_many but this is still a problem since it took a lot of time to debug and it causes entire process to hang instead of just returning an error/splitting the reads internally. Imo there should be an error if read_at is called like this and read_many should split the big reads internally.

### Related issues

for this Pull Request.

#633

### Additional Notes

I want to also implement the internal splitting inside `read_many` if it makes sense. Also can implement internal splitting for read_at as well if error doesn't make sense. Also want to do the same stuff for write functions, I am guessing the same thing happens with those as well.

### Checklist

### DIFF
--- a/glommio/src/error.rs
+++ b/glommio/src/error.rs
@@ -235,6 +235,18 @@ pub enum GlommioError<T> {
 
     /// Timeout variant used for reporting timed out operations
     TimedOut(Duration),
+
+    /// The read/write size is bigger than allowed by the block device.
+    ///
+    /// Single read/write size shouldn't exceed the value in /sys/block/.../queue/max_sectors_kb.
+    ///
+    /// This only happens when doing direct IO
+    MaxIOSizeExceeded {
+        /// Size of the requested IO operation.
+        size: usize,
+        /// Single request limit of the underlying block device.
+        limit: usize,
+    },
 }
 
 impl<T> From<io::Error> for GlommioError<T> {
@@ -289,6 +301,9 @@ impl<T> fmt::Display for GlommioError<T> {
             },
             GlommioError::ReactorError(err) => write!(f, "Reactor error: {err}"),
             GlommioError::TimedOut(dur) => write!(f, "Operation timed out after {dur:#?}"),
+            GlommioError::MaxIOSizeExceeded { size, limit } => {
+                write!(f, "Max IO size exceeded when doing direct IO. Expected a max size of {limit} but got {size}")
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


ozgrakkurt:
Also is it at least possible to make `pub fn max_sectors_size() -> usize`  on both `DmaFile` and `ImmutableFile` ?

glommer:
There's something going on here. I don't think anything is supposed to hang when I/O is too large, and this smells a bit like a bug (where, is the question)

glommer:
FWIW, that file on /sys is read-write, so it doesn't really mean any intrinsic I/O size limit from the perspective of the filesystem. The kernel should be able to split transfers larger than that. Have you tried running your example in different kernels?

ozgrakkurt:
> FWIW, that file on /sys is read-write, so it doesn't really mean any intrinsic I/O size limit from the perspective of the filesystem. The kernel should be able to split transfers larger than that. Have you tried running your example in different kernels?

I ran with 5.15, 6.1 and 6.2. On 5.15 and 6.2, the process goes into D state and hangs forever in uninterruptible state. On 6.1 it hangs forever but can be shut down with ctrl+c. I'll try newest kernel as well. Which kernel do you normally run on/test?

ozgrakkurt:
> > FWIW, that file on /sys is read-write, so it doesn't really mean any intrinsic I/O size limit from the perspective of the filesystem. The kernel should be able to split transfers larger than that. Have you tried running your example in different kernels?
> 
> I ran with 5.15, 6.1 and 6.2. On 5.15 and 6.2, the process goes into D state and hangs forever in uninterruptible state. On 6.1 it hangs forever but can be shut down with ctrl+c. I'll try newest kernel as well. Which kernel do you normally run on/test?

actually nevermind. It hangs in D state if compiled with `--release` flag but can be exited with ctrl-c if compiled in dev mode

ozgrakkurt:
It does work on kernel `6.6.10-zabbly+` installed on Debian so maybe can close this

ozgrakkurt:
closing this as it seems to be a kernel bug and the hanging doesn't happen on newer kernel (6.6) for me

PR #649, opened 2024-04-12
Fix stat to use the fd instead of the path

### LINKED ISSUE #648: Statx is invoked on path instead of fd
As title says, it looks like the statx call is done on the path instead of the fd directly. This opens up race conditions &  breaks the state reported for O_TMPFILE because the directory stat is returned instead of the unnamed file.

vlovich:
### What does this PR do?

Fixes #648. Noticed that the reported inode for two sibling O_TMPFILE was identical because the stat was being invoked on the parent directory (i.e. stat never worked for temporary files). Similarly, you can now remove the path backing a file and stat continues to work.

### Motivation

2 sibling O_TMPFILE files having the same inode & being unable to get the size of a temporary file.

### Checklist

### DIFF
--- a/glommio/Cargo.toml
+++ b/glommio/Cargo.toml
@@ -26,7 +26,7 @@ buddy-alloc = "0.4"
 concurrent-queue = "1.2"
 crossbeam = "0.8"
 enclose = "1.1"
-flume = { version = "0.10", features = ["async"] }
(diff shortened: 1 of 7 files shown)


glommer (approved):

glommer:
tnx. Will merge, but format tests failing

PR #653, opened 2024-04-16
Fix first Reactor sleep taking at least 30-50ms

### LINKED ISSUE #652: Timer adds a surprising amount of delay vs thread::sleep
For the following test case I see `std::thread::sleep` take ~11ms but `Timer` can take up to 40+ms. That's a surprising amount of added latency and I don't see anything in the docs calling this out as intended behavior so figured I'd file a bug while I investigate.

```
    #[test]
    fn timer_duration() {
        test_executor!(async move {
            let start = Instant::now();
            std::thread::sleep(Duration::from_millis(10));
            let elapsed = start.elapsed();
            assert!(elapsed >= Duration::from_millis(10), "Thread sleep returned too soon: {elapsed:?}");
            assert!(elapsed <= Duration::from_millis(30), "Thread sleep took way too long to run: {elapsed:?}");
            eprintln!("Thread sleep took {:?}", elapsed);

            let start = Instant::now();
            Timer::new(Duration::from_millis(10)).await;
            let elapsed = start.elapsed();
            assert!(elapsed >= Duration::from_millis(10), "Timer expired too soon: {elapsed:?}");
            assert!(elapsed <= Duration::from_millis(30), "Timer took way too long to run: {elapsed:?}");
        });
    }
```

vlovich:
Noticed this problem in unit tests where the very first timer sleep would take a very long time because the membarrier was being registered and tests have more than 1 thread.

Initializing the membarrier strategy on Reactor::new seems like a good idea because it's highly likely any use of the Reactor will involve going to sleep and I can't imagine a use-case where that's not the case unless you're not interacting with anything within Glommio in the first place.

With this change we see that the very first timer now completes within 11ms (I added a grace window of an extra ms in case of CI).

### What does this PR do?

The first construction of the IO uring reactor now registers the membarrier.

### Motivation

I had tests that were flakily failing because a very short sleep was taking an almost unbounded amount of time (observed up to 60ms even for a 5ms sleep). I think it's cleaner if the membarrier initialization is front-loaded into Reactor construction vs paying that penalty on the very first .await later where it's a bit more non-obvious and surprising. 

### Related issues

Fixes #652 

### Additional Notes

I believe the cost of registration when there's > 1 thread running got worse sometime after Linux 6.6 because tests in my project that do something similar started regularly taking >30ms after upgrading to 6.8 whereas before they were mostly succeeding < 30ms (it was my grace window for how long a 10ms sleep could take).

### Checklist

### DIFF
--- a/glommio/src/executor/mod.rs
+++ b/glommio/src/executor/mod.rs
@@ -93,6 +93,36 @@ pub fn executor() -> ExecutorProxy {
     ExecutorProxy {}
 }
 
+/// You probably don't need to call this explicitly unless you have created threads before
+/// you constructed an executor. You may also want to call this explicitly if you're sandboxing
+/// your app and dropping permissions required to use privated expedited membarrier commands to
+/// improve Glommio performance.
+///
+/// This run some early initialization that may save ~30-80ms if you create threads before you
+/// tough Glommio code & don't otherwise register with membarrier. This is an idempotent call
+/// that can be invoked as many times, but only really makes sense to do early in main before
+/// you've constructed any threads. Additionally, the explicit membarrier registration this
+/// attempts is also useful if you're later dropping the privilege required to using private
+/// expedited commands so that the membarrier synchronization strategy can be used within
+/// the executor hot loop.
+///
+/// The detailed motivation is that internally glommio prefers to use [membarrier](https://man7.org/linux/man-pages/man2/membarrier.2.html)
+/// as a high performance synchronization barrier of the underlying io_uring memory
+/// shared with the kernel (if the feature is available). If there exist any threads in the program,
+/// registering the barrier has been observed to take a relatively long and highly variable amount
+/// of time (as high as 30-80ms). If we register before any threads are constructed, this takes almost no time.
+/// That's why this can be useful if you're integrating Glommio into an existing application to move this to
+/// the front of main.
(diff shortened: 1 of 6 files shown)


vlovich:
Ping on this in case you missed it in your review batch @glommer

glommer:
I had missed it, indeed.

glommer:
Can we add a comment on the code explaining why we're doing this, so nobody else removes that in the future?

vlovich:
Good note. I've added much more extensive documentation for the subtleties. I also realized that it's placement was incorrect as it needed to preceed the construction of threads in BlockingThreadPool. I also realized it's necessary as an external API (or we can add a dependency on the ctor crate to do this transparently) for 2 reasons:

1. The user may start their own threads before glommio
2. The user may sandbox their app before glommio & thus glommio will get a privilege denial for trying to register the membarrier

I don't know the policy on adding dependency on 3p crates so I went with an explicit API that users can call if it's needed for them (most don't) but ctor magic would be more user-friendly although portability may be a concern (I think it's well supported on tier 1 platforms but I'm sure there's always subtleties that can appear with that kind of low-level stuff).

glommer (approved):

glommer:
As I get older, I dislike magic more and more.

Issue #650, opened 2024-04-12
RFC: 0-copy support for writing what was written


vlovich:
If I'm copying from 1 file to another, currently I have to allocate a `DmaBuffer` & `copy_from_slice` the `ReadResult` into the `DmaBuffer`. @glommer would it make sense to extend `DmaBuffer` to have a `From<ReadResult>` trait and extend it's internal storage enum to hold a `ReadResultInner`? Or is this a bad idea for some reason?

glommer:
As long as ownership is clear, I don't see anything wrong with that. By the time it materializes on our side, the kernel is done with it already so there's no risk there.

The second one is more complex. I had issues before with the kernel and us not agreeing with what "the duration of the write" is. The standard poll methods are bad for it, because they don't guarantee that the same buffer is used in two invocations if the write doesn't complete. If we go for a specialized API, may work.

vlovich:
I'm not sure I follow the explanation about the second case. You're saying the completion event comes back but the kernel keeps using the referenced buffer? I'm sure I'm understanding that wrong because that would apply equally to the DmaBuffer case I would imagine which needs to know when to free the underlying buffer.

It sounds like something more nuanced though because you mention a second write and polling but I'm not familiar enough with io_uring to follow. What do you mean by specialized API?

PR #655, opened 2024-04-18
Allow a ReadResult to be converted to a DmaBuffer

vlovich:
### What does this PR do?

Data read from one spot can be passed directly to write it elsewhere without any interim buffers.

### Motivation

0-copy shuttling of data to disk.

### Related issues

#650 

### Additional Notes

This is a hack but it let's us write reads from files without creating an interim copy. The reason it's a hack is that it's weird to have a DmaBuffer, which represents a mutable view of data, be convertible from a ReadResult - notice the panic in attempts to access it mutably.

I would rather have a new struct called `ImmutableBuffer` that internally looks identical to `DmaBuffer` and implements From<DmaBuffer>, From<ReadResult>. This way it's impossible to do DmaBuffer::from(read_result).as_mut_slice() which panics at runtime. This would be a source code change but I think given that the functions would be generic over `T: Into<ImmutableBuffer>`, maybe it's OK? For now stuck to what was discussed in the RFC.

I know there's formatting errors but I want to make sure we don't want to introduce the new ImmutableBuffer struct & generic write_at to cleanup the hack.

### Checklist

### DIFF
--- a/glommio/src/io/dma_file.rs
+++ b/glommio/src/io/dma_file.rs
@@ -804,7 +804,7 @@ pub(crate) mod test {
     use futures::join;
     use futures_lite::{stream, StreamExt};
     use rand::{seq::SliceRandom, thread_rng};
-    use std::{cell::RefCell, convert::TryInto, path::PathBuf, time::Duration};
+    use std::{cell::RefCell, convert::TryInto, ops::Deref, path::PathBuf, time::Duration};
 
     macro_rules! dma_file_test {
         ( $name:ident, $dir:ident, $kind:ident, $code:block) => {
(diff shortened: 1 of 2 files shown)


vlovich:
Kk. Updated with formatting fixes.

glommer:
ok. There are some conflicts, though.

vlovich:
Yeah. Sometimes I wish each test was a separate file so that merge conflicts due to adding tests weren't a thing.

vlovich:
Fixed.

glommer:
still failing

vlovich:
A `})` line got lost resolving the conflict. Sad that I'm still having to rely on text-based language-agnostic merge resolution tools.

glommer:
When I was working at RedHat in 2005 I had this idea of a patch tool that would parse the AST and show you a summary of what changed, so you could read things like "This function was moved somewhere else". Totally impossible back then. Maybe possible now ?

glommer (approved):

vlovich:
Yeah, every engineer I've talked to has had the same idea about an AST-aware diff tool. We also intuitively know how hard it is which is why we don't bother trying to start companies around the idea and tackle easier problems like distributed databases :D.

Looks like SemanticMerge supports rust in some fashion although I don't know the state of it based on their website https://docs.plasticscm.com/semanticmerge
I should try it out.

There's also https://semanticdiff.com/ which can't be used for semantic merging but maybe could have caught this after the fact as a way to confirm the merge resolution was done correctly.

Would be nice if tooling made this easier.

PR #656, opened 2024-04-19
fix: export publicly 'StallDetection'

aurelilys:
Hello,

### What does this PR do?

Here a little mr for publicly export the struct 'StallDetection'.
I'm sorry if the struct documentation isn't very thorough, but I'm open to modifying it.

### Motivation

I wanted to implement 'StallDetectionHandler' in my project, but unfortunately the struct 'StallDetection' is not publicly exported.

### DIFF
--- a/glommio/src/executor/stall.rs
+++ b/glommio/src/executor/stall.rs
@@ -16,13 +16,20 @@ use std::{
     time::{Duration, Instant},
 };
 
+/// Store information about detected stall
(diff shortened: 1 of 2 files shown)


vlovich:
@glommer seems reasonable to me. I imagine the motivating use-case is to log the structured message to an external system like Sentry.

glommer (approved):

Issue #646, opened 2024-03-28
RFC: A Send safe `DmaFile` to let the fd be shared across threads


vlovich:
Because all the APIs work on specific locations within the fd, the `fd` owned by `DmaFile` is in some way Send safe as there's no kernel state that could be mutated unsafely with the exception of closure (discussed below). I want a lighterweight mechanism to have other threads reference the FD without needing to `dup`.

I have a use-case where I sometimes return a fd + reference to a file region in response to a `read` request that can be lazily read on an arbitrary thread (or have sendfile invoked on it). Right now, I'm doing this by invoking `dup` on `DmaFile` to grab a reference. However, the problem with this is that aside from needing a synchronous syscall, I risk running into resource exhaustion with the number of open files (e.g. if I'm doing 2M reads/s, if even 10% hit this path that's 200k duplicate fds in flight) which is a lot of kernel memory for no real value.

I'd like some kind of lighterweight mechanism where I can share access to the fd without involving the kernel. There's two options that come to mind for me.

Option 1: Change the internals around to store `Arc<RawFd>` & introduce `BorrowedDmaFile` counterpart to `OwnedDmaFile`. Either `BorrowedDmaFile` could be converted into a `DmaFile` that doesn't auto-close / disallows invoking close or we introduce a new `DmaFileRef` class that `BorrowDmaFile` can be converted into (the former is a smaller refactor & less code duplication, the latter feels potentially cleaner). 

Option 2: Implement `FromRawFd` for `DmaFile` similar to how `GlommioFile` does. Not sure it's actually possible since `FromRawFd` doesn't let us control whether or not the DmaFile auto closes on drop.

I feel like option 1 is cleaner since it ensures the file was opened by Glommio consistently rather than just taking over ownership of any old fd. It also provides better opportunity to inject some safety into this mechansim. The main question is about safety with respect to file closure. One possible technique would be to check `strong_count` when the owning `DmaFile` is dropped / has close invoked on it - if it's 1 then close normally, otherwise add the FD for a deferred closure once it's count drops to 0 (how the original thread notices it & risks of resource leaks because the owning thread finishes with other threads that hold non-owning references still running). There may be ways to deal with this, or we can just declare the `borrow` API `unsafe` and document that you have to be supremely careful about file closures.

[...]

PR #657, opened 2024-04-24
Make DmaFile and OwnedDmaFile clonable

vlovich:
### What does this PR do?

DmaFile and OwnedDmaFile can now be cloned instead of just `dup`. This lets us safely share the FD for DmaFile with background threads. I limited this to just DmaFile because DmaFIle doesn't have a remembered cursor position. This could probably also be replicated for BufferedFile but I'd rather there be a motivating use-case to expose that functionality.

### Motivation

The current mechanism for sharing an FD with a background thread requires `dup` which can limit how many references to a file can exist concurrently in background threads. The use-case is a file server that responds to requests by handing back an FD and a region to respond with
- using `dup` limits how many concurrent requests can be referencing the file at once due to kernel limits on simultaneous open FDs. However, by letting DmaFile be clonable, it lets us send the FD to any thread without risking race conditions about the FD closing while references exist.

This does have a new consequence that close requests may fail if clones exist, but this only impacts those who actually try to clone in the first place, so should have no impact on existing users.

### Related issues

#646 

I went with a simpler version of just making the relevant types cloneable. That felt cleaner than whatever I was originally going for with `BorrowedDmaFile`. I also added a polling interface to try to consume the last reference which can be useful if you are implementing a graceful close mechanism on one thread (i.e. you promise to not hand out any more references to the FD & poll closing) with references potentially held open on other threads.

### Checklist


glommio/src/io/glommio_file.rs
```diff
@@ -47,12 +48,12 @@ pub(crate) struct GlommioFile {
...
 impl Drop for GlommioFile {
     fn drop(&mut self) {
-        if let Some(file) = self.file.take() {
+        if let Some(file) = self.file.take().and_then(Arc::into_inner) {
             debug!(
                 "File dropped while still active. ({:?} / fd {}).
 That means that while the file is already out of scope, the file descriptor is still registered \
                  until the next I/O cycle.
-This is likely file, but in extreme situations can lead to resource exhaustion. An explicit \
+This is likely fine, but in extreme situations can lead to resource exhaustion. An explicit \
```
vlovich, line 56: Someone had files on the brain when writing the comment :smile:

glommer (approved):

PR #658, opened 2024-04-24
Implement a weak reference to a DmaFile

vlovich:
### What does this PR do?

Introduces `WeakDmaFile` to be able to construct a non-owning reference that can be upgraded to an owning one.

### Motivation

When I went to migrate my current usage of wrapping DmaFile with Rc, I realized there's a spot I use weak Rc. So instead of having a weird pattern of Rc over something using Arc internally, it would be nice to use a proper weak reference to the DmaFile.

### Related issues

#646

### Checklist

### DIFF
--- a/examples/hyper_client.rs
+++ b/examples/hyper_client.rs
@@ -170,7 +170,7 @@ mod hyper_compat {
             while let Some(next) = response.frame().await {
                 let frame = next.unwrap();
                 if let Some(chunk) = frame.data_ref() {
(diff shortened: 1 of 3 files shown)


glommer (approved):

vlovich:
I haven't incorporated running clippy into my muscle memory. Should start doing that on my project. Fixed.

vlovich:
Fixed my vscode settings for this project so hopefully won't come up again (clippy run by rust-analyzer & highlights problems inline on save).

Issue #663, opened 2024-05-11
Any interest in adding advisory lock support?


vlovich:
Adding advisory locks outside of Glommio is unsafe because it's a single status bit on the fd - i.e. locking shared & then locking exclusively works, but it's dangerous to coordinate unlocking safely as the first lock to unlock unlocks the file. Doing this safely needs to happen within Glommio where it can hold some status about the state & coordinate acquiring a lock safely.

This adds some extra state but hopefully that's not the main concern. There's two approaches for tackling this.

The first is the most featureful where we can try to implement a thread-safe RwLock-like implementation for advisory locks. After thinking about it though, that seems kinda complicated as it involves having a way to signal across threads. I think the only place right now where that happens is the task queue management and it's not a reusable piece of code - likely I'd want to pull in AtomicLInk from intrusive-collections to make this easier. But even with that, it's tricky to implement correctly.

The alternative implementation is that we only acquire 1 lock to be acquired at a time & simply return an error if anyone else has started trying to acquire the lock. That's much simpler to reason about and implement and likely fits with typical usage where you want to make sure there's exactly 1 owner of the advisory lock. I would prefer the alternative implementation as it's less likely to contain bugs & the feature set is still useful.

The returned guard would hold a clone of the lock to ensure that it's impossible to close the FD until the lock is released / dropped. There's either an `unsafe fn leak` method on the guard that consumes self & drops the file letting close work or a `close_locked(AdvisoryLock)` method on the DmaFile to give ownership of the lock back to the file to close - I think `close_locked` makes the most sense but I could see `leak` also being useful so maybe both?

The APIs added to DmaFile would be:

```
/// If the file is currently locked in the kernel, an error is returned immediately without blocking indefinitely. This returns an
/// error if a lock has already been acquired (or is attempting to be acquired) on this file handle instance (including duped instances).
async fn try_lock_exclusive(&self) -> Result<AdvisoryFileGuard>;
/// If the file is currently locked exclusively in the kernel, an error is returned immediately without blocking indefinitely. This returns an
/// error if a lock has already been acquired (or is attempting to be acquired) on this file handle instance (including duped fds). /// Otherwise a handle holding the shared lock is returned.
async fn try_lock_shared(&self) -> Result<AdvisoryFileGuard>;
/// If the file is currently locked in the kernel, the function will wait indefinitely until an exclusive lock is acquired. This
/// returns an error if a lock has already been acquired on this file handle (including duped instances).
[...]

Issue #448, opened 2021-10-20
Eventfd not closed after executor finish


thirstycrow:
As the following benchmark runs, there's an increasing number of eventfds listed by `lsof`. With 2 new executors created in each round of test, there's 12 more open eventfds exists after the executors finish.

```rust
use glommio::channels::shared_channel;
use glommio::prelude::*;
use std::sync::mpsc::sync_channel;
use std::time::{Duration, Instant};

fn test_spsc(capacity: usize) {
    let runs: u32 = 10_000_000;
    let (sender, receiver) = shared_channel::new_bounded(capacity);

    let sender = LocalExecutorBuilder::new()
        .pin_to_cpu(0)
        .spawn(move || async move {
            let sender = sender.connect().await;
            for _ in 0..runs {
                sender.send(1).await.unwrap();
            }
            drop(sender);
        })
        .unwrap();

    let receiver = LocalExecutorBuilder::new()
        .pin_to_cpu(1)
        .spawn(move || async move {
            let receiver = receiver.connect().await;
            for _ in 0..runs {
                receiver.recv().await.unwrap();
            }
        })
        .unwrap();

    sender.join().unwrap();
    receiver.join().unwrap();
}

fn main() {
    for i in 0..10000 {
        println!("==========");
        println!("Round {}", i);
        //test_spsc(10);
        test_spsc(100);
        test_spsc(1000);
        test_spsc(10000);
    }
}
```

glommer:
As a status update, I spent some time trying to fix this, but it is really hard because tasks often get destroyed under our nose. This brought me back to the refcount hell in the task structures. I'll keep looking at it.

zaidoon1:
is there a workaround to this bug? I'm hitting this issue too and maxing out my fd limit (65K) & memory.

glommer:
This turned out to be pretty hard to fix, so I didn't get back to it =(
There's currently no workaround, unfortunately.

I will notice, though, that creating executors is a fairly expensive operation, and this bug manifests with executors being created / destructed very frequently. Outside of test cases - like described in this bug, I'd recommend you work with long-lived executors when possible, which would not manifest the issue.

glommer:
My recommendation would be to try to move the timer logic inside the executor as well.
The executor is essentially a thread, so you can spawn it early, and move stuff inside.

[...]

PR #664, opened 2024-05-11
Implement very simple advisory lock mechanism

vlovich:
### What does this PR do?

Introduces very basic advisory lock integration. As I mentioned in the issue, going for a very simple non-fancy approach that only lets you acquire a single advisory lock per file entry.

### Motivation

Implementing a database & would like to use a safe version of advisory locks to implement mutual exclusion for writers. It's not possible to implement advisory locking in pure safe code outside because the resulting usage becomes unsound very easily (the Rust advisory-lock crate pretends it's safe but I'd argue that crate is unsound because it doesn't guarantee the usage of the lock is correct within the context of the current process).

### Related issues

#663 

### Additional Notes

#448 prevents us from implementing "blocking" versions that wait forever for the lock to become available. The reason is that it would require spawning a new thread which would leak memory (it would also be horribly slow & interact poorly with cancellation).

### Checklist

### DIFF
--- a/glommio/src/io/glommio_file.rs
+++ b/glommio/src/io/glommio_file.rs
@@ -27,6 +27,113 @@ pub(super) type Device = u64;
 pub(super) type Inode = u64;
 pub(super) type Identity = (Device, Inode);
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AdvisoryLockState {
+    Unlocked = 0,
+    Locking = 1,
+    Locked = 2,
+}
+
+struct AdvisoryLockStateGuard<'a>(&'a AdvisoryLockStateHolder, bool);
+
+impl AdvisoryLockStateGuard<'_> {
+    fn commit(mut self) {
+        self.1 = false;
+        self.0.mark_locked();
+    }
+}
+
+impl Drop for AdvisoryLockStateGuard<'_> {
+    fn drop(&mut self) {
+        if self.1 {
+            self.0.abort_locking();
+        }
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct AdvisoryLockStateHolder {
+    state: std::sync::atomic::AtomicU8,
+}
+
+impl std::fmt::Debug for AdvisoryLockStateHolder {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("{:?}", self.state(),))
+    }
+}
+
+impl AdvisoryLockStateHolder {
+    fn map_state(state: u8) -> AdvisoryLockState {
+        match state {
+            0 => AdvisoryLockState::Unlocked,
+            1 => AdvisoryLockState::Locking,
+            2 => AdvisoryLockState::Locked,
+            rest => {
+                panic!("Memory corruption resulting in unexpected state value? state = {rest:?}")
+            }
+        }
+    }
+
+    fn state(&self) -> AdvisoryLockState {
+        Self::map_state(self.state.load(std::sync::atomic::Ordering::Relaxed))
+    }
+
+    fn mark_locking<'a>(&'a self, file: &Path) -> Result<AdvisoryLockStateGuard<'a>> {
+        self.state
+            .compare_exchange_weak(
+                AdvisoryLockState::Unlocked as u8,
+                AdvisoryLockState::Locking as u8,
+                std::sync::atomic::Ordering::Relaxed,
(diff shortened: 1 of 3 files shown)


glommio/src/io/glommio_file.rs
```diff
@@ -205,6 +356,77 @@ impl GlommioFile {
...
+
+    pub(crate) async unsafe fn funlock(&self) -> Result<()> {
+        let lock_state = self.lock_state.as_ref().unwrap();
+        assert_eq!(
+            lock_state.state(),
+            AdvisoryLockState::Locked,
+            "Not holding the lock!"
+        );
+
+        let source = self
```
vlovich, line 410: One thing I've noticed here is that the background thread dispatch is quite expensive (the syscall takes microseconds but the round-trip to the background thread takes >10ms on my 3.9 GhZ CPU & the background thread should be idle). While that's fine for acquiring the lock (can argue the latency is just the cost of acquisition), it's a bit unfortunate for unlock - not sure if I should do the funlock explicitly since `try_take_last_clone_unlocking_guard ` is much slower than drop + try_take_last_clone (except that pattern isn't quite as safe if you're trying to enforce a clean shutdown).

glommer: is the main cpu busy? My suspicion is that you have a too high poll period, and if the main thread is busy, it will not know that it is completed until poll happens.

glommer: I am not very familiar with advisory locks in practice. Never used them, so it's hard for me to talk about the caveats. But it seems to me that:
* you have state in userspace that tracks whether the lock is held (which is great, btw)
* you will only try to acquire the lock in the kernel when you already know you will succeed.

If that's the case, maybe we can just call the blocking call? Blocking for microseconds is absolutely not a problem, especially if this is a lock (not called thousands a time per second). You'd want this to happen on the blocking thread if this is an operation that can take dozens of milliseconds, potentially.

Is there any case where we are concerning that locking (or unlocking) could take more than 1ms ?

vlovich: Benchmarking it on my machine (which isn't the best but it's the tool I have without resorting to reading the Linux kernel source), it's ~100ns. I originally had the blocking version also in this PR but that approach doesn't work because we can easily deadlock since we run out background threads to distribute work to. Probably no way to implement those without proper I/O uring support (or doing something silly like doing a try lock & then spawning a completely new thread just for waiting if it fails).

So yeah, I'll move these back onto the main thread.

vlovich: Actually, just want to confirm. While it's fine for local files, this may show problems for NFS/SMB shares of which I don't have any to really test since those will map to network I/O. Handle it via documentation or some other way?

vlovich: I could use fstatfs to determine if it's NFS/SMB & fall back to the blocking path in that case.

vlovich: I'm going to assume NFS/SMB is an unlikely use case for DmaFile for now so leaving it out. Can always add a fstatfs call later to distinguish network FS from local at the cost of an extra syscall in the advisory lock path.

glommer:
Is this ready to go ? Looks good on a new review.

In related news, I am considering removing the clippy tests from CI. I still love the idea, but we struggled a lot already trying to figure out a good way to pin versions to make sure it stays stable, yet it keeps breaking. Do you have any thoughts ?

vlovich:
I thought it was but need to fix failing tests.

I actually don't mind the clippy stuff in CI. Keeps the codebase consistent which outweighs the nitpickiness annoyance of it for me personally. Probably could do with a pre push hook that runs the required clippy lints and doctests before so that I don't churn so much on it though :)

If we're running clippy on a floating version (eg nightly or latest stable) then we could just not do that and stick to running clippy only on pinned versions at the risk of having more work when bumping the minimum / maximum rust version we build against (+ having downstream projects get more complaints if they're running newer than the version we test against). All options come with trade offs 😃

glommer:
yeah, the pre-hooks may work. I'll take a look at it later this week.
It was supposed to be a pinned version, but somehow it keeps being updated.

vlovich:
BTW the failure here was just a normal compile error :).

vlovich:
@glommer since right now CI is bitrotted in master, I've rebased this branch onto #680 to verify this PR passes CI. PR is ready for merging but may require a rebase once #680 is merged.

glommer:
I just merged the other PR.

vlovich:
Rebased - ready for merge.

glommer (approved):

PR #665, opened 2024-05-28
Define CFLAG -D_GNU_SOURCE as per liburing ./configure

yokomizor:
### What does this PR do?

The `./configure` script for liburing sets `CFLAG=-D_GNU_SOURCE`, so we should set it on our `build.rs` as well.

### Motivation

Fixes the following build error with musl libc:
`error: unknown type name 'cpu_set_t'`

On musl, `cpu_set_t` is only defined when _GNU_SOURCE is set.

### Related issues

none

### Additional Notes

none

### Checklist

### DIFF
--- a/glommio/build.rs
+++ b/glommio/build.rs
@@ -32,6 +32,7 @@ fn main() {
         .file(src.join("queue.c"))
         .file(src.join("syscall.c"))
         .file(src.join("register.c"))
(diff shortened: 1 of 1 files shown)


glommer (approved):

glommer:
thanks!

PR #680, opened 2025-04-16
Fix CI bitrot

vlovich:
cache@v2 was deprecated a while back and all builds are failing.
Additionally cargo deny had been updated in a backward-incompatible way bitrotting the current config.
Finally some of the packages have been disallowed so update to latest crates to  fix that issue.

### What does this PR do?

Make CI pass on master again.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
           cat << EOF > "run-gha-workflow.sh"
           PATH=$PATH:/usr/share/rust/.cargo/bin
           echo "`nproc` CPU(s) available"
-          rustup install 1.65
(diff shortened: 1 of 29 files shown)


vlovich:
@glommer test_spin seems to be quite flaky, particularly in GitHub CI - I'm guessing the assertion is brittle when there's a lot of concurrent tests being run on a not so powerful machine?. Maybe the way it's written could be improved to be more robust? For now I just reduced the timings.

glommer:
yes, those tests are all very timing dependent, which is terrible. The right way to do it is to have a simulated clock, but I always kept deferring it.

glommer (approved):

Issue #641, opened 2024-03-11
What's the problem with performance, implementation or usage?


TheLudlows:
https://github.com/bytedance/monoio/blob/master/docs/en/benchmark.md

As described in the test.

![bench](https://raw.githubusercontent.com/bytedance/monoio/master/.github/resources/benchmark/monoio-bench-100B.png)

glommer:
Likely implementation.
For example, I have made the explicit decision of keeping a hash of the completion entries instead of just casting their address as unsafe.

That's a decision I don't regret, since early io_uring code was full of issues that essentially led to wrong addresses being added there, completions disappearing, etc. I am sure it's less of an issue now, but I have never done the work to methodically go chase performance issues.

TheLudlows:
It looks like it's almost 30% worse，is any way to improve it?

vlovich:
Discussed previously in #554. Probably just needs someone to run things under a profiler to figure it out. I know that monoio relies on nightly features (e.g. fast thread local) and that could also be contributing to the performance difference (it doesn't require it anymore but the benchmarks are run against nightly with that feature on).

Would be interesting to hear from @ihciah

bryandmc:
Their implementation also uses fastpoll which is much better than what we do which is poll+read if I am remembering all this correctly. Honestly the way to resolve this now is probably to re-do sockets with all the new io_uring features that have become available.. Of which I would include buffer select, buffer rings, fastpoll, and zc where possible. Also worth checking the Semaphore implementation as they mention in #554.

vlovich:
@bryandmc do you know if this impacts disk I/O performance at all or if this is just issues in the net stack?

bryandmc:
@vlovich from an io_uring perspective, we already do most (if not all?) of the things that ensure fast disk reads. The additional performance could be obtained through profiling, etc, but unlike the net stack I don't think there are features we have "left on the table" that we aren't currently using.. Because of that, I would defer to @glommer explanation, which is just that it hasn't been optimized at all. Probably some easy performance wins for anyone with a little time and a profiler..

fsaintjacques:
@bryandmc, well, it's missing multishot receive which can be a big win.

Issue #554, opened 2022-05-16
why is glommio slower on benchmark against bytedance's monoio?


hiqsociety:
why is glommio slower on benchmark against bytedance's monoio?
https://github.com/bytedance/monoio

bryandmc:
Because they are using fastpoll, probably.. at least I think that's why there was faster if I remember correctly. I had an alpha fastpoll branch and we were faster so maybe that's the different. 

We aren't sure when we will have support for faster sockets... it's something I want to do but haven't had time since the birth of my child. In time, we will get it done. Patches welcome!

nazar-pc:
I just did random file read testing of glommio vs monoio. Not only `Semaphore` from `glommio` appears to be significantly slower than generic `async_lock::Semaphore`, monoio was 11.1x faster when reading random ~40kiB chunks across SATA SSD (Samsung SSD 860 EVO 1TB) with io concurrency set to 32.

In my code the abstraction looked like this (requires Rust Nightly):
<details>

```rust
pub trait ReadAtSync: Send + Sync {
    /// Fill the buffer by reading bytes at a specific offset
    fn read_at(&self, buf: &mut [u8], offset: usize) -> io::Result<()>;
}

pub struct GlommioFile<'a> {
    file: &'a Rc<DmaFile>,
    semaphore: Semaphore,
}

impl ReadAtAsync for GlommioFile<'_> {
    async fn read_at(&self, mut buf: Vec<u8>, offset: usize) -> io::Result<Vec<u8>> {
        let _permit = self.semaphore.acquire().await;
        let read_result = self
            .file
            .read_at(offset as u64, buf.len())
            .await
            .map_err(|error| {
                io::Error::new(
                    io::ErrorKind::Other,
                    format!("Failed to read with glommio: {error}"),
                )
            })?;
        buf.copy_from_slice(&read_result);
        Ok(buf)
    }
}

impl<'a> GlommioFile<'a> {
    pub fn new(file: &'a Rc<DmaFile>, io_concurrency: usize) -> Self {
        Self {
            file,
            semaphore: Semaphore::new(io_concurrency),
        }
    }
}

pub struct MonoioFile<'a> {
    file: &'a File,
    semaphore: Semaphore,
}

impl ReadAtAsync for MonoioFile<'_> {
    async fn read_at(&self, buf: Vec<u8>, offset: usize) -> io::Result<Vec<u8>> {
        let _permit = self.semaphore.acquire().await;
[...]

PR #684, opened 2025-05-28
Add IORING_FEAT_FAST_POLL

redactedontop:
### What does this PR do?

This PR adds FAST_POLL, which while requires bumping kernel MSVC, allows for performance on-par with monoio.

### Motivation

I wanted glommio to be on par with monoio performance-wise.

### Related issues

#554
#641


vlovich:
Is there a way to turn on the feature only when the kernel supports it?

Issue #605, opened 2023-09-23
DmaFile returns EOPNOTSUP when reading (error code 95)


vlovich:
I'm sure I must be doing something wrong but I've been staring at this a few days and can't figure out what it might be. I'm consistently failing to read from DmaFile for some reason:

```
thread 'tests::basic' panicked at 'Failed to read what we just wrote: EnhancedIoError { source: Os { code: 95, kind: Uncategorized, message: "Operation not supported" }, op: "Reading", path: Some("./myfile2CyhlQ.test"), fd: Some(8) }', lib.rs:90:18
```

Here's the test:
```
    #[test]
    fn basic() {
        let normal_tmpfile = tempfile::Builder::new()
            .prefix("myfile")
            .suffix(".test")
            .tempfile_in(std::env::current_dir().unwrap())
            .unwrap();
        eprintln!("Running on {:?}", normal_tmpfile.path());

        let ex = glommio::LocalExecutorBuilder::new(glommio::Placement::Fixed(0))
            .make()
            .unwrap();

        ex.run(async move {
            let file = glommio::io::OpenOptions::new()
                .read(true)
                .write(true)
                .create(true)
                .dma_open(normal_tmpfile.path())
                .await
                .unwrap();

            let initial_read = file
                .read_at_aligned(0, 4096)
                .await
                .expect("Failed to read from empty file");
            assert_eq!(initial_read.len(), 0);

            let mut buf = file.alloc_dma_buffer(4096);
            eprintln!("Allocated {} bytes", buf.len());
            buf.as_bytes_mut().fill(0);
            eprintln!("Writing [0, {}]", buf.len());
            file.write_at(buf, 0)
                .await
                .expect("Failed to write [0, 4096)");

            eprintln!("Reading [0, 4096)");
            file.read_at(0, 4096)
                .await
                .expect("Failed to read what we just wrote");
        });
    }
```

Tested configurations:
Kernels tested:
6.5.4-1-MANJARO
6.5.4-2-ck
6.1.55-1-MANJARO
I've also tried pinning the test to P cores (13900K) in case that was some kind of issue

| Filesystem | Logical Block Size | Physical Block Size | Test Passes |
|-|-|-|-|
|tmpfs | ? | ? | [x] |
| btrfs | 4096 | 4096 | [x] |
| ext4 | 4096 | 4096 | [] |
| xfs | 4096 | 4096 | [] |
| xfs | 512 | 512 | [] |
| f2f | 4096 | 4096 | [] |

[...]

PR #702, opened 2026-01-13
Update dependencies

Michael-Wigham:
### What does this PR do?

- Bumps MSRV to 1.92
  - Applies compiler-suggested std API updates and style cleanups.
  - Modernises internals for newer dependency versions
    - Moves file descriptor handling to OwnedFd/BorrowedFd where appropriate for clearer ownership and safer close semantics.
- Updates crate dependencies across the workspace, including updating the liburing submodule to 2.13.
- Updates sysfs probing logic:
  - Stops assuming `minor == 0` identifies the parent block device; this is incorrect for partitions and NVMe devices where valid block devices frequently have non-zero minors
  - Adds partition-safe queue discovery (walk-up to find a queue/directory).
  - Makes queue/io_poll optional (if files are missing - assume unsupported) to handle older kernels/drivers and avoid hard failures when sysfs files are absent.
  - Simplifies cached block property access and removes the “probe iopoll by performing an I/O” path from the reactor.
  - Fixes CI failures on partition-backed and virtual block devices where `queue/*` files are missing or located on the parent disk.
- spsc_queue changes:
  - Eliminate manual allocation and raw pointer arithmetic for buffer, replacing it with an owned slice.
  - Cleans up lookahead/limit handling and switches head/tail progression to wrapping arithmetic to ensure correctness across wrap boundaries.
  - The underlying queue algorithm remains unchanged; this is a safety and maintainability improvement rather than a semantic change.
- Cleans up examples and benchmarks:
  - Removes unused hyper timer glue that was generating warnings.
  - Updates the SPSC benchmark to pin producer/consumer threads to CPUs and prints clearer latency/throughput numbers.

### Motivation

This is a maintenance PR to keep Glommio building cleanly on modern toolchains and dependency versions, and to unblock CI failures.

Keeping the crate aligned with recent stable Rust:
- Reduces friction for contributors using current toolchains
- Prevents accumulating breakage from upstream ecosystem changes
- Lowers the long-term maintenance cost for future feature work

### Notes on behaviour changes

- Poll-ring eligibility is now determined by sysfs capability values (e.g. `queue/io_poll`) combined with conservative device gating, rather than probing support via a test DMA read.
  - This avoids partially-working states where a device appears pollable until the first I/O fails.
  - This may relate to #605 (devices appearing “pollable” until the first read routes via the poll ring and fail).
- Sysfs reads are now more defensive:
  - Missing files fall back to safe defaults instead of panicking.
  - This ensures Glommio continues to work on older kernels, non-NVMe devices, and CI environments.

### Related issues

No related issues - standalone MR.

### Additional Notes

- No public API changes.
- Functional behavior is unchanged for NVMe devices with full sysfs support.
- Behaviour on non-NVMe, virtual, or partition-backed devices avoids runtime and CI failures.
- The examples workspace is now edition = "2024" to align with the updated toolchain and formatting expectations.
- The SPSC benchmark output is intended for easier interpretation in CI/local perf runs and is not a functional API change.
  - The new bench also supports comparing different core pairings; on SMT systems, this can highlight differences between sibling vs non-sibling cores.

### Checklist

- [] I have added unit tests to the code I am submitting
- [] My unit tests cover both failure and success scenarios
- [] If applicable, I have discussed my architecture

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
           cat << EOF > "run-gha-workflow.sh"
           PATH=$PATH:/usr/share/rust/.cargo/bin
           echo "`nproc` CPU(s) available"
-          rustup install 1.70
+          rustup install 1.92
           rustup show
           rustup default stable
           cargo install cargo-sort
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,3 @@
 [workspace]
-members = [
-  "examples",
-  "glommio",
-]
+members = ["examples", "glommio"]
 resolver = "2"
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ an [introductory article.](https://www.datadoghq.com/blog/engineering/introducin
 
 ## Supported Rust Versions
 
-Glommio is built against the latest stable release. The minimum supported version is 1.70. The current Glommio version
+Glommio is built against the latest stable release. The minimum supported version is 1.92. The current Glommio version
 is not guaranteed to build on Rust versions earlier than the minimum supported version.
 
 ## Supported Linux kernels
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -7,19 +7,19 @@ edition = "2021"
 
 [dev-dependencies]
 byte-unit = "5.1.4"
-yansi = "~0.5.1"
 clap = "4.5.3"
 fastrand = "2"
 futures = "~0.3.5"
 futures-lite = "2.6.0"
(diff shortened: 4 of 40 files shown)


humb1t:
This fork works like a charm, thanks @Michael-Wigham !

A question to datadog team. If there is no maintainers inside the company ready to support this library, to prevent the hard fork, can you please consider some OSS mainteiners to be added to the project and repo itself moved outside of DataDog organization?
I really appreciate all the work that you have made for glommio and I think the project deserves to live and move on. While 11 months without commits in master and reasonable amount of PRs waiting to be merged makes asks for such a move to  let it be free.

PR #703, opened 2026-02-11
Fix #700: Remove public Clone from spsc_queue to prevent memory corruption

### LINKED ISSUE #700: Memory corruption in glommio::channels::spsc_queue under concurrent use
### Description

I’m seeing a heap corruption crash when using `glommio::channels::spsc_queue` from multiple producer and consumer threads.

The code only uses safe Rust and the public `glommio::channels::spsc_queue` API, but at runtime it fails with:

```text
Avvio demo Glommio Multi-Thread...
malloc(): chunk size mismatch in fastbin
malloc(): chunk size mismatch in fastbin
Aborted (core dumped) cargo run
```

This looks like memory corruption / unsoundness in safe code.

I realize this is an SPSC queue and that my sample intentionally uses multiple producers and multiple consumers, but since the API is fully safe, it shouldn’t be possible to trigger heap corruption (ideally the type system or runtime checks would prevent this usage, or at worst it should fail cleanly without UB).

---

### Environment

* Glommio version: `0.9.0` (from github)
* Rust version: `rustc 1.xx.x (stable)`
* OS: Linux x86_64 Fedora
* Build: `cargo run --release` and `cargo run` both affected

---

### Minimal reproducible example

```rust
use std::error::Error;
use glommio::channels::spsc_queue;

fn main() -> Result<(), Box<dyn Error>> {
    println!("Avvio demo Glommio Multi-Thread...");
    let (sender, mut receiver) = spsc_queue::make::<String>(4096);

    let mut producers = Vec::new();
    let mut consumers = Vec::new();

    // 2 producers
    for i in 0..2 {
        let sender = sender.clone();
        producers.push(std::thread::spawn(move || {
            for j in 0..500_000 {
                let msg = format!("Messaggio {} dal produttore {}\n", j, i);
                // ignoring the return value for simplicity
                let _ = sender.try_push(msg);
            }
        }));
    }

    // 2 consumers
    for _ in 0..2 {
        let receiver = receiver.clone();
        consumers.push(std::thread::spawn(move || {
            for _ in 0..500_000 {
                let _ = receiver.try_pop();
            }
        }));
    }

    for producer in producers {
        producer.join().unwrap();
    }
    for consumer in consumers {
        consumer.join().unwrap();
    }

    Ok(())
}
```

Running this with `cargo run` produces:

```text
Avvio demo Glommio Multi-Thread...
malloc(): chunk size mismatch in fastbin
malloc(): chunk size mismatch in fastbin
Aborted (core dumped) cargo run
```

---

### Expected behavior

Even if using `spsc_queue` in a way that violates its logical “single producer / single consumer” design, the use of safe APIs should not result in undefined behavior or heap corruption.

Ideally one of the following would happen instead:

* The type system prevents this misuse (e.g. `Producer` / `Consumer` not being `Clone`, or being more constrained), or
* There are documented guarantees that misusing the queue can cause UB, or
* The queue detects this situation and fails in a defined way (panic, error, etc.), rather than corrupting memory.

---

### Actual behavior

* Program prints `Avvio demo Glommio Multi-Thread...`

* Then glibc reports:

  ```text
  malloc(): chunk size mismatch in fastbin
  malloc(): chunk size mismatch in fastbin
  Aborted (core dumped)
  ```

* This suggests memory corruption happening internally, triggered only through safe `Producer::try_push` and `Consumer::try_pop` calls.

---

### Additional notes

* `Producer<T>` and `Consumer<T>` are `Send` but `!Sync`, and they are `Clone`. This allows creating multiple cloned producers/consumers that operate concurrently on the same underlying queue from different threads.
* The module name `spsc_queue` suggests single-producer/single-consumer behavior, but the current API surface (especially `Clone` + `Send`) makes it easy to use it in multi-producer/multi-consumer scenarios without any compiler errors or warnings.
* Even if this usage is considered invalid, the fact that it leads to heap corruption in safe Rust feels like at least a documentation or soundness issue.

If this usage is explicitly unsupported and considered undefined behavior, it might be helpful to:

1. Clarify that in the documentation for `spsc_queue::make`, `Producer`, and `Consumer`, and/or
2. Tighten the API (e.g. remove `Clone` or add runtime checks) so that this pattern doesn’t compile or at least doesn’t lead to memory corruption.

dahankzter:
## Summary

Fixes #700 - Removes the public `Clone` implementation from `Producer<T>` and `Consumer<T>` to prevent memory corruption in safe code.

## Problem

The `spsc_queue` is designed for Single-Producer-Single-Consumer access with `Ordering::Relaxed` for performance. The public `Clone` trait implementation allowed external code to create multiple producers or consumers, which violates SPSC guarantees and causes:

- **Double-free** when multiple consumers read the same value from the queue
- **Heap corruption** (`malloc(): chunk size mismatch in fastbin`)
- **Memory leaks** when multiple producers overwrite values without dropping
- **Undefined behavior** in safe Rust code

## Solution

1. **Removed** `Clone` trait implementation from `Producer<T>` and `Consumer<T>`
2. **Added** crate-private `clone_internal()` methods for internal use within glommio
3. **Updated** `shared_channel` to use `clone_internal()` for connection handoff
4. **Added** documentation explaining SPSC guarantees and safety requirements

## Breaking Change

This is a **breaking API change**. External code that clones `Producer` or `Consumer` will no longer compile. However, such code was already unsound and could cause heap corruption.

Internal uses within glommio (e.g., `shared_channel`) are safe because they ensure only one producer/consumer is active at any given time.

## Testing

- All existing tests pass
- The reproduction case from #700 now correctly fails to compile (preventing the bug)
- Verified in Linux container with kernel 6.16 and io_uring support

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
           cat << EOF > "run-gha-workflow.sh"
           PATH=$PATH:/usr/share/rust/.cargo/bin
           echo "`nproc` CPU(s) available"
-          rustup install 1.70
+          rustup install 1.92
           rustup show
           rustup default stable
           cargo install cargo-sort
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,3 @@
 [workspace]
-members = [
-  "examples",
-  "glommio",
-]
+members = ["examples", "glommio"]
 resolver = "2"
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ an [introductory article.](https://www.datadoghq.com/blog/engineering/introducin
 
 ## Supported Rust Versions
 
-Glommio is built against the latest stable release. The minimum supported version is 1.70. The current Glommio version
+Glommio is built against the latest stable release. The minimum supported version is 1.92. The current Glommio version
 is not guaranteed to build on Rust versions earlier than the minimum supported version.
 
 ## Supported Linux kernels
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -7,19 +7,19 @@ edition = "2021"
 
 [dev-dependencies]
 byte-unit = "5.1.4"
-yansi = "~0.5.1"
 clap = "4.5.3"
 fastrand = "2"
 futures = "~0.3.5"
 futures-lite = "2.6.0"
 glommio = { path = "../glommio" }
+http-body-util = "0.1.0"
 
 # hyper and tokio for the hyper example. We just need the traits from Tokio
 hyper = { version = "1.2.0", features = ["full"] }
 num_cpus = "1.13.0"
-sys-info = "~0.8.0"
-http-body-util = "0.1.0"
 serde_json = "1.0.114"
+sys-info = "~0.9.1"
+yansi = "~1.0.1"
 
 [[example]]
 name = "echo"
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -3,7 +3,7 @@ name = "examples"
 version = "0.0.0"
 license = "Apache-2.0 OR MIT"
 publish = false
-edition = "2021"
+edition = "2024"
 
 [dev-dependencies]
 byte-unit = "5.1.4"
--- a/examples/deadline_writer.rs
+++ b/examples/deadline_writer.rs
@@ -69,11 +69,11 @@ impl IntWriter {
 
                 println!(
                     "{}: Wrote {} ({}%), {:.0} int/s, scheduler shares: {} , {:.2} % CPU",
-                    Paint::blue(format!("{}s", elapsed.as_secs())),
+                    format!("{}s", elapsed.as_secs()).blue(),
                     self.count.get(),
-                    Paint::new(format!("{:.0}", ratio)).bold(),
+                    format!("{:.0}", ratio).bold(),
                     intratio,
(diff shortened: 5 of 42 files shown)


dahankzter:
All failures show the same error:
  Os { code: 95, kind: Unsupported, message: "Operation not supported" }
  Operation: "Reading" from DMA files

  This occurs because:
  - DMA files require `O_DIRECT` support (direct I/O)
  - GitHub Actions CI runs tests on `/tmp` (likely tmpfs/overlayfs)
  - These filesystems don't support `O_DIRECT` operations
  - The test output even warns: *"Glommio currently only supports NVMe-backed volumes formatted with XFS or EXT4"*

henrik-leovegas:
## Update: Rebased on PR #702

This PR has been rebased on top of #702 (dependency updates) to inherit its passing CI infrastructure. 

### What changed from the rebase:

**From PR #702** (most of the file changes):
- Updated dependencies including `rand` from 0.10 to 0.9
- Various other dependency and infrastructure updates
- This explains why the PR shows 617 additions / 537 deletions across 43 files

**Our actual fix** (the core changes for issue #700):
- `glommio/src/channels/spsc_queue.rs` - Removed public `Clone`, added `clone_internal()`
- `glommio/src/channels/shared_channel.rs` - Updated to use `clone_internal()`
- `glommio/benches/shared_channel.rs` & `glommio/benches/competing_io.rs` - Fixed for rand 0.9 API

All CI checks now pass ✅

